PR #152, opened 2025-06-25
fix(js-client): only add version parameter for CDN apis, default to published

edodusi:


packages/js-client/src/constants.ts
```diff
@@ -15,7 +15,7 @@ export const STORYBLOK_AGENT = 'SB-Agent';
 export const STORYBLOK_JS_CLIENT_AGENT = {
   defaultAgentName: 'SB-JS-CLIENT',
   defaultAgentVersion: 'SB-Agent-Version',
-  packageVersion: '6.0.0',
+  packageVersion: '7.0.0',
```
imranolas, line 18: I wonder if we can pull the version from the package.json directly

```suggestion
  packageVersion: '7.0.1',
```

edodusi: yep, but who's updating the package.json 😅

PR #174, opened 2025-07-03
test: add test script to nuxt package

m-shum:
+ This commit enables e2e testing of the @storyblok/nuxt package when running `pnpm nx run-many --target=test --all`


imranolas (approved):
Thanks for the PR 🙏 

The CI failure is expected since you're an external contributor and can't retrieve the environment secrets. We'll need to accommodate how to handle these in future.

PR #201, opened 2025-07-17
chore: migrate region-helper

imranolas:
Moves `@storyblok/region-helper` from its standalone repository into the monorepo

## Changes
- Add region-helper package to `packages/`
- Remove standalone repo infrastructure (workflows, changeset config, dependabot)
- Update ESLint config for monorepo compatibility
- Add CODEOWNERS and issue template entries
- Update package manifest and README formatting

Fixes ESLint 9.x compatibility while preserving original `eslint-config-unjs` configuration.

resolves WDX-49


packages/region-helper/eslint.config.mjs
Dawntraoz, line 1: Thanks for migrating it @imranolas 🫂 Appreciate it!

repo-manifest.json
Dawntraoz, line 1: Maybe I'm losing the mono repo context here again, but why are the other packages removed by adding `region-helper`? Do you know if this is expected?

Dawntraoz (approved):
Since the packages that were upgraded were just `devDependencies` I don't think it is a blocker for this PR, but @demetriusfeijoo it will be nice to check locally together that it works as expected after the upgrade here 🙏

packages/region-helper/CHANGELOG.md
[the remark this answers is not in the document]
imranolas: No dumb questions. This is handled by nx now. The changelog has been replaced by github releases

packages/region-helper/.editorconfig
alvarosabu, line 1: Non-blocking: 

Hi @imranolas do you think this would affect the editor for others on the monorepo?

@Dawntraoz @demetriusfeijoo to prevent formatting conflicts between different IDE setups while preserving individual editor preferences, consider running `eslint --fix` in CI or as a pre-commit hook. This way ESLint handles the formatting rules rather than relying on individual IDE configurations.

packages/region-helper/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+1. In the root of the repository, run `npx changeset`
+2. Stage the new created Changeset file and push it
+3. Open a PR from `feat/branch` to `main`
+4. Review and merge the PR
+5. A `bump pull request` is going to be created automatically created by the `versioning.yml` action.
+6. Open the automatically generated PR (`chore: release`) review it and merge.
+7. After it gets merged, another `action` is going to run, the `publishing.yml` action, which is responsible for publishing the new version to the NPM, to create a new `tag`, and also to create a Github release.
+
+> Note: If the Github checks don’t run for the second PR (step 6), you may close and reopen it. For the video version check the [internal docs](https://www.notion.so/storyblok/Recordings-d8838dc8a76f49e6b393db654a580bfa?pvs=4#6b34969027a24ddf86c85780c123efd5).
```
demetriusfeijoo, line 96: Hey @imranolas

I think we can may remove these steps once the release process is going to be different now 🙏

packages/region-helper/.editorconfig
jeremycolin: I also agree we should not have package specific styles config

jeremycolin (commented):
non clocking as well.

demetriusfeijoo (approved):
Hey @imranolas 👋 

I tested it locally and everything worked great. 🙏 
I have left a comment about prettier formatting, but I don't think it is a blocker, so I'm approving it.

Great job and many thanks for handling it 🚀 🚀

PR #208, opened 2025-07-22
feat: add type-suffix option to ts generate command

omonk:
We need to upgrade from v3 of the storyblok cli as it has a vulnerability in the version of `form-data` via `axios`

The problem we're encountering is that the new type generator removes the behaviour of type suffixes. We use Storyblok types throughout our nextjs app and would rather keep this behaviour the same so:

1. We avoid having to rewrite the names 
2. Types are clearly coming from Storyblok components

This PR adds a flag to the ts generate command to provide the opposite of `prefix` and adds a `--types-suffix` option to the command.

### DIFF
--- a/packages/cli/src/commands/types/generate/actions.test.ts
+++ b/packages/cli/src/commands/types/generate/actions.test.ts
@@ -292,6 +292,20 @@ describe('getComponentType', () => {
     expect(getComponentType('test_component', options)).toBe('TestComponent');
   });
 
+  it('should apply typeSuffix when provided', () => {
+    const options: GenerateTypesOptions = {
+      typeSuffix: 'CustomSuffixValue',
+    };
+    expect(getComponentType('test_component', options)).toBe('TestComponentCustomSuffixValue');
+  });
+
+  it('should handle empty typeSuffix', () => {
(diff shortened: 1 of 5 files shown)


omonk:
@alvarosabu unsure of the failures in this CI run, they're unrelated to my changes. Are you able to review?

imranolas:
> @alvarosabu unsure of the failures in this CI run, they're unrelated to my changes. Are you able to review?

@omonk No, this isn't your fault. Our environments don't permit external PRs to access workflow secrets which is why this fails. It's expected.

omonk:
@imranolas thanks for confirming, sounds reasonable.

Do we know if/when this might be accepted, merged and released? Thanks!

alvarosabu:
> @imranolas thanks for confirming, sounds reasonable.
> 
> Do we know if/when this might be accepted, merged and released? Thanks!

Im currently downloading it to test it

omonk:
@alvarosabu good feedback on the branching

Thanks for reviewing and merging, when do you think we might be able to use it in a new version? 

I was curious, what was the intention behind removing the default suffix that v3 had?

PR #241, opened 2025-08-06
refactor(js-client): update package.json exports and class export

### LINKED ISSUE #130: Type are not resolved correctly when  having `"type": "module"` in package.json

When importing `storyblok-js-client` inside a typescript project where `"moduleResolution": "nodenext"` is used in the typescript `compilerOptions` and `"type": "module"` is set inside the package.json, the types are not resolved correctly.

It seems like, this is a known typescript issue: [#60561](https://github.com/microsoft/TypeScript/issues/60561#issuecomment-2784613209)

---

## Expected Behavior

Types should be correctly resolved from the `storyblok-js-client` package when `type` is set to `module`.

## Current Behavior

Types are not resolved correctly and typescript will throw the following error:

```
This expression is not constructable.
  Type 'typeof import(".../node_modules/storyblok-js-client/dist/types/entry.esm")' has no construct signatures. ts(2351)
```

## Steps to Reproduce

Please see a simple [reproduction on Stackblitz](https://stackblitz.com/edit/storyblok-js-client-typescript-issue).

imranolas:
Simplifies the js-client package structure by consolidating exports and switching from Vite to tsdown for building. The package now uses a cleaner export pattern with the Storyblok class as the default export.

Key changes:
- Added `"type": "module"` to package.json for proper ES module support
- Simplified exports to use single entry points (`./dist/index.js` and `./dist/index.cjs`)
- Removed separate entry files (entry.esm.ts, entry.umd.ts) and consolidated exports in main index.ts
- Switched build system from Vite to tsdown with new tsupdown.config.ts
- Updated Storyblok class to be exported directly from index.ts
- Removed unused dependencies (vite, vite plugins) and added tsdown
- Updated type definitions to use `./dist/index.d.cts`

This provides a more standard ES module interface while maintaining backward compatibility through CommonJS exports.

fixes #130


edodusi (approved):
As long as the release notes clearly highlights the changes we are making to the bundle, I think we can release this 👍

stijnbernards:
@edodusi The exports have been completely changed in this update and it has been released as a `patch` update.

We can no longer access: `StoryblokContentVersionKeys` and have to use `version: ISbStoriesParams['version'] = "published",` now for example.

Which is a breaking change imho.

PR #267, opened 2025-08-22
chore(astro): upgrade all Astro playgrounds to the latest version

dipankarmaikap:
This PR upgrades the **test**, **SSR**, and **SSG** playgrounds to the latest dependencies.

### Changes included:

* Updated dependencies across all playgrounds
* Moved common components into a new `shared` folder, allowing all playgrounds to reference the same source.
* Removed duplicated component code from each playground to reduce maintenance overhead.

### Benefits:

* Consistent dependency versions across playgrounds.
* Simplified component management with a single shared source.
* Easier future upgrades and less risk of divergence between playgrounds.


maoberlehner:
> @imranolas @maoberlehner weird, for some reason the linters on both richtext and react have a memory leak on this PR: Screenshot 2025-08-26 at 17 01 58
> 
> https://github.com/storyblok/monoblok/actions/runs/17156197963/job/48884599933?pr=267
> 
> I tested this branch locally, and I can reproduce the lint issues on those packages, I confirm lint is working on main, it must be a dependency change in this PR

Seems like a known issue: https://github.com/typescript-eslint/typescript-eslint/issues/11455
Just trying some combination of ESLint related package versions. Seems promising.

maoberlehner:
[This commit fixes the callstack error,](https://github.com/storyblok/monoblok/pull/267/commits/3474590553620e3e7f795929e9970d3c6dff4932) but because of upgrading the `@antfu/eslint-config` ESLint base config, there are now linting errors. If we don't want to go that route, we need to downgrade TypeScript.

maoberlehner:
@imranolas makes sense! I've created a new PR: https://github.com/storyblok/monoblok/pull/275

As for this repo: I think the `pnpm-lock` was deleted, and dependencies were installed from scratch, installing all the latest dependencies of everything, leading to the incompatibility with our linter config. I've reverted the lock file and reinstalled the dependencies, so only the actually updated dependencies are also updated in the lock file.

PR #275, opened 2025-08-27
chore: make ESLint config compatible with latest TS ESLint

maoberlehner:
See: typescript-eslint/typescript-eslint#11455


alvarosabu:
@maoberlehner should we revise if we need to update our core eslint package?

PR #281, opened 2025-09-01
feat: wdx 106 migrate cli mapi client

alvarosabu:
This PR migrates the Storyblok CLI from a custom MAPI client implementation to the new `@storyblok/management-api-client` package. This change improves code maintainability, reduces duplication, and provides better type safety through centralized API client management.

### 🔧 Core Infrastructure
- **Replaced custom MAPI client** with `@storyblok/management-api-client` package
- **Updated API client initialization** to use new configuration structure with `accessToken` instead of direct token
- **Improved error handling** with better type safety and consistency across API calls
- **Enhanced request/response interceptors** for better logging and debugging

### 📦 Package Dependencies
- Added `@storyblok/management-api-client` as a dependency
- Updated package.json with new client library

### 🏗️ Command Refactoring
Migrated all CLI commands to use the new client:

#### Components
- **Pull/Push components** now use typed API methods (`components.list()`, `components.create()`, `components.update()`)
- **Component groups** migrated to `componentFolders.list/create/update()`
- **Component presets** migrated to `presets.list/create/update()`
- **Internal tags** migrated to `internalTags.list/create/update()`

#### Spaces
- **Space creation** migrated to use new typed endpoints
- **Language handling** updated with `fetchLanguages()` using new MAPI client
- **Enhanced space properties** with org and partner related fields

#### User Management
- **User retrieval** simplified with better error handling
- **Login flow** improved with new token structure
- **User schema** enhanced with `friendly_name` property

#### Stories
- **Story fetching** refactored to handle pagination with headers
- **Split fetchStories methods** for better separation of concerns
- **Improved pagination handling** with proper header-based pagination

#### Datasources
- **Pull/Push/Delete operations** migrated to new client methods
- **Better error handling** and response processing

### 🧪 Testing Updates
- **Updated test mocks** to use new token structure with `accessToken`
- **Fixed mock user objects** to match new schema structure
- **Enhanced test coverage** for error scenarios
- **Updated component tests** for new API response structure

### 🔍 Type Safety Improvements
- **Added proper TypeScript types** from the management API client
- **Replaced custom interfaces** with generated types from OpenAPI specs
- **Enhanced type safety** across all API interactions

### 📋 OpenAPI Schema Updates
- **Enhanced space schema** with language, organization, and partner properties
- **Added user friendly_name** property to user schema
- **Improved API documentation** alignment


packages/cli/src/commands/datasources/push/actions.ts
```diff
@@ -90,14 +101,17 @@ export const updateDatasourceEntry = async (space: string, entryId: number, entr
   try {
     const client = mapiClient();
 
-    await client.put<{
-      datasource_entry: SpaceDatasourceEntry;
-    }>(`spaces/${space}/datasource_entries/${entryId}`, {
-      body: JSON.stringify({
+    // TODO: ask Imran why this update method is called differently
```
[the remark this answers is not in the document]
alvarosabu: Hey @imranolas, I left this TODO to ask you if it's something to be corrected, or if it has a reason to be different from other SDKs.

packages/cli/src/commands/datasources/push/actions.ts
```diff
@@ -11,10 +11,14 @@ export const pushDatasource = async (space: string, datasource: SpaceDatasource)
...
-    const { data } = await client.post<{
-      datasource: SpaceDatasource;
-    }>(`spaces/${space}/datasources`, {
-      body: JSON.stringify(datasource),
+    const { data } = await client.datasources.create({
+      path: {
+        space_id: Number.parseInt(space),
+      },
+      body: {
+        datasource,
```
alvarosabu, line 19: Hi @imranolas I'm getting a type error in the body of a couple of create methods, but only in the create, update methods are fine. 
Screenshot 2025-09-02 at 08 40 44


Screenshot 2025-09-02 at 08 38 07

packages/cli/src/commands/datasources/push/actions.ts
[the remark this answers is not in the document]
imranolas: I don't recall but I'd guess it's because datasources and entries are both handled in this file and we need different methods for them.

packages/cli/src/commands/components/pull/actions.test.ts
```diff
@@ -142,10 +143,12 @@ describe('pull components actions', () => {
     expect(result).toEqual(mockResponse.components[0]);
   });
 
-  it('should throw an masked error for invalid token', async () => {
-    mapiClient().dispose();
+  // TODO: Ask team regarding resseting the mapi client options
```
maoberlehner, line 146: As a developer testing it and also as a user using it, I'd love to be able to use `createMapiClient`. Maybe we should export it?

packages/cli/src/commands/components/pull/actions.ts
```diff
@@ -9,11 +9,14 @@ export const fetchComponents = async (space: string): Promise<SpaceComponent[] |
...
-    }>(`spaces/${space}/components`, {
+    const { data } = await client.components.list({
+      path: {
+        space_id: Number.parseInt(space, 10),
+      },
+      throwOnError: true,
     });
-    return data.components;
+
+    return data?.components as SpaceComponent[];
```
maoberlehner, line 19: suggestion: I think we should strive to not use `as` as much as possible. In this case we can use a typeguard instead: `if (!data.components) throw new Error('Error message that makes sense')`

packages/cli/src/commands/components/pull/actions.ts
```diff
@@ -24,11 +27,17 @@ export const fetchComponent = async (space: string, componentName: string): Prom
   try {
     const client = mapiClient();
 
-    const { data } = await client.get<{
-      components: SpaceComponent[];
-    }>(`spaces/${space}/components?search=${encodeURIComponent(componentName)}`, {
+    const { data } = await client.components.list({
+      path: {
+        space_id: Number.parseInt(space, 10),
```
maoberlehner, line 32: suggestion (non-blocking): I've noticed that `space_id` allows both `string` and `number`, so parsing it is not strictly necessary here.

General side note: in cases where we want to transform a string that must be a valid number to a number (e.g. `"123"`) `Number(string)` is often preferable to `parseInt()`. `parseInt` is the right solution when dealing with messy input (e.g. `"20px"`).

packages/cli/src/commands/components/constants.ts
```diff
@@ -1,54 +1,11 @@
...
-  description: string;
-}
-
-export interface SpaceComponentInternalTag {
-  id: number;
-  name: string;
-  object_type?: 'asset' | 'component';
-}
+export type SpaceComponent = Components.Component;
+export type SpaceComponentGroup = ComponentFolders.ComponentFolder;
```
maoberlehner, line 6: question(non-blocking): would it make sense to make the type names the same? Talking about either a Group or a Folder makes it confusing.

packages/cli/src/commands/languages/actions.test.ts
```diff
@@ -61,18 +66,10 @@ describe('pull languages actions', () => {
...
     });
   });
-  it.skip('should throw an masked error for invalid token', async () => {
-    // TODO: Fix this test, is the only of it's kind that is not working, it's not clear why
-    // Test return "Compared values have no visual difference." but fails anyway
-    const error = new FetchError('Non-JSON response', { status: 401, statusText: 'Unauthorized', data: null });
-    await expect(fetchLanguages('12345', 'invalid-token', 'eu')).rejects.toThrow(
-      new APIError('unauthorized', 'pull_languages', error, `The user is not authorized to access the API`),
-    );
-  });
```
maoberlehner, line 75: Make it work, or make it go away. Love it!

packages/cli/src/commands/languages/actions.ts
```diff
@@ -1,27 +1,19 @@
...
-    return {
-      default_lang_name: response.space.default_lang_name,
-      languages: response.space.languages,
-    };
+    const space = await fetchSpace(spaceId);
+    if (space?.default_lang_name !== undefined && space?.languages?.length) {
+      return {
+        default_lang_name: space?.default_lang_name,
+        languages: space?.languages,
+      } as SpaceInternationalization;
```
maoberlehner, line 15: question(non-blocking): Can we use `satisfies SpaceInternationalization` here to avoid `as`?

maoberlehner (commented):
Looks good! Just a few suggestions / questions.

packages/cli/src/commands/components/constants.ts
alvarosabu: Not sure if it's something in our shoes to fix, the API mixes those, because to the end user, they are folders, but in the API properties, they are called groups.

PR #283, opened 2025-09-02
fix: wdx 36 list custom bloks richtext

### LINKED ISSUE #162: 🐛 [Bug]: List of components in richtext do not get resolved
### Package

### Bug Description

First of all: glad to see you are actively working on this and moving it all to a monorepo. 👍 

I am not a 100% sure if I am doing something wrong but when I add multiple components / bloks right next to each other within a richtext, the object structure I get is as follows (compare Screenshot). As can be seen, the four components next to each other, end up all in one array-item of type 'blok'. The `<StoryblokRichText ... />` component of `@storyblok/react` does not seem to be capable of rendering multiple items and only renders the first element. Not sure if the issue lies in storyblok/react or storyblok/richtext. 🤷‍♂ 

I found a workaround (compare code-example below) but that seems hacky and it would be ideal if the richtext component was capable of handling that itself. Let me know if you need further information or help to reproduce this. 

Image

### Steps to Reproduce

1. Add multiple components within richtext
2. Use the `<StoryblokRichText ... />` from `@storyblokreact` to try to render them

### Expected Behavior

Components within richtext, even if they are placed right next to each other, are being rendered by the richtext-component without any further modification.

### Actual Behavior

Only the first component (of multiple components) is being rendered. 

### Code Sample

```javascript
// Here is some code (workaround) that I used to make it work.
/**
 * Flatten the document structure:
 * 1. If an item is a "blok" with multiple items in its `attrs.body` array,
 *    we take those items out to the top level of the object
 * 2. If an item is a standard node (like paragraph, list, etc.),
 *    we leave it as is.
 *
 * This is done to ensure that the document structure is flat
 * as the StoryblokRichText component seems to expect a flat structure.
 */
const flattenBlokContent = (blok: StoryblokRichTextDocumentNode) => {
  if (blok.type === 'blok' && Array.isArray(blok.attrs?.body)) {
    return blok.attrs.body.map((nestedBlok) => ({
      type: 'blok',
      attrs: {
        // eslint-disable-next-line no-underscore-dangle
        id: nestedBlok._uid, // Use the component's unique ID.
        body: [nestedBlok], // Place the component inside the body array.
      },
    })) as StoryblokRichTextDocumentNode[];
  }
  return blok;
};

  const finalDoc = {
    type: 'doc',
    // This seems to be a Storyblok richtext limitation where it expects a flat structure.
    content: blok?.text?.content?.flatMap((contentBlok) => flattenBlokContent(contentBlok)) || [],
  } as any; // not sure how to type this correctly

<StoryblokRichText doc={finalDoc} />
```

### Environment

```shell
System:
    OS: macOS 15.1.1
    CPU: (8) arm64 Apple M3
    Memory: 78.66 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.19.1 - ~/.local/state/fnm_multishells/36344_1750757960440/bin/node
    npm: 10.8.2 - ~/.local/state/fnm_multishells/36344_1750757960440/bin/npm
    pnpm: 9.14.2 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 137.0.7151.122
    Chrome Canary: 140.0.7259.0
    Edge: 138.0.3351.55
    Safari: 18.1.1
```

### Error Logs

```shell

```

### Additional Context

Using the latest packages:

    "@storyblok/react": "5.1.1",
    "@storyblok/richtext": "3.4.0",

alvarosabu:
Potentially closes #162


packages/react/src/core/richtext-hoc.ts
```diff
@@ -42,6 +48,8 @@ export function createRichTextHook(
     // Use the same component resolver that's exported
     const componentResolver = createComponentResolver(Component, { isServerContext });
 
+    const { resolvers: _, ...optionsWithoutResolvers } = options;
```
maoberlehner, line 51: question(non-blocking): I don't understand this. Maybe we can clarify the code somehow to make it self explanatory why this is necessary?

Issue #280, opened 2025-09-01
🐛 [Bug]: `useStoryblokState` can cause state de-sync


FabianKoder:
### Package

@storyblok/react (React Integration)

### Bug Description

I have the following setup:

- A Next.js page with SSG props
  - rawStory (The raw fetched Storyblok story)
  - relatedStories (A list of minimal related story data that are fetched and processed on the server-side)
- Using `const story = useStoryblokState(rawStory, { resolveLinks: 'url' });` - this is used during rendering, since this state is potentially altered by the bridge integration in comparison to the original story object coming from the server
- Now when I switch pages (mainly relevant for production users) from Story A to Story B via a client-side navigation (e.g. next/link), both props correctly update, but `useStoryblokState` does not immediately update
- Since it uses a useEffect pattern to update its internal state, only on the next render cycle its state gets updated
- This results in a short desync of server-provided data and client data

While this may not look horrible on paper, we have data inside the story that depends on data that comes from the server. This short desync causes weird, unexpected behavior because it is basically a race condition.

### Steps to Reproduce

1. Create a Next.js page with getStaticProps
2. Fetch story data in there and pass it into the props
3. Log the story data passed from the server in the client
4. Use hook useStoryblokState and log its result
5. Change page with a client-side transition, e.g. using next/link
6. See that server-passed story != hook processed story (= still the old story from previous page)
7. On next render, both are "in sync" again

### Expected Behavior

A new story passed into `useStoryblokState` should immediately be reflected in the output of `useStoryblokState` instead of a 1-render-cycle desync because of the internally used useEffect pattern.

### Actual Behavior

Story data passed into `useStoryblokState` desyncs

### Code Sample

```javascript

```

### Environment

```shell
System:
    OS: macOS 15.6.1
    CPU: (11) arm64 Apple M3 Pro
    Memory: 143.47 MB / 36.00 GB
    Shell: 3.7.0 - /opt/homebrew/bin/fish
  Binaries:
    Node: 20.11.0 - ~/.nvm/versions/node/v20.11.0/bin/node
    npm: 10.8.3 - ~/.nvm/versions/node/v20.11.0/bin/npm
    pnpm: 9.4.0 - ~/.nvm/versions/node/v20.11.0/bin/pnpm
    bun: 1.2.8 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 138.1.80.115
    Chrome: 139.0.7258.155
    Safari: 18.6
  npmPackages:
[...]

PR #285, opened 2025-09-04
feat(react): prevent unnecessary re-rendering

maoberlehner:
Updating the state on prop change is an anti pattern. See: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

Also see discussion in #280


maoberlehner:
Found a better solution: monoblok/pull/286

PR #287, opened 2025-09-04
chore: React SDK POC

imranolas:
A reimagining of the Storyblok React SDK to focus purely on rendering and component registration, removing data fetching concerns and letting users handle their own data strategy.

#### Changes
- Removed data fetching hooks (`useStoryblokApi`, `useStoryblokState`)
- SDK now handles only component registration and content rendering through registration with a Bloks/Blok api (inspired by React Router)
- Type safe throughout and integrates well with the types generated by the CLI
- Works with any React setup both client and server side

#### Next.js 15 Demo
- Recreated the Nuxt Storyblok demo site as a React reference implementation
- Next.js 15 with App Router, **SSR**, visual editing, and SWR

### Why is this better?
- Rendering and data fetching are now separate concerns avoid leaky behaviour between SSR and client
- Users can choose their preferred data fetching strategy. This example uses SWR but easily supports others
- A stateless API that eliminates module/global caching 

## Reviewers notes
This PR is a bit of a monster but it broadly falls into the following pieces:
1. React SDK. The changes here leave the API considerably smaller than previously. It's a much simpler API to use and maintain
2. The bulk of the change comes from replacing and relocating the next15 playground app https://github.com/storyblok/monoblok/tree/spike/react-next/apps/next15-demo
3. Along with the demo app I've also committed the generated types for the demo space which makes running and testing considerably easier.

resolves WDX-123


apps/next15-demo/src/app/StoryblokComponent.tsx
```diff
@@ -0,0 +1,134 @@
...
+      <Blok component="contact-form-section" element={ContactFormSection} />
+      <Blok component="price-card" element={PriceCard} />
+      <Blok component="grid-card" element={GridCard} />
+      <Blok component="image-card" element={ImageCard} />
+      <Blok component="richtext-youtube" element={RichtextYoutube} />
+      <Blok component="tabbed-content-entry" element={TabbedContentEntry} />
+      <Blok component="headline-segment" element={HeadlineSegment} />
+      <Blok component="default-page" element={DefaultPage} />
+      <Blok component="nav-item" element={NavItem} />
+    </Bloks>
```
imranolas, line 115: This the crux of the API. It's basically a gloried switch statement. This is user generated code and users can create as many or as few of these as they choose to target different rendering contexts.

apps/next15-demo/src/app/[[...slug]]/page.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  if (slug == null) {
+    slug = "home";
+  }
+
+  if (Array.isArray(slug)) {
+    slug = slug.join("/");
+  }
+
+  const { story, header } = await fetchPage(slug);
```
Comment (imranolas) [inline apps/next15-demo/src/app/[[...slug]]/page.tsx:21]: Async data fetching happens on the server and the result is populated in the SWR cache as fallback. This keeps the cache primed for the subsequent useSWR hook calls.

apps/next15-demo/src/app/[[...slug]]/page.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  return (
+    <SWRProvider
+      fallback={{
+        [`cdn/stories/${slug}`]: story,
+        "cdn/stories/site-config": header,
```
Comment (imranolas) [inline apps/next15-demo/src/app/[[...slug]]/page.tsx:27]: The cache keys simply need to match the request params for useSWR. Similar patterns exist in other data fetching libraries

apps/next15-demo/src/lib/useStory.ts
```diff
@@ -0,0 +1,11 @@
...
+import { ISbStoryData } from "storyblok-js-client";
+import { getStory } from "@/lib/getStory";
+
+const fetcher = (key: string) => getStory(key).then(res => res.data.story);
+
+export const useStory = <T = unknown>(slug: string) => {
+  const { data, error, isLoading } = useSWR(`cdn/stories/${slug}`, fetcher);
+
+  return { story: data as ISbStoryData<T>, error, isLoading };
+}
```
imranolas, line 11: The user creates their own purpose built hook around their chosen data fetching library.

apps/next15-demo/src/app/Story.tsx
```diff
@@ -0,0 +1,23 @@
...
+type Component =
+  | NonNullable<StoryblokComponents.DefaultPage["body"]>[number]
+  | StoryblokComponents.DefaultPage
+  | StoryblokComponents.HeadlineSegment
+  | StoryblokComponents.Button
+  | StoryblokComponents.NavItem
+  | StoryblokComponents.ArticlePage;
+
+export function Story({ slug }: { slug: string }) {
+  const { story, error, isLoading } = useStory<Component>(slug);
```
imranolas, line 16: This is a user hook that wraps the their data fetching library (SWR in this case) using the js client. Since the SWR cache is already pre-populated in the Server Component, this will simply return the data needed to complete SSR.

On the client, the SWR cache is re-hydrated and bridge updates will mutate the cache and trigger live updates.

apps/next15-demo/src/app/components/BridgeHandler.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const bridge = new window.StoryblokBridge({
+      resolveRelations: resolveRelations,
+    });
+
+    bridge.on("input", (event) => {
+      if (!event || !event.story) return;
+      mutate(`cdn/stories/${event.story.slug}`, event.story, {
+        revalidate: false,
+      });
+    });
```
imranolas, line 27: The SWR cache is keyed by slug which is globally unique within the space. When a live update occurs we mutate the cache which triggers a re-render with the updated data. This occurs on the client only.

alexjoverm, line 27:
```suggestion
  const cb = event => mutate(`cdn/stories/${event.story.slug}`, event.story, {
    revalidate: false,
  });
  
  useStoryblokBridge(story.id, cb, { resolveRelations });
```

(would need BridgeHandler to have the `story` from `[[...slug]]/page.tsx` to be passed as property)

PR #292, opened 2025-09-08
fix(react): richtext table and custom attribute casing

### LINKED ISSUE #239: 🐛 [Bug]: RichText table attributes not React conformant
### Package

### Bug Description

In the RichText package, some attributes get applied to the node:

https://github.com/storyblok/monoblok/blob/2e3f2c91a58f85e35240e6eda4d8d0fb0af5d474/packages/richtext/src/richtext.ts#L201-L207

Since we spread them into the component, we get the following errors (not just colspan though):

Image

It is probably a matter of adding all the relevant attributes to the mapper:

https://github.com/storyblok/monoblok/blob/2e3f2c91a58f85e35240e6eda4d8d0fb0af5d474/packages/react/src/utils.ts#L30-L36

### Steps to Reproduce

1. Create RichText field
2. Add table
3. Create resolver for any element, for example table cell
4. Spread node attributes into element
5. Attributes are not React compliant and will cause errors in the console

### Expected Behavior

Node attributes should be React compliant

### Actual Behavior

Node attributes are not React compliant

### Code Sample

```javascript
export function TableHeader(node: StoryblokRichTextNode<ReactElement>) {
	if (node.type !== BlockTypes.TABLE_HEADER) {
		return null;
	}

	return (
		<th className='...' {...node.attrs}>
			{node.children}
		</th>
	);
}
```

### Environment

```shell
System:
    OS: macOS 15.5
    CPU: (11) arm64 Apple M3 Pro
    Memory: 138.55 MB / 36.00 GB
    Shell: 3.7.0 - /opt/homebrew/bin/fish
  Binaries:
    Node: 20.11.0 - ~/.nvm/versions/node/v20.11.0/bin/node
    npm: 10.8.3 - ~/.nvm/versions/node/v20.11.0/bin/npm
    pnpm: 9.4.0 - ~/.nvm/versions/node/v20.11.0/bin/pnpm
    bun: 1.2.8 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 138.1.80.115
    Chrome: 138.0.7204.184
    Safari: 18.5
  npmPackages:
    @storyblok/js: 4.2.0 => 4.2.0 
    @storyblok/react: 5.4.0 => 5.4.0 
    next: 15.3.4 => 15.3.4 
    react: 19.0.0 => 19.0.0 
    storyblok: 3.35.0 => 3.35.0
```

### Error Logs

```shell

```

### Additional Context

_No response_

maoberlehner:
I’ve decided to add as many attributes as possible to the map of HTML attributes that must be converted to camelCase for React. Although only a couple of attributes are supported by default, it is possible to allow custom attributes on links, so in practice, there could be all the double-word HTML attributes. I’ve also added colwidth, although this is not a standard attribute, because the table module seems to set it.

Fixes WDX-82
Fixes #239

### DIFF
--- a/packages/react/src/__tests__/utils.test.ts
+++ b/packages/react/src/__tests__/utils.test.ts
@@ -4,18 +4,23 @@ import { convertAttributesInElement } from '../utils';
 
 describe('utils', () => {
   describe('convertAttributesInElement', () => {
-    it('should convert HTML class attribute to React className', () => {
-      const input = React.createElement('div', { class: 'test-class' });
+    it('should convert HTML various lower case attributes to React camelCase', () => {
+      // The logic is always the same; we test only the most important cases.
+      const someAttributes = {
+        class: ['className', 'test-class'],
+        colspan: ['colSpan', '2'],
+        for: ['htmlFor', 'foo'],
+        rowspan: ['rowSpan', '3'],
+      };
+      const input = React.createElement(
+        'div',
+        Object.fromEntries(Object.entries(someAttributes).map(([attr, [_, value]]) => [attr, value])),
+      );
       const output = convertAttributesInElement(input);
-      expect(output.props).toHaveProperty('className', 'test-class');
-      expect(output.props).not.toHaveProperty('class');
-    });
-
-    it('should convert HTML for attribute to React htmlFor', () => {
-      const input = React.createElement('label', { for: 'test-input' });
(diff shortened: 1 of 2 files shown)


alvarosabu (approved):
Ugh, JSX was an error haha. Thanks for this

PR #295, opened 2025-09-08
feat: stream based migrations

imranolas:
This PR replaces the batch based migration command with a streaming pipeline that provides progress reporting and improved performance. The previous process relied on loading all the stories into memory whereas now we process the stories in flight. This improves both memory usage and throughput.

resolves wdx-67

### DIFF
--- a/packages/cli/src/api.ts
+++ b/packages/cli/src/api.ts
@@ -3,7 +3,9 @@ import { RateLimit } from 'async-sema';
 
 let instance: ManagementApiClient | null = null;
 let storedConfig: ManagementApiClientConfig | null = null;
-const lim = RateLimit(6);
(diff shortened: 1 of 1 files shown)


edvinasjurele:
I have tried running same migration on 2 stories (via the same `--starts-with` so I surely know that there are 2 stories matching criteria), and I got 2 migrations applied, but only 1 pushed 🤷 not quite sure if this is related to above, but it is highly likely.

image

I checked pages in CMS and indeed there was only 1 updated.

Tried with 3 target pages. Only 2 were updated. So there is some pattern to always loose one story.

image

imranolas:
@edvinasjurele It looks like you've found a bug. Please can you open an issue rather commenting on this closed PR, thanks

PR #297, opened 2025-09-09
feat(vue): add support for slots in `<StoryblokComponent>`

### LINKED ISSUE #60: StoryblokComponent not forwarding slots

---

## Expected Behavior

For the lib/StoryblokComponent.vue component to be useful, it would be helpful if we could use slots with the component.

## Current Behavior

Slots are not forwarded through the dynamic component in lib/StoryblokComponent.vue

## Steps to Reproduce

```
<template>
  <StoryblokComponent v-if="story" :blok="story.content">
    <div>Some slot content here</div>
  </StoryblokComponent>
</template>
```

The above code will not render the slot content.

PR to resolve: https://github.com/storyblok/storyblok-vue/pull/263

maoberlehner:
Fixes WDX-144
Fixes #60

### DIFF
--- a/packages/vue/src/components/StoryblokComponent.vue
+++ b/packages/vue/src/components/StoryblokComponent.vue
@@ -41,5 +41,9 @@ if (!componentFound && VueSDKOptions) {
     :is="componentName"
     ref="blokRef"
     v-bind="{ ...$props, ...$attrs }"
-  />
+  >
+    <template v-for="(_, name) in $slots" #[name]="slotProps">
+      <slot :name="name" v-bind="slotProps"></slot>
+    </template>
+  </component>
 </template>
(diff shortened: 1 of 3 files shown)


packages/vue/src/components/StoryblokComponent.vue
```diff
@@ -41,5 +41,9 @@ if (!componentFound && VueSDKOptions) {
     :is="componentName"
     ref="blokRef"
     v-bind="{ ...$props, ...$attrs }"
-  />
+  >
+    <template v-for="(_, name) in $slots" #[name]="slotProps">
```
alvarosabu, line 45: question:

Does this include the default slot as well @maoberlehner ?

PR #318, opened 2025-09-18
fix(cli): creating one rollback file per migration

### LINKED ISSUE #317: Storyblok CLI - Migrations rollbacks process is unclear
### Discussed in https://github.com/storyblok/monoblok/discussions/315

<div type='discussions-op-text'>

<sup>Originally posted by **edvinasjurele** September 17, 2025</sup>
After introducing stream-based migrations (https://github.com/storyblok/monoblok/pull/295), **v4.6.3**, it seems many rollback files are being created, which results in 1 story = 1 file (before it was 1 rollback file for 1 run), which by the way extremely complicates rollback process. 

As documented https://github.com/storyblok/monoblok/blob/main/packages/cli/src/commands/migrations/rollback/README.md rollback command is targetting specific file BY FULL NAME:
```
storyblok migrations rollback MIGRATION_FILE --space YOUR_SPACE_ID
```
which raises us questions of how to rollback the recently executed migration run easily.

image

Moreover, if I we have 2 migrations for same block, f.e.:
```
blogAuthor.description.js
blogAuthor.title.js
```
this results in essentially duplicated stories which seems are identical:

image

1. I’m not entirely sure how the process is intended to work. Could you guide us through the steps to go from point A to B and back to A, especially in scenarios involving multiple stories (i.e., handling multiple files)?

2. Perhaps it would make sense to combine rollbacks into a single file instead of generating multiple files. For instance, in our space of 12k stories, if we run a migration for 7k files, we’ll end up with 7k rollback files, which makes rolling back nearly impossible in such cases.

3. Perhaps the migration command should be improved to target a block, similar to how the migration run works, rather than targeting a specific file. Handling individual files feels tedious and non-revertable. Ultimately, this comes down to the vision or process that you, as the creators of the CLI, want to define and enforce.</div>

maoberlehner:
With the introduction of streaming, the saving of rollback files changed from one rollback file per migration to one rollback file per story and migration (See: https://github.com/storyblok/monoblok/pull/295).

With this fix, we change the behavior back to one rollback file per migration.

To better accommodate the requirements of streaming, we changed the file format for rollbacks from pure JSON to line-separated JSONL. This makes it easier to append to the rollback file before each story migration.

Fixes #317


alvarosabu (approved):
LGTM

PR #325, opened 2025-09-22
feat: support custom prefix `fieldTypePrefix` for generated Storyblok types

rydkvist:
#### What’s new

* Introduces `fieldTypePrefix` option for type generation.
* Allows replacing the default `Storyblok*` prefix with a custom one.

#### Why

* Makes generated types agnostic to Storyblok naming.
* Easier to align with project-specific conventions.

#### Notes

👉 This is a suggestion - if directly replacing identifiers isn’t preferred, an alternative could be to keep the original `Storyblok*` types and re-export prefixed aliases.


rydkvist:
> Thank you for your contribution! I think replacing `Storyblok` is not the best way to go. Instead, we should introduce a dedicated placeholder like `[TYPE_PREFIX]SomeType` (name and format TBD).

I see, I mean to give some perspective, in our project we approach naming for our codebase from an agnostic point of view. So for example for Storyblok types, we tend to use the CMS prefix to be agnostic. 

This PR would help with that, however I understand that this file is also a direct reflection of the Field Types in Storyblok, and it can make sense for the prefix to be `Storyblok` permanently. 

It is also possible to work around this by re-exporting these types as alias in a file within our project, so I mean... yeah no biggie! We can close this PR :)

rydkvist:
Closing the PR based on this [comment](https://github.com/storyblok/monoblok/pull/325#issuecomment-3336128444)

PR #327, opened 2025-09-22
feat: add support for separate type files with `--separate-files` flag

rydkvist:
### **Description**

Introduce the optional `--separate-files` flag to the `types generate` command, giving users more flexibility in how type definitions are structured.

### **What’s changed**

- Added `--separate-files` option to `types generate`
- Default behavior remains unchanged: generates a single `{filename}.d.ts` file
- When enabled, generates one `{componentName}.d.ts` file per component
- Re-used `sanitizeFilename` utility to ensure safe filenames
- Scoped Storyblok imports only into files that require them
- Extended unit tests to cover separate-files mode

### **Notes**

* Default single-file mode is preserved for backwards compatibility
* Future improvement: allow custom filenames for global vs. per-component files


packages/cli/src/commands/types/generate/actions.test.ts
```diff
@@ -815,3 +818,138 @@ describe('generateStoryblokTypes', () => {
...
+
+    const result = await generateTypes(spaceData, { separateFiles: true, strict: false });
+    expect(Array.isArray(result)).toBe(true);
+    const files = result as { name: string; content: string }[];
+    expect(files).toHaveLength(2);
+
+    expect(files[0].content).toContain('// This file was generated by the storyblok CLI.');
+    expect(files[0].content).toContain('// DO NOT MODIFY THIS FILE BY HAND.');
+    expect(files[0].name).toBe('Hero Section X')
+    expect(files.some(f => f.content.includes('export interface HeroSectionX'))).toBe(true);
```
maoberlehner, line 865: nitpick: why did you use `some` here instead of explicitly selecting a file with `files[0]`?

packages/cli/src/commands/types/generate/actions.test.ts
```diff
@@ -815,3 +818,138 @@ describe('generateStoryblokTypes', () => {
...
+
+    // join() receives the sanitized filenames (we only mocked return; assert args)
+    expect(join).toHaveBeenCalledWith(expect.any(String), 'Hero.d.ts');
+    expect(join).toHaveBeenCalledWith(expect.any(String), 'Button.d.ts');
+
+    // Should not try to use the single-file filename
+    expect(join).not.toHaveBeenCalledWith(expect.any(String), 'IGNORED.d.ts');
+  });
+
+  it('default behavior still returns a single string when separateFiles is not set', async () => {
```
maoberlehner, line 934: nitpick: I think this test is unnecessary because we already test that the default behavior works with preexisting tests.

packages/cli/src/commands/types/generate/actions.test.ts
```diff
@@ -815,3 +818,138 @@ describe('generateStoryblokTypes', () => {
...
+    const spaceData: SpaceComponentsData = {
+      components: [withMultilink, simpleText],
+      datasources: [],
+      groups: [],
+      presets: [],
+      internalTags: [],
+    };
+
+    const res = await generateTypes(spaceData, { separateFiles: true, strict: false });
+    const files = res as { name: string; content: string }[];
```
maoberlehner, line 900: nitpick: we should avoid `as` and instead use a type guard.


```diff
-   const res = await generateTypes(spaceData, { separateFiles: true, strict: false });
+   const files = await generateTypes(spaceData, { separateFiles: true, strict: false });
+   if (!Array.isArray(files)) {
+     throw new Error('generateTypes didn't return an array!')
+   }
```

packages/cli/src/commands/types/generate/actions.ts
```diff
@@ -296,133 +296,205 @@ async function loadCompilerOptions(path: string) {
...
+    (acc, [_, v]) => (v && typeof v === 'object' && 'required' in v && (v as any).required ? [...acc, _] : acc),
+    ['component', '_uid'],
+  );
+};
+
+const collectUsedTypesFromProps = (
+  props: JSONSchema['properties'] | undefined,
+  used: Set<string>,
+) => {
+  if (!props) return;
```
maoberlehner, line 310: question: why is this check here and not where this method is called?

packages/cli/src/commands/types/generate/actions.ts
```diff
@@ -296,133 +296,205 @@ async function loadCompilerOptions(path: string) {
...
+const getRequiredFields = (schema: Record<string, any> | undefined): string[] => {
+  return Object.entries(schema || {}).reduce<string[]>(
+    (acc, [_, v]) => (v && typeof v === 'object' && 'required' in v && (v as any).required ? [...acc, _] : acc),
+    ['component', '_uid'],
+  );
+};
+
+const collectUsedTypesFromProps = (
+  props: JSONSchema['properties'] | undefined,
+  used: Set<string>,
```
maoberlehner, line 308: suggestion (non-blocking): In the spirit of pure functions, I'd prefer we create a new set and merge all sets in `buildSchemaForComponent` if this doesn't make the implementation overly complex.

packages/cli/src/commands/types/generate/actions.ts
```diff
@@ -296,133 +296,205 @@ async function loadCompilerOptions(path: string) {
...
+) => {
+  if (!props) return;
+  Object.values(props).forEach((property: any) => {
```
maoberlehner, line 311: question: can we avoid `any` here?

packages/cli/src/commands/types/generate/index.test.ts
```diff
@@ -405,5 +406,51 @@ describe('types generate', () => {
         path: undefined,
       });
     });
+
+    it('handles generateTypes returning an array (separate-files mode) without crashing', async () => {
```
maoberlehner, line 410: suggestion: `without crashing` is a little redundant here. Like, obviously, we expect our code not to crash ;)


```suggestion
    it('should pass separateFiles option to generateTypes when --separate-files flag is used', async () => {
```

packages/cli/src/commands/types/generate/actions.test.ts
rydkvist: Good catch, and thanks for the suggestion

packages/cli/src/commands/types/generate/actions.test.ts
rydkvist: Good catch. I was coding a bit late at night at the time of this commit 😁

packages/cli/src/commands/types/generate/actions.ts
rydkvist: It was to resolve this

rydkvist:
> I like the direction this PR is going! Yet in multiple places, it is still very visible that this feature was added at a later point.
> 
> E.g., there is `const typedefString = await generateTypes` where `typedefString` now can be an array.
> 
> I didn't invest too much thinking time into what that would mean for the code, but just as a general thought, I think it would be nice if there is only one part in the logic where `--separate-files` changes the behavior.
> 
> For example, by always dealing with an array of types and at the place where we save the types, deciding if we save this array in one or multiple files based on the config.

I understand, let me revisit the code and see what I can do 👍🏼

rydkvist:
Hi @alvarosabu and @maoberlehner, please have another look now. I hope it looks better now. 

Fixed a type discrepancy for `path` as well, it was included in the Options type, but in reality it's not a valid option. It's only a valid option on the types command, so I created a type for it (`TypesCommandOptions`)

dipankarmaikap:
Hey @rydkvist , thanks for opening this PR and for your patience. I will take a closer look and see how we can move forward. There is one new detail we need to decide before proceeding. In the latest version, we generate new types for datasources, and all datasource types are now stored in the same single file `storyblok-components.d.ts`. If we want to support separate files, we will need to create a new type file for datasources (eg. `datasources.d.ts`) and then reference those types from the components that use them in the separate `componentName.d.ts` files.

Would you be able to update your PR to account for this and resolve the conflicts?

dipankarmaikap:
Closing this PR in favor of the official PR #406. Since there was no response from the original author, the changes are being continued there.

PR #330, opened 2025-09-24
fix(js-client): `inlineAssets` prefer story meta data

### LINKED ISSUE #328: 🐛 [Bug]: Image data is overridden when `inlineAssets` is `true`
### Package

storyblok-js-client (Client SDK)

### Bug Description

When using the `storyblok-js-client` SDK and setting the option `inlineAssets` to `true`, all information of the current image is overridden. This is problematic, when for instance the alt text is different on the image, that is used in the story. 

If you look into [the implementation](https://github.com/storyblok/monoblok/blob/948ce49676e94dfcbb372f9007885045885f6954/packages/js-client/src/index.ts#L849-L855), you see, that the whole image object is replaced with the master asset object, which might have different fields, then the image, that is used in the story.

### Steps to Reproduce

1. create a Story in Storyblok that has an image included
2. select an image from the Asset Library
3. override the alt text of the image in the story
4. request the story using the `storyblok-js-client` with `inlineAssets: true`.
5. See the alt text from the master image instead of the change alt text

### Expected Behavior

when inlining assets, only properties, that are not present on the image attached to the story should be merged. For instance, the `is_private` property is not present on the attached image. Asset fields like the alt text or the caption should not be overridden. 

### Actual Behavior

Currently, due to the complete override of the image, all information that is changed on the attached image will be overridden from the master image which leads to not translated alt text or captions.

### Code Sample

```javascript
// how it is currently done in your code
if (processedNode.fieldtype === 'asset' && Array.isArray(response.data.assets)) {
  // Replace the assets array with the actual asset objects
  processedNode = {
    ...processedNode,
    // this will completely override all root level fields of the attached asset.
    ...response.data.assets.find((asset: any) => asset.id === processedNode.id),
  };
}
```

### Environment

```shell
System:
    OS: macOS 26.0
    CPU: (14) arm64 Apple M4 Pro
    Memory: 1.12 GB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.19.0 - ~/.nvm/versions/node/v22.19.0/bin/node
    npm: 10.9.3 - ~/.nvm/versions/node/v22.19.0/bin/npm
    pnpm: 10.16.1 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 140.0.7339.186
    Edge: 140.0.3485.81
    Safari: 26.0
  npmPackages:
    @storyblok/astro: ^7.2.5 => 7.3.0 
    @storyblok/richtext: ^3.7.0 => 3.8.0 
    storyblok: ^4.4.1 => 4.6.2 
    vue: ^3.5.21 => 3.5.21
```

### Error Logs

```shell

```

### Additional Context

_No response_

maoberlehner:
When asset meta data is overwritten at the story level, we should prefer and keep the story meta data instead of the global asset meta data.

This is especially important for `alt` texts of images because it should be context specific.

Fixes WDX-165
Fixes #328

### DIFF
--- a/packages/js-client/src/index.ts
+++ b/packages/js-client/src/index.ts
@@ -843,7 +843,7 @@ export class Storyblok {
 
       let processedNode = { ...node };
       if (processedNode.fieldtype === 'asset' && Array.isArray(response.data.assets)) {
-        // Enricht the asset with an actual asset object
+        // Enrich the asset with an actual asset object
         processedNode = {
           ...response.data.assets.find((asset: any) => asset.id === processedNode.id),
           ...processedNode,
--- a/packages/js-client/src/index.ts
+++ b/packages/js-client/src/index.ts
@@ -837,20 +837,16 @@ export class Storyblok {
         return node;
       }
 
-      // Handle arrays
       if (Array.isArray(node)) {
         return node.map(item => processNode(item));
       }
 
-      // Process object
       let processedNode = { ...node };
-
-      // Check if this is an asset field
       if (processedNode.fieldtype === 'asset' && Array.isArray(response.data.assets)) {
-        // Replace the assets array with the actual asset objects
+        // Enricht the asset with an actual asset object
         processedNode = {
-          ...processedNode,
           ...response.data.assets.find((asset: any) => asset.id === processedNode.id),
+          ...processedNode,
         };
       }
 
--- a/packages/js-client/src/index.test.ts
+++ b/packages/js-client/src/index.test.ts
@@ -390,6 +390,118 @@ describe('storyblokClient', () => {
       expect(mockGet).toHaveBeenCalledTimes(1);
     });
 
+    it('should enrich inline assets with data from an asset object', async () => {
+      const story = {
+        data: {
+          story: {
+            id: 123456,
+            uuid: 'story-uuid-123',
+            name: 'Page',
+            slug: 'page',
+            full_slug: 'folder/page',
+            created_at: '2023-01-01T12:00:00.000Z',
+            published_at: '2023-01-02T12:00:00.000Z',
+            first_published_at: '2023-01-02T12:00:00.000Z',
+            content: {
+              _uid: 'content-123',
+              component: 'page',
+              image: {
+                id: 87196701025710,
+                alt: 'story alt',
+                name: 'image',
+                focus: '',
+                title: '',
+                source: '',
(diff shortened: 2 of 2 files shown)


packages/js-client/src/index.ts
```diff
@@ -837,20 +837,16 @@ export class Storyblok {
...
       let processedNode = { ...node };
-
-      // Check if this is an asset field
       if (processedNode.fieldtype === 'asset' && Array.isArray(response.data.assets)) {
-        // Replace the assets array with the actual asset objects
+        // Enrich the asset with an actual asset object
         processedNode = {
-          ...processedNode,
           ...response.data.assets.find((asset: any) => asset.id === processedNode.id),
+          ...processedNode,
```
leifmarcus, line 849: I think, this implementation will not work either. It will always override all properties. But, you want for instance, when the root asset has an alt text and the story asset doesn’t, you don’t want an empty alt text in the story.

So, you should have a more refined way of merging the assets.

maoberlehner: > But, you want, for instance, when the root asset has an alt text and the story asset doesn’t, you don’t want an empty alt text in the story.

Why not? An empty alt text is exactly what you want if the image, in the context of a particular story, is merely representational.

I think it is impossible to come up with a merging algorithm that fulfills every use case. E.g., somebody who wants to remove the alt text for a representational image in a certain story expects the alt text not to get merged back from the global asset object.

With the current change, the behavior is now in line with how it behaves without using the `js-client`. Story settings trump global asset object settings, while properties that do not exist in the story are merged in.

leifmarcus: The problem is, that the Storyblok API is not setting the Alt Text to `undefined`, when it is not set. You actually have no possibility to say, that you want to use the alt text from the root image. 

How would you as a user of the js client handle those problems? 

Let say, you have an alt text on the root image and you want to did set one at the attached image, but then you decide to remove it again (Which happens a lot during content editing) and then you are not able anymore to get the alt text of the root image back. Imagine this for thousands of images, not only a single one.

alvarosabu (approved):
QG ✌️

packages/js-client/src/index.ts
leifmarcus: I guess I understand your point. The JS Client should not fix the API problems. Hope, the Product Team will catch up on this, as I think, we will need to have this feature soon.

Thanks for looking into it and handing it over.

PR #333, opened 2025-09-25
chore: improve login UX with clearer token option and guidance

rydkvist:
### Description
- Renamed "SSO" login option to "Personal Access Token" to avoid confusion
- Added inline CLI hint with link to generate token
- Clarified that tokens work for all accounts, including SSO
- Updated README with clearer token login instructions


maoberlehner (commented):
LGTM what do you think @alvarosabu?

PR #347, opened 2025-10-07
feat: cli-config-file

alvarosabu:
DRAFT of discovery phase

WDX-181

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -1,6 +1,5 @@
 #!/usr/bin/env node
 import dotenv from 'dotenv';
-import { loadConfig } from 'c12';
 
 import { handleError, konsola } from './utils';
 import { getProgram } from './program';
@@ -29,6 +28,7 @@ konsola.br();
 konsola.title(` Storyblok CLI `, colorPalette.PRIMARY);
 
 program.option('--verbose', 'Enable verbose output', false);
+program.option('--global-flag <global-flag>', 'Enable global flag', 'default');
 program.version(pkg.version, '-v, --vers', 'Output the current version');
 program.helpOption('-h, --help', 'Display help for command');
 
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/datasources/pull/index.ts
```diff
@@ -26,7 +29,14 @@ datasourcesCommand
 
     // Command options
     const { space, path } = datasourcesCommand.opts();
-    const { separateFiles, suffix, filename = 'datasources' } = options;
+    const {
+      separateFiles = false,
+      suffix,
+      filename = 'datasources',
```
alvarosabu: nitpick:

Since defaults were added at the commander level, you can remove them from the deconstructions. This probably applies to all the command files where defaults were added

packages/cli/src/utils/object.ts
```diff
@@ -0,0 +1,30 @@
+export type PlainObject = Record<string, any>;
+
+export function isPlainObject(value: unknown): value is PlainObject {
+  return Boolean(value) && typeof value === 'object' && !Array.isArray(value);
+}
+
+export function mergeDeep<T extends PlainObject>(target: T, source?: PlainObject): T {
```
alvarosabu: nitpick:

Normally, I'm more in favor of local solutions, but if we want to avoid the maintenance of this function, `defu` https://github.com/unjs/defu should be available through `c12` and is more battle-tested and performant. If we want to make sure we can explicitly install `defu` 

Otherwise, we should add a unit test for it at least.

packages/cli/src/index.ts
```diff
@@ -16,17 +16,26 @@ import './commands/create';
...
 konsola.title(` Storyblok CLI `, colorPalette.PRIMARY);
 
-program.option('--verbose', 'Enable verbose output');
+// Set all the global config options
+for (const option of GLOBAL_OPTION_DEFINITIONS) {
+  if (option.parser) {
+    program.option(option.flags, option.description, option.parser, option.defaultValue);
+  }
+  else {
+    program.option(option.flags, option.description, option.defaultValue);
```
alvarosabu: blocker:

The CI is broken due to the tsc check on the CLI

Screenshot 2025-11-11 at 11 47 52

Try constraining the default generic parameter from `GlobalOptionDefinition` from 'unknown' to the types required by the commander

PR #349, opened 2025-10-08
fix(cli): skip layout-only fields when generating types

CassWindred:
The type generator currently skips tab properties in the JSON schema. Similar to these, "Groups" as described in the documentation provide no utility beyond organising fields in the Storyblok editor. 

Currently, these properties are all generated as `unknown`, this PR proposes extending the tab-skipping logic to avoid generating unrequired unknown values.


maoberlehner (approved):

maoberlehner:
Thank you for your contribution @CassWindred! Will be merged soon.

PR #358, opened 2025-10-21
fix(cli): retry mechanism when multiple systems query the MAPI

maoberlehner:
When multiple systems hit the MAPI with the same access token, the retry mechanisms quickly gave up. By decreasing the semaphore batch size and increasing the max retries, we make such situations less likely. However, it is still possible to run into limits when many systems query the MAPI with the same access token in parallel.

See the following article for the rational behind the full jitter retry delay: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

### DIFF
--- a/packages/cli/src/commands/migrations/run/index.ts
+++ b/packages/cli/src/commands/migrations/run/index.ts
@@ -114,7 +114,7 @@ migrationsCommand.command('run [componentName]')
           query,
           starts_with: startsWith,
         },
-        batchSize: 100,
+        batchSize: 12,
         onTotal: (total) => {
           storiesProgress.setTotal(total);
           migrationsProgress.setTotal(total);
@@ -141,7 +141,7 @@ migrationsCommand.command('run [componentName]')
         space,
         publish,
         dryRun,
-        batchSize: 100,
+        batchSize: 12,
         onProgress: () => {
           updateProgress.increment();
         },
--- a/packages/mapi-client/src/utils/calculate-retry-delay.ts
+++ b/packages/mapi-client/src/utils/calculate-retry-delay.ts
@@ -1,0 +1,20 @@
+/**
+ * Calculates the delay for the next retry attempt using exponential backoff with full jitter.
+ *
+ * @param attempt The current retry attempt number (starting from 0 for the first retry).
+ * @param baseDelay The initial delay in milliseconds (e.g., 100).
+ * @param maxDelay The maximum possible delay in milliseconds (e.g., 20000).
+ * @returns The calculated delay in milliseconds to wait before the next attempt.
+ */
+export function calculateRetryDelay(
+  attempt: number,
+  baseDelay: number = 100,
+  maxDelay: number = 20000,
+): number {
+  const exponentialBackoff = baseDelay * 2 ** attempt;
+  const cappedBackoff = Math.min(exponentialBackoff, maxDelay);
+  // Apply full jitter: a random value between 0 and the capped backoff
+  const jitter = Math.random() * cappedBackoff;
+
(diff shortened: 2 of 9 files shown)


alexjoverm (changes requested):
LGTM! Seems that the CI/CD is failing though, but other than that looks great

Also pinged you re the batch size, but not a blocker at all.

packages/cli/src/commands/migrations/run/index.ts
```diff
@@ -114,7 +114,7 @@ migrationsCommand.command('run [componentName]')
           query,
           starts_with: startsWith,
         },
-        batchSize: 100,
+        batchSize: 6,
```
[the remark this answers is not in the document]
maoberlehner: So I did go down the rabbit hole of batch sizes, exponential backoffs, and jitter. I arrived at a little bit more sophisticated delay mechanism and a `batchSize` of `12`. The higher the `batchSize`, the higher the `maxRetries` must be, and I think `12` retries already is quite something. With a `batchSize` >= `20`, I quickly run into the `maxRetries` limit when I start two or more migrations in parallel. With the new settings, 4 parallel migrations are still stable.

packages/mapi-client/src/utils/calculate-retry-delay.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Calculates the delay for the next retry attempt using exponential backoff with full jitter.
+ *
+ * @param attempt The current retry attempt number (starting from 0 for the first retry).
+ * @param baseDelay The initial delay in milliseconds (e.g., 100).
+ * @param maxDelay The maximum possible delay in milliseconds (e.g., 20000).
+ * @returns The calculated delay in milliseconds to wait before the next attempt.
+ */
+export function calculateRetryDelay(
```
alexjoverm: praise: this is really sophisticated solution.

I was reading all about it in https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/ and it's super interesting. Thanks!

packages/cli/src/utils/calculate-retry-delay.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Calculates the delay for the next retry attempt using exponential backoff with full jitter.
+ *
+ * @param attempt The current retry attempt number (starting from 0 for the first retry).
+ * @param baseDelay The initial delay in milliseconds (e.g., 100).
+ * @param maxDelay The maximum possible delay in milliseconds (e.g., 20000).
+ * @returns The calculated delay in milliseconds to wait before the next attempt.
+ */
+export function calculateRetryDelay(
```
alexjoverm: question: As this logic already lives in the MAPI client, and the CLI is leveraging it, is it needed to repeat this logic at this level? (same with `delay.ts`)

packages/cli/src/commands/migrations/run/index.ts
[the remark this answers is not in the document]
alexjoverm: This brings the MAPI client and CLI reliance to a next level, awesome!

packages/cli/src/utils/calculate-retry-delay.ts
maoberlehner: Yeah, it's not using the client for the login: https://github.com/storyblok/monoblok/blob/bugfix/cli-migrations-retry/packages/cli/src/commands/login/actions.ts#L38

I don't like the repetition either, but I did want to use the same algo in both cases, yet I don't think it is warranted to introduce a new package for shared code just for that. But maybe we should think about it for the future.

PR #361, opened 2025-10-23
fix(react): update nextjs peer dep and re-structure playgrounds

### LINKED ISSUE #360: 🐛 [Bug]: Next.js v16 peer dependency error
### Package

### Bug Description

Next.js v16 is released, and got peerDependency error when i install @storyblok-react package with Next.js v16.

### Steps to Reproduce

1. Create Next.js project with version 16.
2. Add @storyblok/react to package.json
3. npm install

### Expected Behavior

Should be no error when installing the packages.

### Actual Behavior

Conflicting peer dependency: next@15.5.6

### Code Sample

```javascript

```

### Environment

```shell
Next.js v16 with Node v24
```

### Error Logs

```shell
npm error Could not resolve dependency:
npm error peerOptional next@"^13 || ^14 || ^15" from @storyblok/react@5.4.16
```

### Additional Context

_No response_

alexjoverm:
After investigating, testing and playing with it it, seems adding the peer dep was all needed. 

At the same time, I took the chance to rename the playgrounds and fix a few minor things:
- `next13-app-router` -> `next-e2e-testing` (it was the one used for it, so now it's clearer that it's the one used for testing, and should be changed only when automated tests change, not for manual testing - just as we have it in the Nuxt SDK)
- `next15` -> `next-latest` (so we don't need to rename it in every major release. Also, upgraded to Next 16)
- `next15-static` -> `next-latest-static` (same)

Closes #360

### DIFF
--- a/packages/react/cypress/e2e/index.cy.ts
+++ b/packages/react/cypress/e2e/index.cy.ts
@@ -2,10 +2,6 @@ describe('@storyblok/react/rsc', () => {
   describe('Bridge', () => {
     it('Is loaded by default', () => {
       cy.visit('http://localhost:3000/?_storyblok_tk[timestamp]=1677494658');
-      // Verify we're in an iframe context (which Cypress provides automatically)
-      cy.window().then((win) => {
-        expect(win.self).to.not.equal(win.top);
-      });
       cy.get('#storyblok-javascript-bridge').should('exist');
     });
   });
--- a/packages/react/playground/next-13-app-router/package.json
+++ b/packages/react/playground/next-13-app-router/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@storyblok/playground-react-next-e2e-testing",
+  "name": "@storyblok/playground-next-13-app-router",
   "type": "module",
   "version": "0.1.0",
   "private": true,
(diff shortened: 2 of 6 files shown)


alexjoverm:
@maoberlehner I see your point, and that makes sense. If we intend them to be doubled-role, then let's remove the testing part 😄 . 

I went ahead and renamed them to `-13` and `-latest`. I also added the latest to the test suite, so now we have both 13 and 16 in the test suite. Let me know what you think

packages/react/playground/next-latest/src/app/components/Teaser.tsx
```diff
@@ -1,9 +1,6 @@
 import { storyblokEditable } from '@storyblok/react/rsc';
-import { headers } from 'next/headers';
```
alexjoverm: @maoberlehner note for the review:

I'm unsure about what was the intention of this here. It made the playground fail (both Next 15 and 16) as soon as you wanted to use the bridge for live editing, as for that you need the `StoryblokProvider` pattern, and that makes the component a client component.

I also realized that the playground didn't include StoryblokProvider, and bridge wasn't enabled. I updated to include that, according to how it's documented https://www.storyblok.com/docs/guides/nextjs

PR #369, opened 2025-11-07
fix(richtext)!: HTML and Markdown parsing for links and edge cases

maoberlehner:
Parsing did not work correctly for certain combinations of `<a>` and nested tags inside or around it.

To ensure parsing is aligned with the Tiptap editor format, we now use the Tiptap editor's `generateJSON` command, which gives us HTML-to-richtext conversion out of the box.

BREAKING CHANGE: Instead of our own resolver format, we now allow users to override the Tiptap extensions used for parsing.

Fixes WDX-141

---

> [!NOTE]
> Replaces custom HTML/Markdown parsing in `packages/richtext` with Tiptap-based parsing (via `@tiptap/html` + extensions), updates tests, and refreshes frameworks/deps across the repo.
> 
> - **Richtext parsing (BREAKING)**:
>   - Replace custom HTML/Markdown parsers with Tiptap JSON generation using `@tiptap/html` and a full set of Tiptap extensions.
>   - Remove legacy parsing deps (`node-html-parser`, `markdown-it-github`), add Tiptap packages.
>   - Update unit tests for new parsing behavior.
>   - Parsing customization now via Tiptap extensions instead of previous resolver format.
> - **Tooling/Deps**:
>   - Bump React to `19.2.0`, Vue to `3.5.22`, Next-related deps, and Nuxt/Vite stacks in examples/playgrounds.
>   - Add `happy-dom` to Vitest where used and perform miscellaneous lockfile upgrades.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f128bb627b2a7403a07ee8f27601acf2c6b37e68. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/richtext/package.json
+++ b/packages/richtext/package.json
@@ -99,7 +99,8 @@
     "@tiptap/extension-text-style": "^3.10.2",
     "@tiptap/extension-underline": "^3.10.2",
     "@tiptap/html": "^3.10.2",
-    "markdown-it": "^14.1.0"
+    "markdown-it": "^14.1.0",
+    "storyblok-js-client": "workspace:*"
   },
   "devDependencies": {
     "@arethetypeswrong/core": "^0.18.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1024,6 +1024,9 @@ importers:
       markdown-it:
         specifier: ^14.1.0
         version: 14.1.0
+      storyblok-js-client:
+        specifier: workspace:*
+        version: link:../js-client
     devDependencies:
       '@arethetypeswrong/core':
         specifier: ^0.18.2


packages/richtext/playground/vanilla/public/test.html
```diff
@@ -38,19 +123,62 @@ <h2 id="heading-2">Heading 2</h2>
...
 </table>
+
+<!-- ── Details / Accordion ───────────────────────────────────────────── -->
+
+<details>
+  <summary>Click to expand</summary>
+  <p>Hidden content inside a details element.</p>
+</details>
+
+<!-- ── Realistic CMS migration (Porsche-like) ────────────────────────── -->
```
maoberlehner: Maybe consider removing the brand name here. Might be an issue because of some legalese twists, when not now, then maybe at some point when Porsche might not be a customer anymore. Probably not worth the risk.

packages/richtext/src/html-parser.test.ts
```diff
@@ -1023,15 +774,605 @@ describe('htmlToStoryblokRichtext', () => {
...
+      expect.arrayContaining([
+        expect.objectContaining({ type: 'bold' }),
+        expect.objectContaining({ type: 'italic' }),
+        expect.objectContaining({ type: 'underline' }),
+      ]),
+    );
+    expect(result.content[17]).toMatchObject({ type: 'details' });
+    expect(result.content[17].content[0]).toMatchObject({ type: 'detailsSummary' });
+    expect(result.content[17].content[1]).toMatchObject({ type: 'detailsContent' });
+  });
```
maoberlehner: Considering that we now rely on tiptap itself to do the heavy lifting, I wonder if we can drastically reduce the coverage here and focus on our custom stuff only (classes, image optimization, etc.)?

On the other hand, the tests are already here, so maybe they don't hurt either. I leave it to you.

maoberlehner:
I did a couple of tests with various ways of nesting different elements, and overall, everything worked fine. Only two minor findings/considerations:

## 1. null attributes

```ts
const doc = {
  "type": "doc",
  "content": [
    {
      "type": "paragraph",
      "content": [
        {
          "type": "text",
          "marks": [
            {
              "type": "link",
              "attrs": {
                "href": "/",
                "uuid": null,
                "anchor": null,
                "target": null,
                "linktype": "url"
              }
            }
          ],
          "text": "Hello"
        }
      ]
    }
  ]
}
richTextResolver().render(doc)
// <p><a uuid="null" target="null" href="/">Hello</a></p>
```

## 2. resolvers vs tiptapExtensions

I guess it's useful for backward compatibility to keep both, yet I think it is redundant and potentially confusing. E.g., which one wins in the following example? Maybe we can keep resolvers but deprecate them?

```ts
richTextResolver({
  resolvers: {
    [MarkTypes.LINK]: (node) => {
      return `<a href="${node.attrs.href}">${node.text}</a>`;
    },
  },
  tiptapExtensions: {
    link: StoryblokLink,
  },
}).render(x)
```

alexjoverm:
@maoberlehner no worries, it's a great suggestion! I share the vision of using the already provided react renderer, and going fully idiomatic per framework, reducing our custom code.

That said, for this PR specifically I'd prefer to keep the current approach because:

- it would mean maintaining two rendering approaches (React via Tiptap, and the framework-agnostic one)
- we still have custom logic deferring from TipTap, and quite some things to move around to change the approach for React
- as this PR originated from a high-prio customer issue and has grown into a major richtext revamp already, adding an internal architecture change on top would increase the scope and time considerable.
 
Even though AI can accelerate the refactor, this kind of internal architecture change still needs thorough review and iteration before shipping to production, so it's not something we'd rush. But definitely something we can revisit later. Would you be ok with tackling the `renderToReactElement` refactor as a separate effort?

packages/astro/playground/shared/storyblok/Teaser.astro
```diff
@@ -1,17 +1,28 @@
...
 ---
 
 <div {...storyblokEditable(blok)}>
   <h2 class="heading">
     {blok.headline}
   </h2>
-  <div set:html={renderedRichText} />
+
+  {
+    richTextSegments.map((segment, index) => {
```
alexjoverm: @dipankarmaikap with the current logic, how can a user override a mark or a node?

For example, on Vue or React, often the user wants to override a `link` to render to a `RouterLink` or a ReactRouter or Next.js `Link` component, instead of a plain `<a>` tag.

I'm not sure if Astro has their own Link component for routing, but in any case, sometimes users want to override default marks or nodes and use components instead. 

How would they do it here? Can you add an override example here in the playground? For inspiration, you can see the link or heading override in the Next.js playground https://github.com/storyblok/monoblok/pull/369/changes?mode=single#diff-b31212b3df5bff9f904df8f375b02b0018045aeb174c18272c7f007a6720af3fR13-R21

dipankarmaikap (approved):

PR #382, opened 2025-11-24
feat(cli): add reporter for migrations

maoberlehner:
See: WDX-190


packages/cli/src/utils/reporter.ts
```diff
@@ -0,0 +1,157 @@
+import { saveToFileSync } from './filesystem';
```
alexjoverm: nitpick (non-blocking):

Personally, I’d avoid placing the reporter under utils, since it’s more of a core component/service than a util or helper (similar with Logger). Following DDD principles, I'd rather put it in its own `src/reporter` folder, or `src/lib/reporter`

alexjoverm (commented):
Great job, @maoberlehner ! I ran all kind of tests and they work as expected. They're kept under 10, no issues generating it with 1000+ stories migrations, etc.

The only thing I'm unsure is a specific case. When you apply a migration that _skips_ all stories (but none failed), it give the `FAILURE` status. I'm unsure if that actually should be `SUCCESS`, as none failed. But it's going to depend on the definition of `SUCCESS` or `FAILURE` that we want to give it.

What do you think? I'll wait for you answer before approving

--- How to reproduce above edge case ---

Create a migration for a component which doesn't change the content of it (so it'll skip applying it), run it, and you should have a similar report to this:

```json
{
  "status": "FAILURE",
  "meta": {  ... },
  "summary": {
    "migrationResults": {
      "total": 128,
      "succeeded": 0,
      "skipped": 128,
      "failed": 0
    },
    "updateResults": {
      "total": 0,
      "succeeded": 0,
      "failed": 0
    }
  }
}
```

PR #385, opened 2025-11-25
feat(cli): allow storyblok create to use an existing space

dipankarmaikap:
Add `--token` Option, Refactor Success Messaging, and Fix `--skip-space` in `create` Command

### Features
- **New `--token` option:**  
  Users can now pass an access token directly via `--token`. If provided, space creation is skipped and the `.env` file is updated with this token.
- **Environment update:**  
  When using `--token`, the CLI writes the token to the `.env` file for the generated project.

### Bug Fixes
- **Fixed `--skip-space` option:**  
  The `--skip-space` flag now reliably skips space creation and related steps, as verified by new and updated tests.

### Refactoring
- **Success messaging:**  
  Refactored repeated success and "next steps" messages into a helper function to avoid duplication and ensure consistent output.
- **Error handling:**  
  Centralized `.env` file creation error handling for both space and key flows.

### Tests
- **Added test for `--token` option:**  
  Ensures that when `--token` is provided:
  - Space creation and browser opening are skipped.
  - `.env` is updated with the provided key.
  - Success and next steps messages are shown.
- **Improved tests for `--skip-space`:**  
  Verified that space creation and related prompts are skipped, and messaging is correct.

### DIFF
--- a/packages/cli/src/commands/create/index.test.ts
+++ b/packages/cli/src/commands/create/index.test.ts
@@ -132,8 +132,8 @@ const createMockUser = (overrides: Partial<StoryblokUser> = {}): StoryblokUser =
 });
 
 describe('createCommand', () => {
-  describe('--key option', () => {
-    it('should use provided key, skip space creation, and update env', async () => {
+  describe('--token option', () => {
+    it('should use provided token, skip space creation, and update env', async () => {
       vi.mocked(generateProject).mockResolvedValue(undefined);
       vi.mocked(createEnvFile).mockResolvedValue(undefined);
       vi.mocked(fetchBlueprintRepositories).mockResolvedValue([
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/create/index.test.ts
```diff
@@ -672,6 +699,23 @@ describe('createCommand', () => {
       expect(createEnvFile).not.toHaveBeenCalled();
       expect(openSpaceInBrowser).not.toHaveBeenCalled();
     });
+
+    it('should NOT prompt for space creation when --skip-space is provided', async () => {
```
alexjoverm: praise: nice extra test!

packages/cli/src/commands/create/index.ts
```diff
@@ -23,12 +46,13 @@ export const createCommand = program
   .option('-t, --template <template>', 'technology starter template')
   .option('-b, --blueprint <blueprint>', '[DEPRECATED] use --template instead')
   .option('--skip-space', 'skip space creation')
+  .option('--key <key>', 'Storyblok access token (skip space creation and use this token)')
```
alexjoverm: Same as above (naming comment)

alexjoverm (approved):

PR #389, opened 2025-11-28
chore(cli): remove mention of deprecated `blueprint` flag from docs

dipankarmaikap:
1. The `template` flag should be used instead, so we've replaced all mentions of `blueprint` in the docs with the new `template` flag.
2. Added the missing `pnpm release` command in the repos root package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "scripts": {
     "monoblok": "node tools/monoblok/dist/index.mjs",
     "prepare": "husky",
+    "release": "node tools/release.mjs",
     "lint": "nx run-many --target=lint",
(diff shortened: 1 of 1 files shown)


maoberlehner:
> > Why was this change made?
> 
> @maoberlehner I updated the README for the CLI create command. There are no functional changes, and no release is needed. I also added the missing release command to the `package.json`, which does not require a version bump either. If this is not the ideal way to handle it, let me know.

I meant the blueprint to template change. I couldn't find any information on why we want to make this change.

dipankarmaikap:
This was already in the codebase. The deprecation warning comes from the existing Blueprint usage, not from my changes. The official docs hadn’t been updated, so I updated them to match the current API.

PR #398, opened 2025-12-17
feat(cli): add support for region flag to the create CLI command

dipankarmaikap:
This allows users to specify the space region during creation and fully replaces the storyblok create demo CLI command.
Fixes WDX-223

---

> [!NOTE]
> Introduces region-aware project scaffolding and environment setup.
> 
> - Adds `-r, --region` option to `create` command with validation (`isRegion`) and early mismatch checks against account region (for space creation only)
> - New `handleEnvFileCreation` helper and updated `createEnvFile` (now accepts key-value map) to write `STORYBLOK_DELIVERY_API_TOKEN` and `STORYBLOK_REGION` to `.env`; logs success/fail via `konsola`
> - Applies region when: using `--token`, `--skip-space` (region only), or after space creation (session region)
> - Enhances `generateSpaceUrl`/`openSpaceInBrowser` to cover all regions (`eu`, `us`, `cn`, `ca`, `ap`); adds tests
> - Updates Next template URL to `blueprint-core-nextjs`; expands blueprint fetching/mapping and tests (incl. port extraction)
> - Broad test updates for region workflows, error paths, and path handling
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9d3209ac4269ec8de51f53a2915afd35fe104daf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/cli/src/commands/create/index.test.ts
+++ b/packages/cli/src/commands/create/index.test.ts
@@ -1244,5 +1244,68 @@ describe('createCommand', () => {
       expect(createSpace).not.toHaveBeenCalled();
       expect(handleEnvFileCreation).toHaveBeenCalledWith(expect.any(String), undefined, 'us');
     });
+
+    describe('session region fallback behavior', () => {
+      it('should use US session region as fallback when no --region provided with --token', async () => {
+        // Mock session to return US region
+        const { session } = await import('../../session');
+        const mockSession = session();
+        mockSession.state.region = 'us';
+
+        vi.mocked(generateProject).mockResolvedValue(undefined);
+        vi.mocked(handleEnvFileCreation).mockResolvedValue(true);
+        vi.mocked(fetchBlueprintRepositories).mockResolvedValue([
+          { name: 'React', value: 'react', template: '', location: 'https://localhost:5173/', description: '', updated_at: '' },
+        ]);
+
+        await createCommand.parseAsync(['node', 'test', 'my-project', '--template', 'react', '--token', 'my-token']);
+
+        expect(handleEnvFileCreation).toHaveBeenCalledWith(expect.any(String), 'my-token', 'us');
+
+        // Reset to EU for other tests
+        mockSession.state.region = 'eu';
+      });
+
+      it('should use CA session region as fallback when no --region provided with --skip-space', async () => {
+        // Mock session to return CA region
+        const { session } = await import('../../session');
+        const mockSession = session();
+        mockSession.state.region = 'ca';
+
+        vi.mocked(generateProject).mockResolvedValue(undefined);
+        vi.mocked(handleEnvFileCreation).mockResolvedValue(true);
+        vi.mocked(fetchBlueprintRepositories).mockResolvedValue([
+          { name: 'React', value: 'react', template: '', location: 'https://localhost:5173/', description: '', updated_at: '' },
+        ]);
+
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/create/index.ts
```diff
@@ -47,13 +51,21 @@ export const createCommand = program
...
     konsola.title(`${commands.CREATE}`, colorPalette.CREATE);
     // Global options
     const verbose = program.opts().verbose;
     // Command options - handle backward compatibility
     const { template, blueprint, token } = options;
 
+    if (options.region && !isRegion(options.region)) {
+      handleError(new CommandError(`The provided region: ${options.region} is not valid. Please use one of the following values: ${Object.values(regions).join(' | ')}`));
+      return;
+    }
```
maoberlehner: I think this error message is super helpful in case people enter an invalid region!

maoberlehner (changes requested):
Just realized that we should add a test for this functionality.
In case you implement the functionality Cursor recommended, it would be great to have separate test cases for each possible scenario.

christianzoppi (commented):
Hey @dipankarmaikap, thanks for working on this. I think that the PR looks good, and I just second what Markus said about the tests. 

From what I understand, before this PR even if you created a project with a space in a different region (that depends on the region that you logged in into), the codebase would take the access token from the space but it would always use the `eu` API domain, right? 
Should we also change the behavior when a space is created, so we set the region env variable for the codebase with the same region as the space in case the `region` parameter is not manually defined?

alexjoverm (changes requested):
Seems some of the cases are not fully working. In general, I recommend to have them well specified beforehand, so implementation and review goes smooth. 

Here are my findings, I'm building the expected behavior from the info on the internal ticket and the decisions taken on the way:

| Case | Expected behavior | Current behavior |
  | --- | --- | --- |
  | `--token <token>` | Bootstraps the code. Env file is created with the token provided, and region is taken from user session. | ⚠️ Region is `undefined` instead of session region. [`index.ts:166`](https://github.com/storyblok/monoblok/pull/398/changes#diff-e85db3b9733aa67073e3397b203523dcd174b6a7b2ee32e2ef6e53933e9b9451R166) passes `options.region` which is undefined. |
  | `--token <token> --region <region>` | Bootstraps the code. Env file is created with both token and region provided. If the provided region is different from user session one, that's fine as a user can bootstrap a code in different session than logged in. | 👍 works as expected |
  | `--skip-space` | Bootstraps the code. Env file is created with region taken from user session. | ⚠️ No `.env` file is created. [`index.ts:170-177`](https://github.com/storyblok/monoblok/pull/398/changes#diff-e85db3b9733aa67073e3397b203523dcd174b6a7b2ee32e2ef6e53933e9b9451R170-R174) only calls `handleEnvFileCreation` when `options.region` is provided. |
  | `--skip-space --region <region>` | Bootstraps the code. Env file is created with region provided. If the provided region is different from user session one, that's fine as a user can bootstrap a code in different session than logged in. | 👍 works as expected |
  | `--region <region>` | Bootstraps the code + creates a space. Env file is created with the region provided. If the provided region is different from user session one, it **errs as a space can't be created in a different region than the one logged in**. | 👍 works as expected |
  | no region, skip-space or token flags set | Bootstraps the code + creates a space. Env file is created with the region taken from user session. | 👍 works as expected |

PR #400, opened 2025-12-20
feat(cli): add assets pull/push commands

maoberlehner:
Fixes WDX-134
Fixes WDX-216
Fixes WDX-224
Fixes WDX-225

### DIFF
--- a/packages/cli/src/commands/assets/streams.ts
+++ b/packages/cli/src/commands/assets/streams.ts
@@ -1,11 +1,9 @@
 import { Buffer } from 'node:buffer';
 import { basename, extname, join, parse } from 'node:path';
-import { randomUUID } from 'node:crypto';
-import { tmpdir } from 'node:os';
 import { Readable, Transform, Writable } from 'node:stream';
 import { Sema } from 'async-sema';
-import { mkdir, readdir, readFile, stat, unlink, writeFile } from 'node:fs/promises';
-import { appendToFile, saveToFile } from '../../utils/filesystem';
+import { readdir, readFile, stat, unlink } from 'node:fs/promises';
+import { appendToFile, fileExists, saveToFile } from '../../utils/filesystem';
 import { toError } from '../../utils/error/error';
 import type { RegionCode } from '../../constants';
 import { createAsset, createAssetFolder, downloadAssetFile, downloadFile, fetchAssetFolders, fetchAssets, sha256, updateAsset, updateAssetFolder } from './actions';
@@ -13,7 +11,7 @@ import type { Asset, AssetCreate, AssetFolder, AssetFolderCreate, AssetFolderMap
 import { mapiClient } from '../../api';
 import { handleAPIError } from '../../utils/error/api-error';
 import { FetchError } from '../../utils/fetch';
-import { getAssetBinaryFilename, getAssetFilename, getFolderFilename, isRemoteSource } from './utils';
+import { getAssetBinaryFilename, getAssetFilename, getFolderFilename, getSidecarFilename, isRemoteSource } from './utils';
 
 const apiConcurrencyLock = new Sema(12);
 
@@ -438,7 +436,7 @@ export interface LocalAssetPayload {
   context: {
     fileBuffer: ArrayBuffer;
     assetBinaryPath: string;
-    metadataFilePath?: string;
+    assetPath?: string;
   };
 }
 
@@ -478,7 +476,7 @@ export const readLocalAssetsStream = ({
             context: {
               fileBuffer,
               assetBinaryPath,
-              metadataFilePath: filePath,
+              assetPath: filePath,
             },
           } satisfies LocalAssetPayload;
         }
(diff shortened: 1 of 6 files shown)


alexjoverm (changes requested):
Sent an internal message with feedback on the approach - blocking the merge for now until discussion.

packages/cli/src/commands/assets/pull/index.test.ts
alexjoverm: question (non-blocking): are these endpoint responses something we can automated by leveraging the open-api specs (either now, or in the future)

alexjoverm: suggestion: @maoberlehner I think this utils, as long as they're not strictly specific for this test, can be centralized and used across the tests and src. In that way, we don't only avoid duplication, but also ensure tests and src handle it the same way. 

From a quick analysis, this is what was found:

 `getFolderFilename` - This is duplicated:
  - assets/streams.ts:24 (production) - identical logic
  - push/index.test.ts:240 - same pattern inline: ${folder.name}_${folder.uuid}.json

  Asset filename pattern (${name}_${id}${ext}) is used in 3 places:
  - streams.ts:145 - production
  - pull/index.test.ts:162 - test assertion
  - push/index.test.ts:186 - test helper

  getAssetBasename vs getAssetNameAndExt:
  - The test version uses regex: short_filename.match(/(\.[^.]+)$/)
  - The production version in streams.ts:17 uses Node's extname() and basename() with sanitizeFilename()

packages/cli/src/commands/assets/pull/index.ts
```diff
@@ -25,7 +25,7 @@ assetsCommand
   .command('pull')
   .option('-d, --dry-run', 'Preview changes without applying them to Storyblok')
   .option('-p, --path <path>', 'base path to store assets (default .storyblok)')
```
alexjoverm: suggestion: as this option is common to both pull and push, shall we move it up to the root `assets/command.ts`?

packages/cli/src/commands/assets/actions.ts
alexjoverm: suggestion: @maoberlehner AFAIK this function is to fetch **only** external/remote assets, correct? If so:
- Shall we better name it `fetchExternalAsset` or similar?
- Would be possible to add more into to the error message? (code, reason, etc)

packages/cli/src/commands/assets/push/actions.ts
alexjoverm: Better use the wrapping `readFile` function we have in `utils/filesystem.ts`

packages/cli/src/commands/stories/push/actions.ts
```diff
@@ -1,6 +1,12 @@
 import { readFile } from 'node:fs/promises';
 
-export interface ManifestEntry { old_id: string | number; new_id: string | number; created_at?: string }
+export interface ManifestEntry {
```
alexjoverm: @maoberlehner I actually was going to suggest same as below's copilot's comment. Can we centralize this function somewhere? (maybe utils/filesystem)

packages/cli/src/commands/assets/streams.ts
alexjoverm: suggestion (non-blocking): in the same way we have `writeLock` utils centralized in `utils/filesystem.ts`, can we centralize read locks as well?

packages/cli/src/commands/assets/actions.ts
alexjoverm: I just realized this is used also for pulling assets from a Storyblok space. If so, should we use mapiClient instead of a plain fetch? Or if it's used for both internal and externals, maybe we can make it conditional?

packages/cli/src/commands/assets/pull/index.test.ts
```diff
@@ -23,6 +24,15 @@ vi.mock('../../../session', () => ({
   })),
 }));
 
+vi.mock('../../../lib/config/store', async (importOriginal) => {
```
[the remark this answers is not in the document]
maoberlehner: I've tried to use `setActiveConfig` before but the `beforeEach` in your example runs **before** the `preAction` hook:

```ts
programInstance.hook('preAction', async (thisCommand, actionCommand) => {
  // ...
  setActiveConfig(resolvedConfig); // <-- runs after `beforeEach` and overrides the config.
```

packages/cli/src/commands/assets/actions.ts
maoberlehner: Technically it's completely independent from assets, just a helper to download a file. I don't think it's right for the `mapiClient`. Maybe I should name it `downloadFile`?

I will definitely improve the error message, very good point!

alexjoverm: Ah makes sense! I agree, `downloadFile` would sound more descriptive to me as well

packages/cli/src/commands/assets/pull/index.test.ts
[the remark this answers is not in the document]
alexjoverm: Oh good point. Yeah... and likely the `preAction` hook runs just when we run the actual CLI command 😅

packages/cli/src/commands/assets/pull/index.ts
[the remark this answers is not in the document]
maoberlehner: I agree but as this affects the rest of the CLI too, and this PR already is quite extensive I'd prefer to leave it as is for now if you don't mind?

PR #404, opened 2026-01-08
feat(astro)!: astro live preview now support custom payload

### LINKED ISSUE #95: Simplify getLiveStory(Astro) usage by avoiding full Astro object dependency
### Description

As a developer,
I want to simplify the usage of getLiveStory(Astro) by avoiding the need to pass the full Astro global object,
so that utilities remain slim and modular, and I can use them outside of .astro files without excessive Astro-specific dependencies.

Currently, the function only accesses Astro.locals ([code reference](https://github.com/search?q=repo%3Astoryblok%2Fstoryblok-astro%20locals&type=code)), making the full Astro object unnecessary. This adds friction when moving logic to external files or creating reusable utilities, where importing the full Astro context feels excessive and tightly coupled to .astro files.

### Suggested solution or improvement

A more efficient and flexible approach would be to simplify the API to accept only what's needed - either `Astro.locals` directly or wrapped in an object like `getLiveStory({ locals })`. This reduces boilerplate, improves portability, and prepares the utility for future options by using a more scalable argument shape:

```js
// Option 1 (minimalist & intuitive)
getLiveStory(Astro.locals);

// Option 2 (scalable pattern for future growth)
getLiveStory({ locals: Astro.locals });
```
This change:
- Keeps the utility lean and portable.
- Makes it easier to use outside `.astro` files (e.g., in custom logic or helper modules).
- Future-proofs the function in case other parameters (like headers, params, or flags) are needed later, thanks to the object-based signature.

### Additional context

_No response_

### Validations

- [x] Follow our [Code of Conduct](https://www.storyblok.com/trust-center#code-of-conduct?utm_source=github.com&utm_medium=readme&utm_campaign=storyblok-js-client)

### LINKED ISSUE #96: Improve Live Preview by supporting custom payload data passing
### Description

As a developer, I want to provide content editors with a fast and seamless live preview experience - ideally with response times under 300–500ms - so that interactions feel smooth and responsive. A fast feedback loop is crucial for maintaining editor confidence and improving content workflow efficiency.
However, the Astro application is an MPA and performs several server-side operations on each request. These include fetching referenced Storyblok content, pricing information, YouTube metadata, site configuration, and experiment data from third-party APIs. As a result, during live preview, every content change triggers a full server-side reprocessing of the page - leading to delays of up to 7–8 seconds or more. In many cases, this means executing 10+ requests just to render a single page from scratch.

This significantly undermines the purpose of live preview, making the feature feel slow and unreliable. It also increases the load on the server and creates a poor content editing experience. A more efficient live preview mechanism - possibly one that avoids unnecessary refetching or allows partial updates - would greatly improve usability and system performance.

### Suggested solution or improvement

We are aware that the Storyblok bridge sends a POST request to trigger live preview processing, and the request body includes the story object ([code reference](https://github.com/storyblok/storyblok-astro/blob/354113503d954b23d9b5b7ecf36fe6f842304e27/src/live-preview/handleStoryblokMessage.ts#L69)), which is a great starting point:

```js
const result = await fetch(location.href, {
    method: 'POST',
    body: JSON.stringify({
      ...story,
      is_storyblok_preview: true,
    }),
    headers: {
      'Content-Type': 'application/json',
    },
  });
```
Unfortunately, the live preview is only fast on lightweight pages that rely solely on the current page's story data and do not require additional server-side operations.

As mentioned earlier, the story data alone is often not enough, as the server needs to fetch additional data. To address this, we propose including more data in the POST request payload. **By sending the required data along with the request, the server doesn't have to make additional fetch calls. This approach helps maintain the stateless nature of the application, eliminating the need for additional caching solutions.** As a result, it would significantly speed up the live preview feature, even for more complex, data-heavy pages.

I’ve illustrated the idea in the picture:

Image

By including all necessary data in the POST payload, we have successfully achieved near-instant live preview speeds, and we would like to see this improvement incorporated into the package, so the entire community can benefit from it - even those with significant server-side data fetching needs.

https://github.com/user-attachments/assets/15bc4164-498b-4dec-9728-a3415a4de4fe

The biggest problem is that the extra data needed for live preview is unknown to Storyblok and is typically specific to the SDK consumer’s application. There should be a mechanism in place that allows users to provide a webhook or function to pass custom additional data to the Storyblok bridge’s live preview feature. This would make it possible to inject the necessary data dynamically and improve the live preview experience.

### Additional context

There was some discussions on Discord about the live preview speeds: https://discord.com/channels/700316478792138842/1362367702345322616/1362367702345322616

In our custom solution, we have made a `LivePreviewBridge.astro` component that does all what is needed. Schematically:

```js
---
const { serversData, siteConfig, youtubeData } = Astro.locals;
---

<>
  <script>
    import { loadStoryblokBridge } from '@storyblok/js';

    // Add StoryblokBridge script on the page
    await loadStoryblokBridge();
  </script>
  <script
    define:vars={{
      serversData,
      siteConfig,
      youtubeData,
      // more custom data
    }}
  >
    document.addEventListener('DOMContentLoaded', () => {
      window.storyblokRegisterEvent(() => {
        // Create StoryblokBridge instance when the script is loaded
        const storyblokBridge = new window.StoryblokBridge();

        // ...

        const handleInput = async (story) => {
          // Request updated HTML
          const newHTMLResponse = await fetch(location.href, {
            method: 'POST',
            body: JSON.stringify({
              story,
              locals: {
                serversData,
                siteConfig,
                youtubeData,
                // more custom data
              },
            }),
            headers: { 'Content-Type': 'application/json' },
          });
        };

        // ...
        // handle newHTMLResponse
        // ...

        // Update page on change in Live Preview mode
        storyblokBridge.on('input', async ({ story }) => handleInput(story));

        // Reload page on publish and save
        storyblokBridge.on(['published', 'change'], (event) => {
          document.dispatchEvent(new Event('livePreview:changed'));
          if (event.slugChanged) return;
          location.reload(true);
        });
      });
    });
  </script>
</>
```

The code above highlights the key part of dynamically passing custom data via Astro script `define:vars`, which allows injecting custom data into the client script. This is not intended as a solution, but rather to share how we approached and implemented the solution.

### Validations

- [x] Follow our [Code of Conduct](https://www.storyblok.com/trust-center#code-of-conduct?utm_source=github.com&utm_medium=readme&utm_campaign=storyblok-js-client)

### LINKED ISSUE #97: Expose toggle to control live preview behavior dynamically
### Description

As a developer, I want to allow content editors to enable or disable live preview dynamically via a toggle in the editor application, so they can control whether changes are pushed in real time based on their workflow needs.

### Suggested solution or improvement

Currently, we’ve implemented a custom loader icon and a toggle to enable or disable live preview, giving our editors control over when live preview is active.

Image

I am sure this is debatable, if it is up to you to decide whether you want to provide such functionality.

Whether this toggle should be part of the official SDK is open for discussion. While we expect to enable it by default in the long run, during the initial rollout of this feature, we wanted to give content editors the ability to control it themselves. This approach allows them to disable live preview on the fly if something doesn't work as expected. Currently, however, setting livePreview: true in the SDK configuration:
```
    storyblok({
      accessToken: "OsvN..",
      livePreview: true,
    }),
```

the SDK will automatically POST every change back to your application — so toggling live preview on and off dynamically isn’t supported by the official package.

Alternatively, the ability to disable live preview could be controlled via event dispatching (e.g., `window.dispatchEvent`) or by setting a global flag on the client side, such as `window.StoryblokBridge?.livePreview = false;`, if such API is exposed.

### Additional context

_No response_

### Validations

- [x] Follow our [Code of Conduct](https://www.storyblok.com/trust-center#code-of-conduct?utm_source=github.com&utm_medium=readme&utm_campaign=storyblok-js-client)

dipankarmaikap:
This PR enhances Astro live preview by introducing persistent server-side data support, a new `getPayload` API, and multiple ways to control live preview behavior at the route and editor level.

These changes improve performance, enable more flexible data handling, and give both developers and editors finer control over how and when live preview updates occur.

### Persistent server-side data

Astro live preview can now preserve server-side data across preview updates. This prevents unnecessary refetching and enables more advanced data flows.

**Key additions:**

* Introduces `StoryblokServerData.astro`, a new component for embedding sanitized server-side data.
* Embedded data survives live preview updates.
* All server data is automatically sanitized to prevent malicious injection.

**Example:**

```astro
---
import StoryblokServerData from '@storyblok/astro/StoryblokServerData.astro';

const users = await getUsers();
---

<StoryblokServerData users={users} />
```

This component embeds sanitized server-side data into the page so it can be reused across live preview updates without refetching.

---

### New `getPayload` API

A new `getPayload` API is introduced to support persistent server-side data without changing the existing `getLiveStory` API.

* `getLiveStory` remains unchanged and fully backward compatible.
* `getPayload` returns both live preview story data and persistent server-side data in a single call.
* The new API is fully opt-in.

**Example:**

```ts
interface ServerData {
  users?: User[];
}

interface MyStory extends ISbStoryData {
  content: { myField: string };
}

const { slug } = Astro.params;

const payload = await getPayload<ServerData, MyStory>({
  locals: Astro.locals,
});
```

**Returned payload:**

* `story`: live preview story data
* `serverData`: persistent server-side data

---

### Preserve component state during live preview

This PR also adds a way to preserve component or DOM state across live preview updates.

By adding the `data-preserve-state` attribute to a component or HTML element, that element will not be updated during live preview re-renders.

This is useful for cases such as:

* React components that fetch data from external APIs
* Open modals or dialogs
* Editor-only widgets inside the visual editor that should retain their state

**Example:**

```html
<div data-preserve-state>
  <!-- stateful component -->
</div>
```

This feature should be used with caution. In most cases it is not required, but it can be helpful for advanced editor tooling or custom widgets.

---

### Live preview control per route

Developers can disable live preview updates on a per-route basis using a meta tag. This is useful for routes where live preview updates are not desired and a full page reload is preferred.

**Example:**

```astro
{slug === 'test' && (
  <meta name="storyblok-live-preview" content="disabled" />
)}
```

When live preview is disabled for a route:

* Live preview updates on change are prevented
* The page falls back to standard SSR behavior with a full reload

---

### Editor-controlled live preview via events

Editor-facing control is enabled through an existing browser event. Teams can build custom UI, such as a checkbox or toggle, allowing editors to opt in or out of live preview updates.

**Example:**

```ts
document.addEventListener('storyblok-live-preview-updating', (event: Event) => {
  if (!checkbox?.checked) {
    event.preventDefault();
    console.log('Live preview update prevented by user preference');
  } else {
    console.log(
      'Live preview is updating...',
      (event as CustomEvent).detail
    );
  }
});
```

---

### Spinner and loading state considerations

If live preview updates are conditionally prevented, loading indicators must account for cancelled events. Otherwise, spinners may remain visible indefinitely.

**Recommended pattern:**

```ts
document.addEventListener('storyblok-live-preview-updating', (event) => {
  if (spinner) {
    spinner.style.display = 'block';

    queueMicrotask(() => {
      if (event.defaultPrevented && spinner) {
        spinner.style.display = 'none';
      }
    });
  }
});
```

This adjustment is only required when consumers actively prevent live preview updates.

---

### `isEditorRequest` helper

A new `isEditorRequest` utility is exposed to detect whether a request originates from the Storyblok editor.

This is useful for rendering editor-only UI or widgets that should not appear on the public site.

---

## Security

* Introduces an internal `sanitizeJSON` utility with test coverage
* Used internally by `StoryblokServerData` to sanitize embedded server data

---

## Fixes

* Fixes #95
* Fixes #96
* Fixes #97
* Fixes WDX-251

---

## Breaking changes

None.

Existing integrations using `getLiveStory` continue to work without modification. All new functionality is opt-in.

---

> [!NOTE]
> Introduces persistent server-side data in Astro live preview and adds APIs/utilities for finer control and safety.
> 
> - Adds `StoryblokServerData.astro` to embed sanitized server data that persists across live preview updates; exports and build pipeline updated
> - New `getPayload()` API (alongside deprecated `getLiveStory`) returns `{ story, serverData }`; middleware now stores both when editor POSTs
> - Live preview engine revamped: debounced updates, request aborting, `morphdom`-based DOM diffing with `data-preserve-state`, partial updates by focused blok, route-level disable via `<meta name="storyblok-live-preview" content="disabled" />`, and two cancelable events (`storyblok-live-preview-updating/updated`)
> - New `sanitizeJSON` utility with tests; used by `StoryblokServerData` to prevent XSS
> - New `isEditorRequest()` helper (with tests) and exposed from package; middleware uses it to validate editor requests
> - Playground examples (SSR widgets, spinner, FAQ components) demonstrating controls; added `morphdom` dependency
> - Releasing docs: support for `alpha/beta/next` branches and guidance for pre-releases
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6449bfd6cd94a945547fa2a2d34ce791dd769973. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/astro/src/live-preview/handleStoryblokMessage.ts
+++ b/packages/astro/src/live-preview/handleStoryblokMessage.ts
@@ -152,20 +152,20 @@ function updateDOMWithNewBody(
     if (newDomFocusElem) {
       // Add the [data-blok-focused] attribute to the above element
       newDomFocusElem.setAttribute(STORYBLOK_FOCUSED_ATTRIBUTE, 'true');
-      // Use morphdom to update the focused element while preserving state
-      updateWithMorphdom(focusedElem, newDomFocusElem);
+      // Update the focused element while preserving state
+      updateDOM(focusedElem, newDomFocusElem);
       shouldUpdateFullBody = false;
     }
     // If focused element not found in new DOM, fall back to full body update
   }
   if (shouldUpdateFullBody) {
-    // Use morphdom to efficiently morph the DOM while preserving state
-    updateWithMorphdom(currentBody, newBody);
+    // Efficiently update the DOM while preserving state
+    updateDOM(currentBody, newBody);
   }
 }
 /**
  * Returns a unique key for an element based on its data-blok-uid attribute
- * Used by morphdom to match elements between old and new DOM trees
+ * Used to match elements between old and new DOM trees
  */
 function getNodeKey(node: Node) {
   if (node.nodeType === 1) { // Element node
@@ -176,7 +176,7 @@ function getNodeKey(node: Node) {
   }
   return undefined;
 }
-function updateWithMorphdom(currentNode: Node, newNode: Node) {
+function updateDOM(currentNode: Node, newNode: Node) {
   morphdom(currentNode, newNode, {
     getNodeKey,
     onBeforeElUpdated: (fromEl, toEl) => {
--- a/packages/astro/playground/shared/components/Widget.astro
+++ b/packages/astro/playground/shared/components/Widget.astro
@@ -6,7 +6,7 @@ interface Props {
 const { disableLivePreview } = Astro.props;
 ---
 
-<div class="widget">
+<div class="widget" data-preserve-state>
   <div class="spinner-container">
     <p>Live preview spinner</p>
     <LivePreviewSpinner />


packages/astro/src/components/StoryblokServerData.astro
```diff
@@ -0,0 +1,24 @@
...
+ * ---
+ *
+ * <StoryblokServerData users={users} config={config} />
+ * ```
+ */
+
+const props = Astro.props;
+---
+
+<script is:inline id="__STORYBLOK_SERVERDATA__" type="application/json" set:html={JSON.stringify(props)} />
```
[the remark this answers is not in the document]
alexjoverm: @dipankarmaikap this one would be important to fix

packages/astro/src/live-preview/handleStoryblokMessage.ts
```diff
@@ -64,13 +74,16 @@ function updateDOMWithNewBody(
   }
 }
 
-async function getNewHTMLBody(story: ISbStoryData) {
+async function getNewHTMLBody(story: ISbStoryData, extraData?: unknown) {
```
alexjoverm: Same thing: `extraData` - `serverData`

packages/astro/playground/shared/components/LivePreviewSpinner.astro
```diff
@@ -0,0 +1,71 @@
+---
+
+---
+
+<div class="spinner" style="display: none;" id="spinner">
```
maoberlehner: issue: the id `spinner` can easily clash with an id a user is using in their code. I think we should prefix it and make it more specific. For example: `storyblok-live-preview-spinner`.

Although less problematic because of Astros scoped styles, to be on the safe side, we might consider to prefix the class too (the user might load global styles somewhere targeting a generic `spinner` class).

packages/astro/playground/shared/components/LivePreviewSpinner.astro
```diff
@@ -0,0 +1,71 @@
+---
+
+---
```
maoberlehner: nitpick: can this be removed?

packages/astro/src/live-preview/handleStoryblokMessage.ts
```diff
@@ -88,3 +126,13 @@ async function getNewHTMLBody(story: ISbStoryData) {
...
+
+/**
+ * Dispatches a cancelable custom event with optional detail payload.
+ * Returns the event object so the caller can check if preventDefault() was called.
+ */
+function dispatchCancelableStoryblokEvent<T>(name: string, detail?: T) {
+  const event = new CustomEvent<T>(name, { detail, cancelable: true });
+  document.dispatchEvent(event);
+  return event;
+}
```
maoberlehner: question: What was your reasoning for introducing a new function instead of adding a third `options` param to `dispatchStoryblokEvent`?

packages/astro/src/live-preview/handleStoryblokMessage.ts
```diff
@@ -1,6 +1,17 @@
...
 let timeout: NodeJS.Timeout;
+
+/**
+ * Checks if live preview is disabled via meta tag
+ */
+function isLivePreviewDisabled(): boolean {
+  const metaTag = document.querySelector<HTMLMetaElement>(
+    'meta[name="storyblok-live-preview"]',
+  );
+  return metaTag?.content === 'disabled' || metaTag?.content === 'false';
```
alexjoverm: question(non-blocking): @dipankarmaikap could this code ever run on the server?

My Astro knowledge is a bit limited, so just sanity-checking whether a client-only guard (window / document) is needed here.

packages/astro/src/lib/helpers.ts
```diff
@@ -46,14 +45,17 @@ export function useStoryblokApi(): StoryblokClient {
  * }
  * ```
  */
-export async function getLiveStory(
-  Astro: Readonly<AstroGlobal>,
-): Promise<ISbStoryData | null> {
-  let story: ISbStoryData | null = null;
-  if (Astro && Astro.locals._storyblok_preview_data) {
-    story = Astro.locals._storyblok_preview_data;
+interface Payload<T = unknown> {
```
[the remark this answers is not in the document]
dipankarmaikap: I think this would be great since it avoids any breaking changes for existing users. Should we also consider adding a deprecated warning so users can take action on their own?

packages/astro/src/utils/isEditorRequest.ts
alexjoverm: @dipankarmaikap this functionality would be ideal to have it in the JS SDK. There are some checks there but are hardcoded. Ideally, it would be encapsulated in a utility that can be also exported so the Astro SDK or any other SDK can reuse it

https://github.com/storyblok/monoblok/blob/main/packages/js/src/index.ts#L93

I'll leave it up to you if you want do it as part of this PR, or separately later in the future  (which is perfectly fine)

PR #441, opened 2026-02-18
feat(live-preview): add initial live preview package

dipankarmaikap:
## @storyblok/live-preview (initial version)

This PR introduces a new **Live Preview** package for Storyblok with a small, framework-agnostic API for handling preview updates on the client.

#### 1. High-level editor event helper (recommended)

`onStoryblokEditorEvent` covers the most common live preview use cases with minimal setup.

```ts id="t6m1x4"
import { onStoryblokEditorEvent } from '@storyblok/live-preview';

onStoryblokEditorEvent((story) => {
  // update your page state with the latest story data
});
```

Behavior:

* Calls the callback on `input` events with the updated story
* Automatically reloads the page on `change` and `published` events
* Safely runs only inside the Storyblok editor

#### 2. Low-level Preview Bridge access

For advanced use cases where full control is needed, the Preview Bridge can be accessed directly.

```ts id="ak9p2d"
import { loadStoryblokBridge } from '@storyblok/live-preview';

const bridge = await loadStoryblokBridge(bridgeOptions);

bridge.on(['input', 'change', 'published'], (event) => {
  // custom preview handling
});
```

Notes:

* The Preview Bridge is enforced as a **singleton per page**
* Configuration is **init-only** and cannot be changed at runtime
* Prevents duplicated events and inconsistent preview behaviour

#### 3. Editor detection utility

An `isInEditor` helper is also exposed for cases where conditional editor-only logic is required. This requires a URL.

### DIFF
--- a/packages/live-preview/package.json
+++ b/packages/live-preview/package.json
@@ -20,10 +20,15 @@
     "storyblok"
   ],
   "exports": {
-    ".": "./dist/index.mjs",
+    ".": {
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.cjs"
+    },
     "./package.json": "./package.json"
   },
-  "types": "./dist/index.d.mts",
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.mjs",
+  "types": "./dist/index.d.cts",
   "files": [
     "dist"
   ],
--- a/packages/live-preview/src/onStoryblokEditorEvent.ts
+++ b/packages/live-preview/src/onStoryblokEditorEvent.ts
@@ -14,7 +14,7 @@ const inputListeners = new Set<(story: ISbStoryData) => void>();
  * Tracks whether the Storyblok Preview Bridge event listeners
  * have already been registered.
  */
-let bridgeInitialized = false;
+let bridgeInitPromise: Promise<void> | undefined;
 
 /**
  * Initializes the Storyblok Preview Bridge and attaches event listeners.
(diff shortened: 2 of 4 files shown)


packages/live-preview/src/loadStoryblokBridge.ts
```diff
@@ -0,0 +1,25 @@
+import type { BridgeParams,} from '@storyblok/preview-bridge';
+import type StoryblokBridge from '@storyblok/preview-bridge';
```
maoberlehner: nitpick: this could be one statement

packages/live-preview/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@storyblok/live-preview",
+  "type": "module",
+  "version": "0.0.0",
+  "private": false,
+  "packageManager": "pnpm@10.30.0",
```
maoberlehner: nitpick: I've noticed that we have different values for `packageManager` throughout monoblok. My gut feeling tells me that we can/should remove this from individual packages and only add it to the root level `package.json`, or use the same version for all of them.

packages/live-preview/src/onStoryblokEditorEvent.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (event.action === 'input') {
+      callback(event.story as ISbStoryData<T>);
+      return;
+    }
+
+    if (event.action === 'change' || event.action === 'published') {
+      window.location.reload();
+    }
+  });
+};
```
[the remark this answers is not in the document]
maoberlehner: Good point. We could fix the lacking `off` mechanism by substituting our own listener mechanism. This is roughly how I did it in a personal project:

```ts
const inputListeners = new Set();
const changeListeners = new Set();
const publishedListeners = new Set();

let isInitialized = false;
function initializeBridge(bridge: StoryblokBridgeApi): void {
  if (isInitialized) return;

  bridge.on("input", (event) => {
    if (!event.story) return;
    for (const listener of inputListeners) {
      listener(event.story);
    }
  });
  bridge.on("change", () => {
    for (const listener of changeListeners) {
      listener();
    }
  });
  bridge.on("published", () => {
    for (const listener of publishedListeners) {
      listener();
    }
  });

  isInitialized = true;
}

export function onInput(callback: LivePreviewInputCallback): () => void {
  inputListeners.add(callback);
  // Every listener returns a cleanup function:
  return () => {
    inputListeners.delete(callback);
  };
}

export function onChange(callback: LivePreviewCallback): () => void {
  changeListeners.add(callback);
  return () => {
    changeListeners.delete(callback);
  };
}

export function onPublished(callback: LivePreviewCallback): () => void {
  publishedListeners.add(callback);
  return () => {
    publishedListeners.delete(callback);
  };
}
```

packages/live-preview/.gitignore
```diff
@@ -0,0 +1,4 @@
+node_modules
+dist
+*.log
+.DS_Store
```
maoberlehner: nitpick: all of this is already ignored in the .gitignore on the root level, could be removed

PR #450, opened 2026-02-26
feat(migrations): add content migration utilities package

maoberlehner:
Introduce @storyblok/migrations with helpers for paginated fetching, local file workflows, reference remapping, and schema cleanup to support programmatic content migrations between Storyblok spaces.

### DIFF
--- a/packages/cli/src/commands/stories/ref-mapper.ts
+++ b/packages/cli/src/commands/stories/ref-mapper.ts
@@ -258,7 +258,6 @@ export const storyRefMapper = (story: Story, { schemas, maps }: {
       : story.content,
     id: Number(maps.stories?.get(story.id) ?? story.id),
     uuid: String(maps.stories?.get(story.uuid) ?? story.uuid),
-    // @ts-expect-error Our types are wrong.
     parent_id: parentId != null ? Number(parentId) : null,
     alternates,
   } satisfies Story;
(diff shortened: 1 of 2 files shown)


packages/migrations/src/get-all-paginated.ts
```diff
@@ -0,0 +1,26 @@
+export async function getAllPaginated<T>(
```
alexjoverm: (blocking) @maoberlehner have you considered this (+ the getAllStories and getAllAssets) to have them in the MAPI client?

This logic is purely API fetching, nothing related to migration utils. And the `getAllPaginated` in particular could be useful for literally any endpoint accepting pagination.

packages/richtext/tsdown.config.ts
```diff
@@ -5,6 +5,7 @@ const sharedConfig = {
   clean: true,
   dts: true,
   external: ['mdast'],
+  hash: false,
```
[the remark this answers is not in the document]
maoberlehner: Without that, TS can't resolve the types leading to it complaining when I re-export them from this package.

packages/migrations/src/get-all-paginated.ts
maoberlehner: Hmm, I'm hesitant to put it into the client. I think it's somewhat of an anti-pattern for production code, so I think adding it to the client would give people the wrong idea.

It'd probably be safe to use sub 1.000 stories, but with spaces with multiple thousands of stories or assets, I'd not recommend using this. It'd not be such a big problem in one-off situations like migrations, but in production systems, you could start using `getAllPaginated` at the beginning with sub-1,000 stories, and everything is fine, but with every story added by a content editor, the build process gets slower and slower until it breaks.

What we could do is promote good practices and expose streaming helpers `getStoryStream`, helping developers do the performant thing.

PR #468, opened 2026-03-02
fix: change storyblokEditable to have narrow type

intellix:
The current type for storyblokEditable() says that it needs SbBlokData which has a very broad index type:
```ts
export interface SbBlokData extends ISbComponentType<string> {
    [index: string]: SbBlokKeyDataTypes;
}
```

This means that in order to satisfy the storyblokEditable function, I need my components to extend a type that breaks type safety.

This function ONLY needs 1x property, so just give it that

### DIFF
--- a/packages/js/src/editable.ts
+++ b/packages/js/src/editable.ts
@@ -1,6 +1,4 @@
-import type { SbBlokData } from './types';
-
-export default (blok: SbBlokData) => {
+export default (blok: { _editable?: string }) => {
   if (typeof blok !== 'object' || typeof blok._editable === 'undefined') {
     return {};
   }


dipankarmaikap (approved):
Thanks for opening this! Took it a step further — changed the type to `<T extends Record<string, unknown>>` so it accepts any object without requiring a shared property, and tightened the internal guard to `!== 'string'` for better type narrowing.

dipankarmaikap:
Hey @intellix, thanks a lot for this contribution! We've taken your fix and improved it a bit further in #685 — switched to `<T extends Record<string, unknown>>` to fully solve the weak-type detection issue raised by @AdrianFahrbach, and tightened the internal guard to `!== 'string'` for proper narrowing. Closing this in favour of that PR.

PR #490, opened 2026-03-10
fix: apply topological sort of stories and claim ids to avoid multiple

alexjoverm:
Fix cross-space `stories push` failures caused by creation ordering, slug collisions between folders and startpages, and content type restrictions on placeholder stories.

### Level-by-level creation with topological sort

The previous streaming pipeline fired concurrent creation requests without regard for depth, causing failures when a child story was created before its parent folder existed. Pass 1 now works in three steps:

1. **Scan** all local `.json` files into a lightweight index (metadata only)
2. **Group by depth** using `full_slug` segment count, with folders sorted before stories within each level
3. **Create level-by-level**, awaiting each depth before proceeding to the next. Within a level, requests are still concurrent (up to 12) for performance.

Because folders are always at a shallower depth than their children, `parent_id` is naturally resolved without an explicit dependency graph.

### Folder/startpage slug collision fix

In Storyblok, a folder and its startpage share the same `full_slug` (e.g. both `de-de/`). The previous `bySlug` map stored a single ref per slug, so the second entry overwrote the first. When processing level 0, the folder claimed the only match, and the startpage could not be matched, resulting in "Slug already taken" errors.

`bySlug` now stores an array of candidates per slug, and matching prefers entries whose `is_folder` matches the local entry:

```ts
const unclaimed = slugCandidates.filter(ref => !claimedRemoteIds.has(ref.id));
const match = unclaimed.find(ref => ref.is_folder === entry.is_folder) ?? unclaimed[0];
```

This ensures folders map to folders and startpages map to startpages, which is critical for correct `parent_id` resolution in child stories.

### Use real component name instead of `__migration_artifact__`

Placeholder stories were created with `component: '__migration_artifact__'`, which the API rejects when the target folder has content type restrictions. Since the real component name is already available from the scan index, the placeholder now uses `entry.component` directly. This requires components to have been pushed before stories (the documented workflow), and produces a clearer error if they haven't.

### `claimedRemoteIds` tracking

A `Set<number>` now tracks which remote IDs have been matched during creation, preventing two local stories from mapping to the same remote story after slug swaps or renames between pushes.

### DIFF
--- a/packages/cli/src/commands/stories/streams.ts
+++ b/packages/cli/src/commands/stories/streams.ts
@@ -355,89 +355,99 @@ export const createStoriesForLevel = async ({
   onStorySkipped?: (entry: StoryIndexEntry, remoteStory: TargetStoryRef, reason: string) => void;
   onStoryError?: (error: Error, entry: StoryIndexEntry) => void;
 }): Promise<void> => {
-  const tasks: Promise<void>[] = [];
-
-  for (const entry of level) {
+  const processEntry = async (entry: StoryIndexEntry) => {
     await apiConcurrencyLock.acquire();
+    try {
+      // Primary: check manifest mapping (from a previous push).
+      const mappedStoryId = maps.stories?.get(entry.id);
+      const mappedRemoteStory = mappedStoryId
+        ? existingTargetStories.byId.get(Number(mappedStoryId))
+        : undefined;
+      if (mappedRemoteStory) {
+        claimedRemoteIds.add(mappedRemoteStory.id);
+        onStorySkipped?.(entry, mappedRemoteStory, 'matched by manifest mapping from a previous push');
+        return;
+      }
 
-    const task = (async () => {
-      try {
-        // Primary: check manifest mapping (from a previous push).
-        const mappedStoryId = maps.stories?.get(entry.id);
-        const mappedRemoteStory = mappedStoryId
-          ? existingTargetStories.byId.get(Number(mappedStoryId))
-          : undefined;
-        if (mappedRemoteStory) {
-          claimedRemoteIds.add(mappedRemoteStory.id);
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/stories/streams.ts
```diff
@@ -205,134 +206,221 @@ export const mapReferencesStream = ({
...
               return;
             }
           }
+        }
 
-          const newRemoteStory = await transports.createStory(localStory);
-          await transports.appendStoryManifest(localStory, newRemoteStory);
-          onStorySuccess?.(localStory, newRemoteStory);
+        if (!entry.is_folder && !entry.component) {
+          throw new Error(`Story "${entry.slug}" is missing a content type (content.component). Every story must define a content field with a valid component.`);
```
dipankarmaikap: non-blocker: We can also add the filename to the user, too.

packages/cli/src/commands/stories/streams.ts
```diff
@@ -205,134 +206,221 @@ export const mapReferencesStream = ({
...
+          const unclaimed = slugCandidates.filter(ref => !claimedRemoteIds.has(ref.id));
+          const match = unclaimed.find(ref => ref.is_folder === entry.is_folder) ?? unclaimed[0];
+          if (match) {
+            const isMatchConfirmed = isCrossSpace || match.uuid === entry.uuid;
             if (isMatchConfirmed) {
-              await transports.appendStoryManifest(localStory, existingBySlug);
-              onStorySkipped?.(localStory, existingBySlug);
+              claimedRemoteIds.add(match.id);
+              await appendToManifest(entry, match);
+              onStorySkipped?.(entry, match);
```
dipankarmaikap: nit-pick(non-blocker): we can add a log to the user, saying why it's skipped.

PR #491, opened 2026-03-10
chore: ci release guard

alexjoverm:
## Summary

- Adds a `release-guard` GitHub Actions workflow that warns when a `feat`/`fix` PR touches multiple packages
- Since we squash-merge, all file changes land in a single commit — Nx release uses file changes (not commit scope) to determine which packages to bump, so cross-package `feat`/`fix` PRs cause unintended version bumps
- The workflow posts a comment explaining the issue and suggesting to split cosmetic changes into a `chore:` PR
- Automatically cleans up stale warnings when the PR is updated to only touch one package

## How it works

1. Triggers on PR open, sync, and title edit
2. Skips non-`feat`/`fix` PRs (those don't trigger version bumps)
3. Lists changed files and groups them by `packages/*/`
4. If 2+ packages are touched, posts (or updates) a warning comment
5. If the PR is later reduced to 1 package, deletes the stale comment

## Test plan

- [ ] Open a `feat(foo):` PR touching files in 2+ packages — verify warning comment appears
- [ ] Push an update removing the extra package changes — verify comment is deleted
- [ ] Open a `chore:` PR touching multiple packages — verify no comment
- [ ] Edit a PR title from `chore:` to `feat:` — verify workflow re-runs

### DIFF
--- a/.github/workflows/release-guard.yaml
+++ b/.github/workflows/release-guard.yaml
@@ -32,7 +32,7 @@ jobs:
         env:
           PR_TITLE: ${{ github.event.pull_request.title }}
         run: |
-          if bash tools/release-guard.sh --ci < /tmp/changed-files.txt > /tmp/touched-packages.txt; then
+          if bash tools/release-guard.sh < /tmp/changed-files.txt > /tmp/touched-packages.txt; then
             echo "cross_package=false" >> "$GITHUB_OUTPUT"
           else
             echo "cross_package=true" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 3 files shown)


dipankarmaikap:
@alexjoverm I think `tools/release-guard.sh` mainly works in a PR context (when a PR title exists) and is mostly intended for CI. When used locally as a pre-commit or commit-msg hook, it will usually do nothing unless a PR already exists and the GitHub CLI is configured.

Maybe the local hook could also do a simpler check on the commit itself. For example, if a commit like `feat: ...` touches multiple packages, it could show a warning asking the user to confirm this is intentional or suggest splitting it into separate commits.

Issue #105, opened 2024-10-22
Published responses cached regardless of caching settings


nathobson:
### Describe the issue you're facing

Responses from the Storyblok API are cached when retrieving published content, even when `apiOptions` contains a cache type of `none`.

### Reproduction

https://stackblitz.com/edit/withastro-astro-971qsd?file=.env

### Steps to reproduce

Using Astro in SSR mode via `output: 'server'` and the `Node` adapter in `mode: standalone`, responses from the Storyblok API are cached when retrieving published content, even when `apiOptions` contains a cache type of `none`. 

For example, the following call via `useStoryblok()` will return the cached response, not reflecting any updated (published) content:

```js
const story = await useStoryblok(`cdn/stories/home`, {
    version: "published",
  }
)
```

I've also tested with the settings as in `playground-ssr` in the [repo](https://github.com/storyblok/storyblok-astro/tree/main/playground-ssr), which has the following settings:

```js
apiOptions: {
  cache: {
    clear: "auto",
    type: "memory",
  },
}
```

This produced the same result.

However, if the exact same `useStoryblok()` call is set to retrieve drafts, everything updates when the page is refreshed, including draft and published content:

```js
const story = await useStoryblok(`cdn/stories/home`, {
    version: "draft",
  }
)
```

There appears to be a [closed and merged PR](https://github.com/storyblok/storyblok-js-client/pull/847) from three weeks ago in the [storyblok-js-client](https://github.com/storyblok/storyblok-js-client) repo but for whatever reason, this doesn't seem to have fixed the issue in the Storyblok Astro SDK.

Tested with the latest release (`v5.0.2`) of `@storyblok/astro`, which appears to contain the latest release of `storyblok/storyblok-js-client` from looking at the version installed in `node_modules`.

### System Info

```shell
System:
    OS: macOS 15.0.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 66.47 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.12.2 - ~/.nvm/versions/node/v20.12.2/bin/node
    npm: 10.5.0 - ~/.nvm/versions/node/v20.12.2/bin/npm
    bun: 1.0.22 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 130.1.71.114
    Chrome: 129.0.6668.103
    Safari: 18.0.1
    Safari Technology Preview: 18.0
```

### Used Package Manager

npm

### Validations

[...]

PR #508, opened 2026-03-14
fix(js-client): fallback to space.version for cache version

intellix:
The way that new versions are discovered today, is that it checks for data.cv and saves it in-memory. The problem is, the only way that data.cv is available, is through heavily cached endpoints. The cdn/spaces/me endpoint is only cached for 2 seconds and has a small response and includes the latest known version, but it's not used at all, until this PR.

With this change it'll allow me to poll the cdn/spaces/me endpoint every 60 seconds and switch back to querying for published content using the SDK. 

Addressed the findings from this comment: https://github.com/storyblok/monoblok/issues/105#issuecomment-4060334565

### DIFF
--- a/packages/js-client/src/index.ts
+++ b/packages/js-client/src/index.ts
@@ -737,14 +737,17 @@ export class Storyblok {
         const isCacheClearable = (this.cache.clear === 'onpreview' && params.version === 'draft')
           || this.cache.clear === 'auto';
 
-        if (params.token && response.data.cv) {
+        // Get cv from story or space
+        const cv = response.data.cv ?? response.data.space?.version;
+
+        if (params.token && cv) {
           if (isCacheClearable
             && cacheVersions[params.token] // there is a cache
-            && cacheVersions[params.token] !== response.data.cv // a new cv is incoming
+            && cacheVersions[params.token] !== cv // a new cv is incoming
           ) {
             await this.flushCache();
           }
-          cacheVersions[params.token] = response.data.cv;
+          cacheVersions[params.token] = cv;
         }
 
         return resolve(response);


intellix:
pnpm patch for anyone that needs this now (me, but sharing anyway): 

for **4.4.4:**

**patches/@storyblok__js.patch**
```patch
diff --git a/dist/storyblok-js.mjs b/dist/storyblok-js.mjs
index 15130c1b6351bcc429d08afc3052b21af4201dbd..ca9ecc87a3a482efdf66dda0ecd983f42a22ac37 100644
--- a/dist/storyblok-js.mjs
+++ b/dist/storyblok-js.mjs
@@ -822,7 +822,8 @@ var $e = class {
         }
         t.version === "published" && e !== "/cdn/spaces/me" && await n.set(i, E);
         const T = this.cache.clear === "onpreview" && t.version === "draft" || this.cache.clear === "auto";
-        return t.token && E.data.cv && (T && S[t.token] && S[t.token] !== E.data.cv && await this.flushCache(), S[t.token] = E.data.cv), c(E);
+        const cv = E.data.cv || E.data.space?.version;
+        return t.token && cv && (T && S[t.token] && S[t.token] !== cv && await this.flushCache(), S[t.token] = cv), c(E);
       } catch (R) {
         if (R.response && R.status === 429 && (r = typeof r > "u" ? 0 : r + 1, r < this.maxRetries))
           return console.log(`Hit rate limit. Retrying in ${this.retriesDelay / 1e3} seconds.`), await Re(this.retriesDelay), this.cacheResponse(e, t, r).then(c).catch(b);

```

intellix:
We needed to update our patch:

**4.4.5**

```patch
diff --git a/dist/storyblok-js.mjs b/dist/storyblok-js.mjs
index 1402bc6791e638e2e94573ca6e98d8b2cec2487d..a69fb4062e3babd362cdce651a0b06624e9779db 100644
--- a/dist/storyblok-js.mjs
+++ b/dist/storyblok-js.mjs
@@ -817,7 +817,8 @@ var Me = class {
         }
         t.version === "published" && e !== "/cdn/spaces/me" && await n.set(i, A);
         const T = this.cache.clear === "onpreview" && t.version === "draft" || this.cache.clear === "auto";
-        return t.token && A.data.cv && (T && S[t.token] && S[t.token] !== A.data.cv && await this.flushCache(), S[t.token] = A.data.cv), c(A);
+        const cv = A.data.cv || A.data.space?.version;
+        return t.token && cv && (T && S[t.token] && S[t.token] !== cv && await this.flushCache(), S[t.token] = cv), c(A);
       } catch (R) {
         if (R.response && R.status === 429 && (r = typeof r > "u" ? 0 : r + 1, r < this.maxRetries))
           return console.log(`Hit rate limit. Retrying in ${this.retriesDelay / 1e3} seconds.`), await pe(this.retriesDelay), this.cacheResponse(e, t, r).then(c).catch(b);
```

maoberlehner:
Thanks for digging into this @intellix — the diagnosis is right, and `main` genuinely can't do what our [caching docs](https://www.storyblok.com/docs/concepts/caching#retrieve-the-latest-cache-version) tell people to do.

I opened #761, which keeps your commit and your polling use case, with one correction. `space.version` and `cv` aren't the same number: when an access token has a Minimum Cache TTL configured, the content endpoints return a `cv` rounded down to the TTL window, while `cdn/spaces/me` keeps reporting the space's raw version. Measured live against a token with a TTL of 3600, `??` between them made the tracked value oscillate — `flushCache()` fired 5 times in 6 requests with no content change, and every content request took a 301.

#761 tracks the space version separately and uses it purely as a flush signal, leaving `cv` to content responses. Your scenario is covered by a test, and I verified end-to-end against a live space that stale published content now recovers after a poll.

Docs fix for the same confusion: storyblok/storyblok-docs-platform#627.

Closing in favour of #761 — happy to have your review there.

maoberlehner:
Superseded by: https://github.com/storyblok/monoblok/pull/761

Thank you, @intellix, for making me look closer at this and spot a flaw in our cv/version implementation!

PR #531, opened 2026-03-25
feat(mapi-client)!: rewrite management API client with resource-based architecture

maoberlehner:


alexjoverm (changes requested):

packages/capi-client/src/utils/rate-limit.ts
alexjoverm: (potentially blocking) @maoberlehner was thinking - any consideration here regarding this? It would be problematic to set the rate limit based on the concurrency limit (that one is 30 always AFAIK) as they're different values, but concurrency is higher.

Shall we filter by the value to ensure it reads it if **only** they're rate limits? (which btw, they're not implemented, and likely the won't soon) - could stay as a noop function for now as well.

https://github.com/storyblok/storyblok-docs-platform/pull/447#discussion_r3015331796

alexjoverm (approved):
LGTM - huge work here @maoberlehner! 

There are 2 pending things, but no need to re-review from my side:
- definition of the delete/remove methods
- comment about rate limit header

PR #546, opened 2026-04-02
fix(cli): add loadComponents util into loader file to avoid potential

alexjoverm:
Replaces the three divergent approaches to reading components/datasources from disk with a single **content-based classification** strategy. Items are identified by their data shape (e.g. `schema` → component, `slug` → datasource), not by filename — making it agnostic to custom filenames (`--filename my-schemas`).

### Before → After

| Before | After |
|--------|-------|
| `shouldUseSeparateFiles()` + `consolidatedFilename()` to detect format, then branch | `loadComponents(dir, { suffix? })` reads all matching JSON, classifies by content |
| `findComponentSchemas` had its own manual read loop | Reuses `loadComponents` |
| Silent dedup or format-detection errors | Duplicate detection: errors with which files conflict |
| 3 different approaches across push/stories/types | 1 shared loader |


packages/cli/src/commands/stories/utils.ts
```diff
@@ -22,54 +20,21 @@ export const isStoryWithUnpublishedChanges = (story: Partial<Story>) => {
...
-      }
-      continue;
-    }
-
-    // Maybe component file:
-    const component = toComponent(content);
-    if (component) {
-      components.push(component);
+  try {
+    const { components } = await loadComponents(directoryPath);
```
maoberlehner, line 26: Question: Do you think we need suffix support here? Previously, it was actually broken, because it read all files regardless of the suffix. So it was absolutely the right call to unify this behavior. But now people can not push stories if they use a suffix.

packages/cli/src/utils/filesystem.ts
```diff
@@ -248,15 +248,24 @@ export async function fileExists(path: string) {
...
-  }
-  return !(await fileExists(join(resolvedPath, consolidatedFilename(defaultFilename, suffix))));
+/**
+ * Filters a list of filenames to only include JSON files matching the suffix convention.
+ *
+ * When `suffix` is set (e.g. "dev"), only `*.dev.json` files are included.
+ * When unset, files with a `.<word>.json` pattern are excluded (they belong to other environments),
+ * except for known sub-resource patterns like `.presets.json`.
+ */
+export function filterJsonBySuffix(files: string[], suffix?: string, subResourcePatterns: string[] = []): string[] {
```
maoberlehner: This behavior clashes with the ability to choose a custom `--filename`. Users might choose a name like `my.components`, which would get filtered out. Impossible to know _why_ for users.

maoberlehner (changes requested):
IMPORTANT: I did a rebase (because it was my fault for having merge conflicts because of waiting so long for the review). So consider this when you check it out on your machine.

Do we need to update `--help` descriptions for commands to reflect the new behavior?

alexjoverm:
Thanks for the input @maoberlehner! Those were great findings. I was a bit surprised tests didn't capture those potential regressions, I took the chance to try to improve the coverage and the messaging

packages/cli/src/commands/components/loader.ts
```diff
@@ -0,0 +1,95 @@
...
+      }
+      else if (isComponentGroup(item)) {
+        groupMap.set(item.id, item);
+      }
+    }
+  }
+
+  if (duplicates.length) {
+    throw new FileSystemError(
+      'file_not_found',
```
maoberlehner: `file_not_found` does not seem quite right for the `duplicates` case.

packages/cli/src/commands/components/loader.ts
```diff
@@ -0,0 +1,95 @@
...
+ * Loads component entities from a directory by reading all matching JSON files
+ * and classifying items by their data shape.
+ *
+ * Throws on duplicate components (same name found in multiple files).
+ */
+export async function loadComponents(
+  directoryPath: string,
+  options?: { suffix?: string },
+): Promise<ComponentsData> {
+  const files = await readdir(directoryPath);
```
maoberlehner: Is there a specific reason why you didn't use our `readDirectory()` helper? If not, consider using it.

maoberlehner (approved):
LGTM except for two minor findings, but feel free to squash after taking a look at them.

PR #551, opened 2026-04-06
fix(astro): lazy-load Storyblok components to prevent circular dependency

### LINKED ISSUE #547: [Bug]: TDZ ReferenceError when Storyblok components have circular dependency through `StoryblokComponent`
### Package

@storyblok/astro

### Bug Description

When two registered Storyblok components form a circular dependency through `StoryblokComponent.astro`, Vite 6's SSR bundler emits them into a single chunk with deferred getters. Because the `virtual:import-storyblok-components` virtual module eagerly accesses `.default` on every glob'd module at module initialization time, the deferred getter fires before the target variable is initialized, producing a TDZ `ReferenceError`.

The cycle is: `Card.astro` imports `Richtext.astro`, which imports `StoryblokComponent.astro` (from `@storyblok/astro`), which imports `virtual:import-storyblok-components`, which eagerly re-imports `Card.astro` via the `import.meta.glob(..., { eager: true })` call and immediately reads `modules[filePath].default` at the top level.

### Steps to Reproduce

1. Create two Storyblok components where one imports the other, and the second imports `StoryblokComponent` from `@storyblok/astro`:

   ```
   Card.astro → imports Richtext.astro
   Richtext.astro → imports @storyblok/astro/StoryblokComponent.astro
   StoryblokComponent.astro → imports virtual:import-storyblok-components
   virtual module → eagerly imports Card.astro (cycle)
   ```

2. Register both in `astro.config.ts`:

   ```ts
   storyblok({
     accessToken: "dummy",
     components: {
       card: "storyblok/Card",
       text: "storyblok/Richtext",
     },
   })
   ```

3. Create a page that renders `Card`.

4. Run `astro build`.

### Reproduction URL

https://stackblitz.com/github/adamchal/storyblok-tdz?startScript=build

### Expected Behavior

The build completes successfully. Circular component dependencies are a natural pattern in CMS-driven sites (a Card can contain rich text, and rich text can contain embedded Storyblok blocks via `StoryblokComponent`).

### Actual Behavior

The build fails during static generation with:

```
ReferenceError: Cannot access '$$Card' before initialization
```

### Environment

```shell
System:
    OS: macOS 15.7.4
    CPU: (16) x64 Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
  Binaries:
    Node: 25.8.2
    npm: 11.12.1
  npmPackages:
    @storyblok/astro: ^9.0.3 => 9.0.3
    astro: ^6.1.3 => 6.1.3
```

### Error Logs

```shell
14:46:08 [ERROR] ReferenceError: Cannot access '$$Card' before initialization
    at get default (file:///tmp/storyblok-tdz/dist/.prerender/chunks/index_Cj6juoCk.mjs:12:20)
    at file:///tmp/storyblok-tdz/dist/.prerender/chunks/index_Cj6juoCk.mjs:35435:45
    at ModuleJob.run (node:internal/modules/esm/module_job:430:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:639:26)
    at async BuildPipeline.getModuleForRoute
```

### Additional Context

The root cause is in the generated `virtual:import-storyblok-components` module (`vite-plugin-import-storyblok-components`). It does:

```js
const modules = import.meta.glob('.../storyblok/**/*.astro', { eager: true });
const storyblokComponents = {};
for (const filePath in modules) {
  const component = modules[filePath].default || modules[filePath]; // <-- TDZ here
  storyblokComponents[camelCaseName] = component;
}
```

A fix would be to defer the `.default` access until render time, e.g. using lazy property getters:

```js
const modules = import.meta.glob('.../storyblok/**/*.astro', { eager: true });
const storyblokComponents = {};
for (const filePath in modules) {
  const camelCaseName = toCamelCase(componentName);
  Object.defineProperty(storyblokComponents, camelCaseName, {
    get() { return modules[filePath].default || modules[filePath]; },
    enumerable: true,
    configurable: true,
  });
}
```

This defers `.default` access to render time, when all module-level variables are fully initialized.

dipankarmaikap:
This PR switches Storyblok component imports to lazy loading to prevent circular dependency issues and improve build stability.

Previously, components were imported using `eager` imports, which could cause runtime errors like:

`ReferenceError: Cannot access '$$Component' before initialization`

This change introduces async component loaders and updates the runtime to resolve components lazily.

Fixes #547

## Changes

- Convert auto-imported Storyblok components to lazy imports
- Convert manual component imports to lazy imports
- Convert fallback component to lazy import
- Update runtime to resolve components asynchronously

### DIFF
--- a/packages/astro/src/vite-plugins/vite-plugin-import-storyblok-components.ts
+++ b/packages/astro/src/vite-plugins/vite-plugin-import-storyblok-components.ts
@@ -90,40 +90,31 @@ function generateModuleCode(
   return `
     // Import utilities and fallback component
     import { toCamelCase } from '@storyblok/astro';
-    ${fallbackImport}
-    ${manualImports.join('\n')}
 
     // Dynamically import all Storyblok components using Vite's glob import
-    const modules = import.meta.glob('${globPattern}', { eager: true });
+    const modules = import.meta.glob('${globPattern}');
     // Process imported modules into a components object
     const storyblokComponents = {};
-    
     for (const filePath in modules) {
       // Extract component name from file path (remove extension)
       const fileName = filePath.split('/').pop();
       const componentName = fileName?.replace(/\\.[^/.]+$/, '');
       
       if (componentName) {
-        // Get the component (handle both default and named exports)
-        const component = modules[filePath].default || modules[filePath];
-        
         // Convert filename to camelCase for Storyblok component naming
         const camelCaseName = toCamelCase(componentName);
-        storyblokComponents[camelCaseName] = component;
+
+        storyblokComponents[camelCaseName] = async () => {
+        const module = await modules[filePath]();
+        return module.default || module;
+        };
       }
     }
     
-    // Manual imports (overwrite auto if same key exists)
-    ${manualImports
-      .map((imp) => {
-        const varName = imp.match(/import\s+(\w+)\s+from/)?.[1];
-        return `storyblokComponents['${varName}'] = ${varName};`;
-      })
-      .join('\n')}
-
+    // Manual components
+    ${manualImports.join('\n')}
     // Add fallback component if enabled
-    ${fallbackImport ? `storyblokComponents['FallbackComponent'] = FallbackComponent;` : ''}
-    
+    ${fallbackImport}    
     // Export the components object for use in Storyblok initialization
     export { storyblokComponents };
   `.trim();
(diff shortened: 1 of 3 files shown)


packages/astro/src/vite-plugins/vite-plugin-import-storyblok-components.ts
```diff
@@ -90,40 +90,31 @@ function generateModuleCode(
...
-        const component = modules[filePath].default || modules[filePath];
-        
         // Convert filename to camelCase for Storyblok component naming
         const camelCaseName = toCamelCase(componentName);
-        storyblokComponents[camelCaseName] = component;
+
+        storyblokComponents[camelCaseName] = async () => {
+        const module = await modules[filePath]();
+        return module.default || module;
+        };
```
maoberlehner, line 110: nitpick: indentation is off

maoberlehner (approved):
one nitpick, otherwise looks good!

PR #553, opened 2026-04-07
fix: correctly generate multilink union types

### LINKED ISSUE #412: 🐛 [Bug]: typescript linktype type does not get properly generated
### Package

storyblok (CLI)

### Bug Description

 Generating types for a link field that does not allow emails or assets does not properly narrow the type down.
```ts
// storyblok.d.ts
export interface StoryblokMultilink {
  linktype?: 'story' | 'url' | 'email' | 'asset'
}

// storyblok-components.d.ts
export interface SomeBlok {
  url: Exclude<StoryblokMultilink, {linktype?: "email"} | {linktype?: "asset"};
}
```
Exclude only has an effect on union types; https://www.typescriptlang.org/docs/handbook/utility-types.html#excludeuniontype-excludedmembers

### Steps to Reproduce

1.     create a Link field on a Blok
2.     leave Enable email field or Enable asset selection unchecked in Link Field Options
3.     storyblok components pull
4.     storyblok types generate

### Expected Behavior

`"email" | "asset"` should be excluded from a restricted link field.
```ts
declare const b: SomeBlok;
b.url.linktype;
//      ^?: "story" | "url"
```

### Actual Behavior

Generated type does not exclude any members.

### Code Sample

```javascript

```

### Environment

```shell
-
```

### Error Logs

```shell

```

### Additional Context

_No response_

dipankarmaikap:
Fixes #412

### DIFF
--- a/packages/cli/src/types/storyblok.ts
+++ b/packages/cli/src/types/storyblok.ts
@@ -88,15 +88,14 @@ interface StoryblokMultilinkBase {
 export type StoryblokMultilink =
   | (StoryblokMultilinkBase & {
     linktype: 'story';
-    story?: StoryblokMultilinkStory;
+    story?: StoryblokMultilinkStory | StoryblokMultilinkLink | StoryblokMultilinkUrl;
   })
   | (StoryblokMultilinkBase & {
     linktype: 'url';
-    story?: StoryblokMultilinkUrl;
   })
   | (StoryblokMultilinkBase & {
     linktype: 'email';
-    email?: string;
+    email: string;
   })
   | (StoryblokMultilinkBase & {
     linktype: 'asset';
--- a/packages/cli/src/types/storyblok.ts
+++ b/packages/cli/src/types/storyblok.ts
@@ -73,7 +73,7 @@ export interface StoryblokMultilinkUrl {
   full_slug: string;
 }
 
-export interface StoryblokMultilink {
+interface StoryblokMultilinkBase {
   fieldtype: 'multilink';
   id: string;
   url: string;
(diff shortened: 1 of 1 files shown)


packages/cli/src/types/storyblok.ts
```diff
@@ -83,11 +83,25 @@ export interface StoryblokMultilink {
...
 }
 
+export type StoryblokMultilink =
+  | (StoryblokMultilinkBase & {
+    linktype: 'story';
+    story?: StoryblokMultilinkStory;
+  })
+  | (StoryblokMultilinkBase & {
+    linktype: 'url';
+    story?: StoryblokMultilinkUrl;
```
maoberlehner: Is this correct? Shouldn't it just be `linktype: 'url';` no story?

packages/cli/src/types/storyblok.ts
```diff
@@ -83,11 +83,25 @@ export interface StoryblokMultilink {
...
+  })
+  | (StoryblokMultilinkBase & {
+    linktype: 'email';
+    email?: string;
```
maoberlehner: Shouldn't it be `email: string;` (non-optional) because even if it is empty, it is still an empty string, but the field does exist?

maoberlehner (changes requested):

packages/cli/src/types/storyblok.ts
dipankarmaikap: Thank you, You are right, Fixed it.

maoberlehner (approved):

PR #554, opened 2026-04-07
fix(angular): improve angular sdk

dipankarmaikap:


packages/angular/src/lib/types.ts
```diff
@@ -0,0 +1,9 @@
+import type { ISbComponentType } from 'storyblok-js-client';
```
maoberlehner, line 1: storyblok-js-client is now a devDependency, won't this cause TypeScript resolution failures for consumers who don't have it installed?

packages/angular/src/lib/components.feature.ts
```diff
@@ -103,3 +103,73 @@ export function isComponentLoader(
...
+  Record<K, Type<unknown> | StoryblokComponentLoader>
+>;
+
+/**
+ * Abstract base class for component resolvers.
+ * Provides common logic for resolving components from a registry with caching support.
+ *
+ * @typeParam K - The key type used in the registry (e.g., string, StoryblokSegmentType)
+ */
+@Directive()
```
maoberlehner, line 121: I have no idea about Angular but the AI told me to ask: `BaseComponentResolver` is a shared service base, not a template directive. Wouldn't @Injectable() be more accurate than @Directive() here? Or do we even need a decorator?

packages/angular/src/lib/sb-blok.directive.ts
```diff
@@ -1,86 +1,116 @@
...
+    }
+
+    // Recreate component only if type changed
+    if (this.currentComponentType !== componentType) {
+      this.clearComponent();
+      this.componentRef = this.viewContainerRef.createComponent(Component);
+    }
+
+    this.currentComponentType = componentType;
+    this.componentRef!.setInput('blok', blok);
```
maoberlehner, line 113: Doesn't this fail if we never enter the `if` above? Can we use optional chaining here instead? `this.componentRef?.setInput('blok', blok);`

packages/angular/src/lib/components.feature.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+  it('should return false for component classes', () => {
+    expect(isComponentLoader(TestComponent)).toBe(false);
+  });
+
+  it('should return false for classes with prototype', () => {
+    class SomeClass {
+      method() {}
+    }
+
+    expect(isComponentLoader(SomeClass as unknown as Type<unknown>)).toBe(false);
```
maoberlehner, line 40: I think the `as` type casts are unnecessary here.

packages/angular/src/lib/sb-blok.directive.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+      _uid: '1',
+      component: 'teaser',
+      headline: 'Cleanup',
+    });
+
+    await waitForRender();
+
+    fixture.destroy();
+
+    expect(true).toBe(true);
```
maoberlehner, line 186: If this assertion ever fails, we've got problems! 😅

maoberlehner (changes requested):

dipankarmaikap:
Closing this in favour of https://github.com/storyblok/monoblok/pull/576

PR #556, opened 2026-04-09
feat(richtext): new static method for richtext rendering

dipankarmaikap:
# Add Static Endpoint for Richtext Rendering

This PR introduces a **static endpoint** to the richtext package.

## What’s New

At build time, this endpoint generates:

* Attribute types
* Render map

This allows us to **directly render Storyblok Richtext JSON** without relying on heavy client-side logic.

The generated package is **less than 2kb (gzip)** and works across **all frameworks**.

## Additional Helpers

This PR also includes a few helper utilities to:

* Handle styles
* Normalize attributes
* Work with different node and mark types

## Type-safe Rendering

Using this approach, I implemented richtext rendering for:

* React
* Astro

Both implementations are:

* Fully type-safe
* Lightweight
* Designed for good developer experience

You can see a working example in the Astro playground:
[https://github.com/dipankarmaikap/astro-playground](https://github.com/dipankarmaikap/astro-playground)

## SDK Integration

The intended approach is:

* Create framework-specific components inside each SDK
* Users can import and use them directly

Examples included:

* Custom blok rendering
* Custom mark rendering
* Custom node rendering

## Benefits

* Tiny bundle size (<2kb gzip)
* Type-safe richtext rendering
* Framework-agnostic
* Minimal runtime logic
* Better developer experience


packages/richtext/src/static/generate/index.ts
```diff
@@ -0,0 +1,17 @@
+import fs from 'node:fs';
+import path from 'node:path';
```
maoberlehner: For Windows compatibility, we should use `pathe` as a drop-in replacement for `node:path`.

packages/richtext/src/static/generate/parse-dom-spec.ts
```diff
@@ -0,0 +1,99 @@
+import type { DOMOutputSpec } from 'prosemirror-model';
```
maoberlehner, line 1: nitpick: inconsistent file naming: other files use `kebab-case` this one camelCase. I think we should stick to `parse-dom-spec.ts`.

packages/richtext/src/static/generate/parseDOMSpec.ts
```diff
@@ -0,0 +1,99 @@
...
+} as const;
+
+export function parseDOMSpec(spec: DOMOutputSpec): RenderSpec | null {
+  // string
+  if (typeof spec === 'string') {
+    return { tag: spec };
+  }
+  // custom override
+  if (Array.isArray(spec) && typeof spec[0] === 'string') {
+    const custom = CUSTOM_SPECS[spec[0] as keyof typeof CUSTOM_SPECS];
```
maoberlehner: `as` casts defeat the point of using TypeScript, so we should avoid it at all costs (and if we can't, add a comment explaining why there is no other way).

Here we can tighten TS types with:

```ts
const CUSTOM_SPECS: Record<string, DOMOutputSpec> = {
  table: ['table', ['tbody', 0]],
} as const;
```

and then remove the casts:

```ts
if (Array.isArray(spec) && typeof spec[0] === 'string') {
  const custom = CUSTOM_SPECS[spec[0]];
  if (custom) {
    spec = custom;
  }
}
```

packages/richtext/src/static/generate/parseDOMSpec.ts
```diff
@@ -0,0 +1,99 @@
...
+    if (custom) {
+      spec = custom as DOMOutputSpec;
+    }
+  }
+  // { dom, contentDOM }
+  if (isDOMObject(spec)) {
+    return {
+      tag: spec.dom.nodeName.toLowerCase(),
+      content: !!spec.contentDOM,
```
maoberlehner: nitpick: personally I prefer `Boolean(spec.contentDOM)` but it's a matter of taste. I'm okay with this variant, too.

packages/richtext/src/static/generate/parseDOMSpec.ts
```diff
@@ -0,0 +1,99 @@
...
+      tag,
+      ...(attrs && Object.keys(attrs).length ? { attrs } : {}),
+      ...(content ? { content: true } : {}),
+      ...(parsedChildren.length ? { children: parsedChildren } : {}),
+    };
+
+    return result;
+  }
+  catch (err) {
+    console.error('parseArraySpec error:', err);
```
maoberlehner: Doesn't this lead to the error being logged twice? Once here and once because of the throw?

packages/richtext/src/static/generate/parse-dom-spec.ts
```diff
@@ -0,0 +1,99 @@
...
+  return (
+    typeof value === 'object'
+    && value !== null
+    && !Array.isArray(value)
+  );
+}
+
+function isDOMObject(
+  value: unknown,
+): value is { dom: Node; contentDOM?: HTMLElement } {
```
maoberlehner: Nice!

packages/richtext/src/static/generate/parse-dom-spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    return result;
+  }
+  catch (err) {
+    console.error('parseArraySpec error:', err);
+    console.error('spec causing error:', spec);
+    throw err;
+  }
+}
+
+function isAttrs(value: unknown): value is Record<string, any> {
```
maoberlehner: `any` also kinda defeats the point of TS. Can we avoid it here?

packages/richtext/src/static/generate/richtext-render-map.ts
```diff
@@ -0,0 +1,98 @@
+import { type AnyExtension, getSchema } from '@tiptap/core';
+import { getStoryblokExtensions } from '../../extensions';
+import type { Schema } from 'prosemirror-model';
+import { MarkType, NodeType } from 'prosemirror-model';
+import { parseDOMSpec } from './parseDOMSpec';
+/**
```
maoberlehner, line 7: mega nitpick: new line would be nice here.

packages/richtext/src/static/render-map.generated.ts
```diff
@@ -0,0 +1,175 @@
+// THIS FILE IS AUTO-GENERATED. DO NOT EDIT.
```
maoberlehner: Is there a reason why those generated files are included in Git?

packages/richtext/src/static/types.ts
```diff
@@ -0,0 +1,39 @@
+import type { PMMark, PMNode, TiptapComponentName } from './types.generated';
+
+export interface SbBlokData {
+  _key: string;
+  component: string;
+  [otherKey: string]: any;
```
maoberlehner: Can we avoid the `any`s in this file, too?

packages/richtext/src/static/types.ts
```diff
@@ -0,0 +1,39 @@
...
+      ? Extract<PMNode, { type: T }>
+      : T extends PMMark['type']
+        ? Extract<PMMark, { type: T }>
+        : never
+  ) & {
+    components?: StoryblokRichTextComponentMap;
+  };
+
+export type StoryblokRichTextComponentMap<
+  R = any,
```
maoberlehner: I'm a fan of naming generics. In this case, I have no idea what `R` stands for. I use the convention of prefixing generics with `T`. E.g. `TComponent`. This prevents clashes with non-generic variants.

packages/richtext/src/static/types.ts
maoberlehner: (applies to all `any` in the PR: if we can, we should not use `any` except when there is no other way, e.g., because we depend on poorly typed third-party code.)

packages/richtext/eslint.config.js
```diff
@@ -3,5 +3,5 @@ import { storyblokLintConfig } from '@storyblok/eslint-config';
 export default storyblokLintConfig({
 
 }, {
-  ignores: ['tests/unit/coverage/', 'dist/', 'playground/astro/.astro/', 'playground/vanilla/public/', 'README.md'],
+  ignores: ['tests/unit/coverage/', 'dist/', 'playground/astro/.astro/', 'playground/vanilla/public/', 'README.md', 'src/static/render-map.generated.ts', 'src/static/types.generated.ts'],
```
maoberlehner: nitpick: could use pattern like `*.generated.*`

maoberlehner:
## Code Review (via Claude Code)

> Automated review + manual inspection of all changed files.

---

### Critical

**1. Heading `resolveTag` — unsanitized template literal allows XSS**

The heading resolve function is:
```ts
(attrs: TiptapNodeAttributes['heading']) => `h${attrs?.level || 1}`
```
If `level` is a non-numeric value (e.g. from crafted richtext JSON), the returned string becomes an arbitrary HTML tag name:
```js
resolveTag({ type: 'heading', attrs: { level: '"><img src=x onerror=alert(1)>' } })
// → 'h"><img src=x onerror=alert(1)>'
```
In SSR contexts (Astro, Nuxt) that render raw HTML, this is exploitable. The `level` value must be validated/clamped to `1-6`.

**2. No tests for the new static module**

~530 lines of runtime code (`style.ts`, `util.ts`, `types.ts`) and ~330 lines of codegen, with zero test coverage. `processAttrs`, `resolveTag`, `parseDOMSpec`, `styleToString`/`stringToStyle`, and all generator logic are untested.

---

### Major

**3. `stringToStyle` breaks on CSS values containing colons**

**File:** `packages/richtext/src/static/style.ts:108`

`rule.split(':')` splits on *every* colon. A value like `background: url(https://example.com)` produces `{ background: 'url(https' }`, discarding the rest. Split on only the first colon:
```ts
const colonIdx = rule.indexOf(':');
const key = rule.slice(0, colonIdx).trim();
const value = rule.slice(colonIdx + 1).trim();
```

**4. `processAttrs` leaks falsy values into styles**

The null guard `if (value == null) continue` skips `null`/`undefined` but passes through `""`, `false`, and `0`:
```js
processAttrs('paragraph', { textAlign: '' })   // → { style: { textAlign: '' } }
processAttrs('paragraph', { textAlign: false }) // → { style: { textAlign: false } }
```
These produce broken inline styles like `style="text-align: "` or `style="text-align: false"`.

**5. `styleToString` doesn't filter null/undefined/false**

```js
styleToString({ textAlign: null })      // → "text-align: null"
styleToString({ textAlign: undefined }) // → "text-align: undefined"
```
Combined with #4, the pipeline `processAttrs → styleToString` can produce `style="text-align: null"` in rendered HTML.

**6. Generated render map bakes in `null` as literal values**

**File:** `packages/richtext/src/static/render-map.generated.ts`

The generator runs `toDOM` with default (null) attrs and serializes the result, leaking nulls into the static map:
- `textStyle` → `"style": "color: null"` (would render `style="color: null"`)
- `anchor` → `"id": null`
- `blok` → `"data-blok": "null"`

Strip null-valued attrs in `extractDOMSpec` before serializing.

**7. `highlight` type/style mismatch**

`STYLE_MAP` maps `highlight.color → backgroundColor`, but generated types show `highlight: Record<string, never>` (no attributes). Either enable `multicolor` in the Highlight extension config so the schema reflects the `color` attr, or remove the dead STYLE_MAP entry.

**8. `.ts` extension in generated import**

`render-map.generated.ts:2` — `from './types.generated.ts'` uses a `.ts` extension. All other imports in the codebase are extensionless. The generator at `richtext-render-map.ts:84` produces this.

**9. `@tiptap/extension-text-align` version inconsistency**

Pinned at `^3.22.3` while all other `@tiptap/extension-*` packages use `^3.10.2`. Also missing from `tiptapGlobals` in `tsdown.config.ts`, which causes `MISSING_GLOBAL_NAME` build warnings on UMD bundles.

---

### Minor

**10. `heading` `data-level` attribute leaks via `DEFAULT_ATTR_MAP`**

`processAttrs('heading', { level: 2, textAlign: 'right' })` outputs `{ "data-level": 2, style: { textAlign: "right" } }`. The `level` is already consumed by `resolveTag` — the extra `data-level` HTML attribute is likely unintended.

**11. `doc` resolves to `<span>`**

`resolveTag({ type: 'doc' })` returns `"span"`. The `doc` node is the root container and shouldn't render a wrapper element. Consumers need to special-case it, which isn't documented.

**12. `ordered_list` types use `any` for `start` and `type`**

Generated as `ordered_list: { start?: any; type?: any; order?: number; }`. Add type hints for `start` (`number`) and `type` (list-style-type values).

**13. `ordered_list` renders `order` as HTML attribute**

HTML `<ol>` only understands `start`; `order` passes through as noise in the markup.

**14. `genAttributeTypes` called twice — `richtext-type.ts:68-72`**

Computes the same result twice. Call once and destructure.

**15. Unused `_def` parameter — `richtext-type.ts:13`**

**16. `parseArraySpec` try/catch rethrows — `parseDOMSpec.ts:49-63`**

The catch block logs and rethrows. Either handle gracefully or remove the try/catch.

**17. Missing JSDoc on exported functions**

`resolveComponent`, `resolveTag`, `isSelfClosing`, `getStaticChildren`, `styleToString`, `stringToStyle` are all exported without docs.

**18. `SbBlokData` potentially duplicated**

Defined locally in `types.ts`. Check if it can be imported from a shared Storyblok package.

---

### Observations

- **Generated files committed to git** — Fine for DX, but PRs that change extensions must regenerate. Consider a CI check that runs `generate` and asserts no diff.
- **PR scope** — 940 additions across 19 files bundles extension changes, codegen, runtime utils, and build config. Consider splitting.
- **No docs** — No README or usage documentation for the new `./static` entry point.

maoberlehner (changes requested):
Looking forward to merging this PR, will make our users very happy I think!

minor nitpick: branch is named `feat/*`, I think our convention is `feature/*`. I don't really mind, because the branch disappears after squashing anyway. Just something to keep in mind.

maoberlehner:
## QA Findings (via Claude Code)

> Ran locally against the branch. Build, generate, lint, typecheck, and all 162 existing tests pass. Bundle: 5.5kb raw / ~2kb gzip (ESM).

---

### Automated Checks

| Check | Result |
|-------|--------|
| `pnpm run generate` | Pass (benign BLOK resolver warning) |
| Generate idempotent | Pass — no diff against committed files |
| `pnpm nx build @storyblok/richtext` | Pass — all dist files produced |
| `pnpm nx test @storyblok/richtext` | **162 tests pass** (4 test files) |
| `pnpm nx lint @storyblok/richtext` | Pass |
| `npx tsc --noEmit` | Pass |
| ESM bundle gzip | **1,991 bytes** (under 2kb target) |
| CJS bundle gzip | ~2,070 bytes (slightly over 2kb) |
| Export map (`./static`) | Correct — `import` + `require` entries |
| `typesVersions` | Correct — `static` → `./dist/static.d.cts` |

Build produces a `MISSING_GLOBAL_NAME` warning for `@tiptap/extension-text-align` on UMD bundles (missing from `tiptapGlobals` in `tsdown.config.ts`).

---

### Runtime Utility Tests

Tested every exported function from `@storyblok/richtext/static` against the built dist.

#### `resolveTag` — all node and mark types

| Input | Output | Correct? |
|-------|--------|----------|
| `paragraph` | `p` | Yes |
| `blockquote` | `blockquote` | Yes |
| `heading` (level 1) | `h1` | Yes |
| `heading` (level 3) | `h3` | Yes |
| `heading` (no attrs) | `h1` | Yes |
| `bullet_list` | `ul` | Yes |
| `ordered_list` | `ol` | Yes |
| `list_item` | `li` | Yes |
| `code_block` | `pre` | Yes |
| `hard_break` | `br` | Yes |
| `horizontal_rule` | `hr` | Yes |
| `image` | `img` | Yes |
| `emoji` | `span` | Yes |
| `table` | `table` | Yes |
| `tableRow` | `tr` | Yes |
| `tableCell` | `td` | Yes |
| `tableHeader` | `th` | Yes |
| `blok` | `span` | Yes |
| `details` | `details` | Yes |
| `detailsContent` | `div` | Yes |
| `detailsSummary` | `summary` | Yes |
| All 13 mark types | Correct tags | Yes |
| Unknown type | `span` (fallback) | Yes |
| `doc` | `span` | Questionable — see code review |
| `text` | `span` | Questionable — see code review |

**Edge cases:**

| Input | Output | Issue? |
|-------|--------|--------|
| `heading` level 0 | `h1` (falsy fallback) | OK |
| `heading` level 7 | `h7` | No clamping — invalid HTML |
| `heading` level -1 | `h-1` | No clamping — invalid HTML |
| `heading` level null | `h1` | OK |
| `heading` level `'<script>'` | `h<script>alert(1)</script>` | **XSS — see code review** |
| `null` node | Throws | No guard |
| `undefined` node | Throws | No guard |
| Empty type `''` | `span` | OK |

#### `processAttrs` — style mapping, attr mapping, edge cases

| Input | Output | Correct? |
|-------|--------|----------|
| `paragraph` + `textAlign: 'center'` | `{ style: { textAlign: 'center' } }` | Yes |
| `paragraph` + `textAlign: null` | `{}` | Yes |
| `highlight` + `color: '#ff0'` | `{ style: { backgroundColor: '#ff0' } }` | Yes |
| `textStyle` + `color: '#f00'` | `{ style: { color: '#f00' } }` | Yes |
| `tableCell` + colwidth/colspan/rowspan/bg | Correct style + attr mapping | Yes |
| `tableHeader` + `colwidth: [150]` | `{ style: { width: '150px' } }` | Yes |
| `tableHeader` + `colwidth: null` | `{}` | Yes |
| `emoji` + `fallbackImage` | Mapped to `src` | Yes |
| `blok` + `body` array | JSON-stringified to `data-body` | Yes |
| `bold` + `{}` | `{}` | Yes |
| `bold` + `undefined` | `{}` | Yes |
| Unknown attr passthrough | `{ unknownAttr: 'value' }` | Yes |
| `extendAttrMap` override | Works correctly | Yes |
| Nested object values | JSON-stringified | Yes |

**Issues found:**

| Input | Output | Problem |
|-------|--------|---------|
| `textAlign: ''` | `{ style: { textAlign: '' } }` | Falsy value leaks |
| `textAlign: false` | `{ style: { textAlign: false } }` | Falsy value leaks |
| `textAlign: 0` | `{ style: { textAlign: 0 } }` | Falsy value leaks |
| `colwidth: [null]` | `{ style: {} }` | Empty style object leaks |
| `colwidth: []` | `{ style: {} }` | Empty style object leaks |
| `heading` + `level: 2` | `{ "data-level": 2 }` | Unintended `data-level` attr |
| `{ constructor: 'x' }` | `{ style: { "function Object()...": "x" } }` | Prototype key leakage |

#### `styleToString` / `stringToStyle`

| Test | Result | Correct? |
|------|--------|----------|
| Round-trip `{ backgroundColor: '#ff0', textAlign: 'center' }` | Preserves values | Yes |
| Empty object | `""` | Yes |
| Empty string | `{}` | Yes |
| Whitespace-only string | `{}` | Yes |
| Trailing semicolon | Handled | Yes |
| No semicolon | Handled | Yes |
| Multiple semicolons | Handled | Yes |
| Leading/trailing spaces | Trimmed | Yes |
| Newlines in string | Handled | Yes |
| No colon (invalid) | `{}` | Yes |
| Empty key/value | `{}` | Yes |

**Issues found:**

| Test | Result | Problem |
|------|--------|---------|
| `styleToString({ color: null })` | `"color: null"` | Literal `null` string |
| `styleToString({ color: undefined })` | `"color: undefined"` | Literal `undefined` string |
| `stringToStyle('background: url(https://example.com)')` | `{ background: 'url(https' }` | **URL truncated at colon** |
| `stringToStyle('content: "hello: world"')` | `{ content: '"hello' }` | Colon in value breaks parsing |

#### `isSelfClosing`

All correct: `br`, `hr`, `img` → `true`; `p`, `span`, `div` → `false`.

#### `getStaticChildren`

| Node | Children | Correct? |
|------|----------|----------|
| `table` | `[{ tag: 'tbody', content: true }]` | Yes |
| `code_block` | `[{ tag: 'code', content: true }]` | Yes |
| `emoji` | `[{ tag: 'img', attrs: { style, draggable, loading } }]` | Yes |
| All other nodes | `null` | Yes |
| All mark types | `null` | Yes |
| Unknown type | `null` | Yes |
| `null` input | Throws | No guard |

#### `resolveComponent`

Works correctly: matches from component map, returns `undefined` for missing keys and when no map is provided.

---

### Security Tests

| Test | Result | Severity |
|------|--------|----------|
| `heading` level XSS injection | Tag name is unsanitized | **Critical** |
| `javascript:` in `href` | Passes through unfiltered | Note (framework responsibility) |
| Attribute injection in `class` | Passes through as string | Note (framework escapes) |
| Style injection in `textAlign` | Passes through as style value | Low |
| Prototype pollution via `__proto__` | No effect (empty object) | OK |
| `constructor`/`toString`/`valueOf` keys | Garbled output in style | Low |

maoberlehner:
> ## Claude Stress Test — `@storyblok/richtext` static entry point
> > Review by Claude (Opus), run locally against the branch. Verified: build, generate, tests (162 pass), lint, typecheck all green. Bundle: 5.5kb raw / ~2kb gzip (ESM).
> 
> ### Critical
> **1. Heading `resolveTag` — unsanitized template literal allows XSS**
> 
> The heading resolve function is:
> ...

@dipankarmaikap take those findings with a grain of salt. Might not all be relevant! E.g., I don't think the XSS vector applies to us. Still, it would not hurt to convert the H levels to numbers.

packages/richtext/src/static/render-map.generated.ts
maoberlehner: Ah, I think you told me about it during our call. But I forgot what the reason was. Please just extend the comment and include info why we need to include them.

packages/richtext/src/static/generate/parse-dom-spec.ts
```diff
@@ -1,49 +1,67 @@
...
 
+/** Parses a ProseMirror DOMOutputSpec into a RenderSpec. */
 export function parseDOMSpec(spec: DOMOutputSpec): RenderSpec | null {
   // string
   if (typeof spec === 'string') {
     return { tag: spec };
   }
   // custom override
-  if (Array.isArray(spec) && typeof spec[0] === 'string') {
+  if (isArraySpec(spec)) {
```
maoberlehner, line 29: 👍

maoberlehner:
@dipankarmaikap please evaluate how relevant those are. I realized that I don't understand the API surface of the `/static` renderer well enough to make a judgment call. Feel free to discard!

## Claude Code Functional QA Findings

Validated against the built `@storyblok/richtext/static` bundle on this branch. Scope: behavioral gaps in the shipped API (structural / test / doc issues are tracked separately in the code review).

### Major

**F1. Render-map default attrs are silently dropped by `resolveTag`.**

```js
import { resolveTag, getStaticChildren } from '@storyblok/richtext/static';

// blok — private data leaks into DOM
resolveTag({ type: 'blok', attrs: {} });          // → 'span'
// render map has: { tag: 'span', attrs: { style: 'display: none' } }  ← dropped

// emoji — outer wrapper attrs lost
resolveTag({ type: 'emoji', attrs: {} });         // → 'span'
// render map has: attrs: { 'data-type': 'emoji' }                      ← dropped

// detailsContent — CSS hook lost
resolveTag({ type: 'detailsContent', attrs: {} }); // → 'div'
// render map has: attrs: { 'data-type': 'detailsContent' }            ← dropped

// ordered_list — default start/order only present if the JSON carries them
resolveTag({ type: 'ordered_list', attrs: {} });   // → 'ol'
// render map has: attrs: { start: 1, order: 1 }                       ← dropped
```

`NODE_RENDER_MAP` / `MARK_RENDER_MAP` are not re-exported either (`typeof api.NODE_RENDER_MAP === 'undefined'`), so consumers have no way to recover these defaults. **Fix:** return `{ tag, attrs, children }` from `resolveTag`, add a `resolveAttrs(node)`, or re-export the maps — and document the intended pattern.

**F2. Emoji rendering is incorrect.**

```js
import { processAttrs } from '@storyblok/richtext/static';

processAttrs('emoji', {
  name: 'smile',
  emoji: '😀',
  fallbackImage: 'https://cdn/smile.png',
});
// Actual
// → { name: 'smile', emoji: '😀', src: 'https://cdn/smile.png' }
//
// Expected (matching non-static StoryblokEmoji.renderHTML)
// <span data-type="emoji" data-name="smile" data-emoji="😀">
//   <img src="https://cdn/smile.png" draggable="false" loading="lazy" style="…" />
// </span>
```

Three problems: no `name → data-name` / `emoji → data-emoji` mapping in `DEFAULT_ATTR_MAP`; `fallbackImage → src` lands on the outer `<span>` because there is no slot mechanism to route it to the inner ``; and the wrapper's `data-type="emoji"` is dropped (see F1).

### Moderate

**F3. `code_block.language` cannot reach the inner `<code>`.**

```js
import { processAttrs } from '@storyblok/richtext/static';

processAttrs('code_block', { language: 'js' });
// → { language: 'js' }       (naive consumer puts this on <pre>)
//
// Non-static StoryblokCodeBlock.renderHTML produces:
// <pre><code class="language-js">…</code></pre>
```

Static and non-static output diverge because there is no mechanism to route a node-level attribute onto a named static child.

**F4. Storyblok-only link metadata leaks to rendered HTML.**

```js
import { processAttrs } from '@storyblok/richtext/static';

processAttrs('link', {
  href: 'https://example.com',
  uuid: 'abc-123',
  linktype: 'url',
  custom: { rel: 'nofollow', target: '_blank' },
});
// Actual
// → {
//     href: 'https://example.com',
//     uuid: 'abc-123',
//     linktype: 'url',
//     custom: '{"rel":"nofollow","target":"_blank"}',
//   }
//
// Renders as: <a href="…" uuid="abc-123" linktype="url" custom='{"rel":…}'>
// Expected:   <a href="…" rel="nofollow" target="_blank">
```

The non-static resolver treats `linktype`/`uuid` as routing metadata and spreads `custom` into real HTML attributes. The static API should normalize to match (or document that consumers must strip these).

**F5. Redundant `colSpan="1"` / `rowSpan="1"` / `data-level` emitted.**

```js
import { processAttrs, resolveTag } from '@storyblok/richtext/static';

processAttrs('tableCell', {
  colspan: 1, rowspan: 1, colwidth: [200], backgroundColor: '#eee',
});
// → { colSpan: 1, rowSpan: 1, style: { width: '200px', backgroundColor: '#eee' } }
//   (colSpan=1 / rowSpan=1 are redundant)

resolveTag({ type: 'heading', attrs: { level: 2 } });         // → 'h2'
processAttrs('heading', { level: 2, textAlign: 'right' });
// → { 'data-level': 2, style: { textAlign: 'right' } }
//   (data-level="2" duplicates information already encoded in <h2>)
```

### Minor

**F6. `reporter` mark wraps text in a stray `<span>`.**

```js
import { resolveTag } from '@storyblok/richtext/static';

resolveTag({ type: 'reporter', attrs: {} });  // → 'span'
// MARK_RENDER_MAP.reporter === null          ← silent-span fallback
```

`reporter` is a validation-only mark. The API should signal "skip" (e.g. return `null`/`undefined`) so consumers can drop it rather than emit `<span>text</span>`. Same concern applies to any future `null` entry.

**F7. No `STYLE_MAP` extension point.**

```js
import { processAttrs } from '@storyblok/richtext/static';

// processAttrs signature today: (type, attrs, extendAttrMap?)
// There is no `extendStyleMap` — so a framework adapter cannot register
// e.g. paddingY → { paddingTop, paddingBottom } without monkey-patching.
processAttrs('paragraph', { paddingY: 8 });
// → { paddingY: 8 }  (lands as HTML attr, not inline style)
```

### Positive / Confirmed working

```js
import {
  resolveTag, resolveComponent, isSelfClosing,
  processAttrs, stringToStyle, styleToString,
} from '@storyblok/richtext/static';

// Heading dynamic resolver — clamps to h1..h6
resolveTag({ type: 'heading', attrs: { level: 0 } });    // 'h1'
resolveTag({ type: 'heading', attrs: { level: 100 } });  // 'h6'
resolveTag({ type: 'heading', attrs: {} });              // 'h1'

// Attribute mappings
processAttrs('tableCell', { colspan: 2, rowspan: 3, colwidth: [200] });
// → { colSpan: 2, rowSpan: 3, style: { width: '200px' } }
processAttrs('highlight', { color: 'yellow' });
// → { style: { backgroundColor: 'yellow' } }
processAttrs('paragraph', { textAlign: 'center' });
// → { style: { textAlign: 'center' } }

// Style round-trip
stringToStyle('background-color: red; font-size: 12px;');
// → { backgroundColor: 'red', fontSize: '12px' }
styleToString({ textAlign: 'center', backgroundColor: 'red' });
// → 'text-align: center; background-color: red'

// Self-closing detection
isSelfClosing('img');  // true
isSelfClosing('p');    // false

// Component map
resolveComponent('paragraph', { paragraph: MyP });   // → MyP
resolveComponent('nonexistent', { paragraph: MyP }); // → undefined
```

Plus: generator output is byte-stable across repeated runs (sha256 verified); `attw` and `publint` report no regressions; ESM and CJS entry points both expose the 7 expected runtime helpers without pulling `prosemirror-model` into the runtime import graph; 196 unit tests pass; lint clean.

packages/richtext/src/static/util.ts
```diff
@@ -0,0 +1,46 @@
...
+  TComponent,
+  ExtraProps extends Record<string, unknown> = Record<string, never>,
+>(
+  type: K,
+  components?: StoryblokRichTextComponentMap<TComponent, ExtraProps>,
+): StoryblokRichTextComponentMap<TComponent, ExtraProps>[K] | undefined {
+  return components?.[type];
+}
+
+export function resolveTag(node: PMNode | PMMark): string {
```
maoberlehner: Claude Code:

resolveTag returns only the tag string, getStaticChildren returns only children, and neither NODE_RENDER_MAP nor MARK_RENDER_MAP is re-exported. Confirmed via
  wdx-331-verify-findings.mjs:
    - resolveTag({ type: 'blok' }) → 'span'; render map says attrs: { style: 'display: none' } (dropped).
    - resolveTag({ type: 'emoji' }) → 'span'; render map says attrs: { 'data-type': 'emoji' } (dropped).
    - resolveTag({ type: 'detailsContent' }) → 'div'; render map says attrs: { 'data-type': 'detailsContent' } (dropped).
    - resolveTag({ type: 'ordered_list' }) → 'ol'; render map says attrs: { start: 1, order: 1 } (dropped).
    - typeof api.NODE_RENDER_MAP === 'undefined', typeof api.MARK_RENDER_MAP === 'undefined'.

The "static renderer" is not a complete rendering kit without this surface. Fix: return { tag, attrs, children } from resolveTag, expose resolveAttrs(node), or re-export the maps — and document the intended pattern.

packages/richtext/src/static/util.ts
```diff
@@ -0,0 +1,46 @@
+import { MARK_RENDER_MAP, NODE_RENDER_MAP } from './render-map.generated';
```
maoberlehner, line 1: Claude Code:

No mechanism to route a parent-node attribute onto a named static child.
Blocks correct rendering for two built-in nodes — emoji (fallbackImage → child `.src`) and code_block (language → child `<code>.class="language-${language}"`).

Without a slot mechanism, consumers of the static API cannot reproduce the output of the non-static resolver. Fix: introduce a childIndex hint in RenderSpec's children or a per-node rewrite hook; update `DEFAULT_ATTR_MAP` accordingly; document.

packages/richtext/package.json
```diff
@@ -142,6 +154,25 @@
...
       "!@storyblok/playground-*"
-    ]
+    ],
+    "targets": {
+      "generate": {
+        "dependsOn": [
+          "^build"
+        ],
+        "outputs": [
+          "{projectRoot}/src/static/**"
```
maoberlehner: Claude Code:

NX outputs: `["{projectRoot}/src/static/**"]` is over-broad.
In my local test the default project inputs protected hand-written files, but the pattern still declares hand-written code as generate output, which is a latent footgun if inputs is ever narrowed or shared cache diverges. Fix: scope outputs to {projectRoot}/src/static/{render-map,types}.generated.ts.

packages/richtext/src/static/style.ts
```diff
@@ -0,0 +1,141 @@
+import type { AttrValue } from './generate/parse-dom-spec';
```
maoberlehner: I don't think we should import from `./generate`. As I understand it,`generate` scripts are safe to use `devDependencies` but only if we don't rely on them in user-facing code.

Claude Code:

prosemirror-model dev-dep leaks into the public module's import graph. (imports AttrValue from ./generate/parse-dom-spec), and parse-dom-spec.ts imports  DOMOutputSpec from prosemirror-model. attw and CJS require both pass today, so runtime is clean — but the coupling is fragile. Fix: move AttrValue into static/types.ts so the public module never transitively references generate/.

packages/richtext/src/static/util.ts
```diff
@@ -0,0 +1,46 @@
...
+  ExtraProps extends Record<string, unknown> = Record<string, never>,
+>(
+  type: K,
+  components?: StoryblokRichTextComponentMap<TComponent, ExtraProps>,
+): StoryblokRichTextComponentMap<TComponent, ExtraProps>[K] | undefined {
+  return components?.[type];
+}
+
+export function resolveTag(node: PMNode | PMMark): string {
+  const type = node.type as string;
```
maoberlehner: `as` necessary here?

packages/richtext/package.json
```diff
@@ -95,6 +103,7 @@
     "@tiptap/extension-superscript": "^3.10.2",
     "@tiptap/extension-table": "^3.10.2",
     "@tiptap/extension-text": "^3.10.2",
+    "@tiptap/extension-text-align": "^3.10.2",
```
maoberlehner: The problem highlighted by Claude Code is that in `pnpm-lock.yaml` the new extension was resolved to the latest version while the others stayed on the locked 3.20.0. It's probably a non-issue, but you could update all to the latest version while you're at it. Or you leave it as is.

Claude Code:

Version drift on @tiptap/extension-text-align. Specifier ^3.10.2, resolved to 3.22.3; other tiptap extensions resolved to 3.20.0. Dedupe to avoid multi-instance @tiptap/core.

maoberlehner (changes requested):
Already looks pretty good to me! The only blocker for me is the import from `./generate`.  Please evaluate how right or wrong Claude is about its findings and feel free to ignore them.

dipankarmaikap:
Hey @maoberlehner , I’ve updated the Tiptap packages to the latest version and fixed a few small things you mentioned. I also added a `richTextRenderer` function that converts Storyblok JSON to HTML.

This is useful for users who don’t need customisation or aren’t using things like `blok`. They can simply use it like in React with `dangerouslySetInnerHTML` and render the output directly.

For each framework, we’ll still provide a dedicated `RichText` component that supports full customisation.

Let me know what you think about the `richTextRenderer` approach, including the name or any other suggestions.

packages/richtext/src/static/util.ts
dipankarmaikap: This will be implemented at the framework level. A RichText rendering component will be added for each framework.

packages/richtext/src/static/util.ts
dipankarmaikap: This will be implemented at the framework level. `blok` nodes will be resolved out of the box, and nodes with static children will also be handled within the framework component. `resolveTag` is not something end users will interact with directly. Instead, we’ll provide a RichText component that uses `resolveTag` internally.

We also provide a `richTextRenderer` for users who simply want to generate HTML without using a framework component.

maoberlehner (approved):

PR #565, opened 2026-04-14
fix(richtext): prevent Vue slot warning for component tags

### LINKED ISSUE #559: [Bug]: Non-function value encountered for default slot. Prefer function slots for better performance.
### Package

@storyblok/nuxt

### Bug Description

When parsing links from richtext as NuxtLink component with the guide from:
https://www.storyblok.com/docs/libraries/js/nuxt-sdk#example-custom-link-with-nuxtlink
The console get's a warning for every link:
"[Vue warn]: Non-function value encountered for default slot. Prefer function slots for better performance."
It comes from the 0 of this line:
`return [asTag(NuxtLink), { to: HTMLAttributes.href }, 0]`
The 0 is for the value inside the link. So the text which has to be rendered.
So there must be a way to write something like this to don't get the warning anymore:
`return [asTag(NuxtLink), { to: HTMLAttributes.href },  () => 0]`
But that doesn't work.
In the old richtext API I could make this: 
`() => node.text`
But now with the changed API it seems that the only way to access the text inside the link is to use this 0. But there has to be a solution to output the link-text while using it in a function.

### Steps to Reproduce

1. create storyblok space and create page with links inside richtext element
2. fetch this richtext with the @storyblok/nuxt package inside Nuxt frontend
3. use this guide to parse links as NuxtLinks component: https://www.storyblok.com/docs/libraries/js/nuxt-sdk#example-custom-link-with-nuxtlink
4. see warnings in console

### Reproduction URL

_No response_

### Expected Behavior

Code-Snippets from official documentation should not throw warning because not using a function
Something like this should work: 
`return [asTag(NuxtLink), { to: HTMLAttributes.href },  () => 0]`

### Actual Behavior

Code-Snippets from official documentation throws warning because not using a function
Just this works without function:
`return [asTag(NuxtLink), { to: HTMLAttributes.href }, 0]`

### Environment

```shell
System:
    OS: macOS 26.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 63.94 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
Binaries:
    Node: 20.20.2 - /opt/homebrew/opt/node@20/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.8.2 - /opt/homebrew/opt/node@20/bin/npm
    pnpm: 10.33.0 - /opt/homebrew/bin/pnpm
    bun: 1.3.11 - /opt/homebrew/bin/bun
Browsers:
    Chrome: 146.0.7680.178
    Firefox: 149.0.2
    Safari: 26.3.1
npmPackages:
    @storyblok/nuxt: 10.1.2 => 10.1.2 
    nuxt: 4.4.2 => 4.4.2
```

### Error Logs

```shell
[Vue warn]: Non-function value encountered for default slot. Prefer function slots for better performance.
```

### Additional Context

_No response_

maoberlehner:
## Summary

- Wraps children in a slot function when `contextRenderFn` detects a component tag (via `typeof tag !== 'string'`), preventing Vue's "Non-function value encountered for default slot" warning
- Adds SSR-based tests that verify the warning no longer fires for both mark components (e.g. `RouterLink`) and node components (e.g. custom `AlertNode`)

Closes #559

## Test plan

- [x] Existing 162 richtext tests pass
- [x] 2 new tests pass (slot warning for component marks + nodes)
- [x] Red-green verified: tests fail without the fix, pass with it

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,4 @@ vite.config.ts.timestamp-*
 .cursor
 .zed
 .rules
+.worktrees
--- a/packages/richtext/src/richtext.test.ts
+++ b/packages/richtext/src/richtext.test.ts
@@ -3,8 +3,9 @@ import { richTextResolver } from './richtext';
 import { getRichTextSegments } from './richtext-segment';
 import { renderSegments } from './render-segments';
 import type { RendererAdapter } from './render-segments';
-import { createTextVNode, h } from 'vue';
+import { createSSRApp, createTextVNode, h } from 'vue';
 import type { VNode } from 'vue';
+import { renderToString } from 'vue/server-renderer';
 import type { StoryblokRichTextNode } from './types';
 import { Mark, Node } from '@tiptap/core';
 import Heading from '@tiptap/extension-heading';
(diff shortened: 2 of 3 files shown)


dipankarmaikap (approved):
LGTM 🚀

PR #569, opened 2026-04-15
refactor(cli): remove change detection from assets push

maoberlehner:
## Summary

- Remove SHA-256 hash comparison and remote binary download from `assets push` — always upload unconditionally
- Remove `--asset-token` option from push (was only needed to download private assets for comparison)
- Remove skipped-asset tracking and reporting (assets are never skipped)
- Add ADR-0003 documenting the decision and alternatives considered

## Context

`assets push` previously downloaded the full remote binary for every asset to SHA-256-compare it against the local file. For a 5 MB asset this meant downloading 5 MB just to decide whether to upload 5 MB. Investigation confirmed that re-uploading preserves the asset URL, so there is no orphan-URL risk.

## Test plan

- [x] All 34 existing push tests pass (3 removed, 3 simplified, rest updated for removed `skipped` state)
- [x] Type checking passes (`pnpm test:types`)
- [x] Lint passes (`pnpm lint:fix`)
- [ ] Manual: push a single asset and verify it uploads without downloading the remote binary
- [ ] Manual: push an asset that already exists remotely and verify it gets re-uploaded

Fixes WDX-300

### DIFF
--- a/adr/0003-remove-change-detection-from-assets-push.md
+++ b/adr/0003-remove-change-detection-from-assets-push.md
@@ -1,0 +1,29 @@
+# ADR-0003: Remove Change Detection from `assets push`
+
+**Status:** Accepted  
+**Date:** 2026-04-15
+
+## Context
+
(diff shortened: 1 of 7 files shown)


dipankarmaikap (approved):
LGTM 🚀 Nice improvement!

PR #570, opened 2026-04-15
fix(cli): make --path a global option and fix path resolution

### LINKED ISSUE #414: 🐛 [Bug]: --path option: makes dot dissapear + CLI help has not
### Package

storyblok (CLI)

### Bug Description

We found a couple of problems in terms of `--path` option.

1. The dot in the path dissapears

In the documentation, the default `--path` value is `.storyblok/migrations`, but when running that in console:
```
storyblok migrations run <name> --space 12345 --path .storyblok/migrations --dry-run
```

We get that it search for folder WITHOUT DOT ⚠️ 

Also, since in the code [here](https://github.com/storyblok/monoblok/blob/main/packages/cli/src/commands/migrations/run/actions.ts#L13)  you already add `migrations`, it all results in path: `/storyblok/migrations/migrations`.

This means either the default value should be `.storyblok` (and not `.storyblok/migrations`) or code should be changed.

NOTE: as a user, I assume `--path` is full path to the folder, but it is not, it is rather acting like `--base-path` or `--path-prefix` and becomes missleading when code adds anything extra..

2. While there is `-p, --path <path>` documented in migration run [options](https://github.com/storyblok/monoblok/tree/main/packages/cli/src/commands/migrations/run#options) in markdown, it is unfortunatelly missing.

Image

### Steps to Reproduce

Image

### Expected Behavior

--path is documented and dot is not stripped away

### Actual Behavior

1. --path is not documented in CLI help
2. dot is not stripped away

### Code Sample

```javascript

```

### Environment

```shell
.
```

### Error Logs

```shell

```

### Additional Context

_No response_

maoberlehner:
## Summary

- Promote `--path` from a per-command option (duplicated across 18 commands) to a single global option, ensuring it consistently appears in CLI help
- Fix inconsistent path resolution in `migrations generate` that used raw `resolve()` instead of `resolvePath()`
- Improve `--path` description to clarify it is a base directory (not the full path)
- Ensure CLI `--path` overrides always take precedence over module-level config values

Fixes WDX-250
Fixes #414

## Test plan

- [x] All 625 CLI tests pass (`pnpm vitest run`)
- [x] Type checking passes (`pnpm test:types`)
- [x] Linting passes (`pnpm lint:fix`)
- [x] Manual: `storyblok migrations run --help` shows `--path` in global options
- [x] Manual: `storyblok migrations run <name> --space <id> --path .custom-dir` resolves path with dot preserved

### DIFF
--- a/packages/cli/src/commands/assets/pull/index.ts
+++ b/packages/cli/src/commands/assets/pull/index.ts
@@ -22,7 +22,6 @@ import {
 const pullCmd = assetsCommand
   .command('pull')
   .option('-s, --space <space>', 'space ID')
-  .option('-p, --path <path>', 'path for file storage')
   .option('-d, --dry-run', 'Preview changes without applying them to Storyblok')
   .option('-q, --query <query>', 'Filter assets using Storyblok filter query syntax. Example: --query="search=my-file.jpg&with_tags=tag1,tag2"')
   .option('--asset-token <token>', 'Asset token for accessing private assets')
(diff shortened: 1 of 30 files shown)


dipankarmaikap (approved):
LGTM 🚀

PR #574, opened 2026-04-20
fix(cli): paginate component and internal-tag list fetches

maoberlehner:
`storyblok components pull` silently dropped list resources past the first page because the MAPI `.list()` calls in `components/pull/actions.ts` were issued without pagination. With the default page size of 25, spaces that had more than 25 component internal tags (or >25 components) lost everything on later pages. A later `components push` then failed on any component that referenced a dropped tag with HTTP 422
`internal_tag_ids: Invalid internal_tag, there is at least one internal_tag not found in our database`.

Extract the `fetchAllPages` helper from `datasources/pull/actions.ts` into a shared `utils/pagination.ts` (iterative, single-allocation), and apply it to `fetchComponents`, `fetchComponent`, and `fetchComponentInternalTags`. The internal-tag fetch now also pushes `by_object_type: 'component'` into the MAPI query, so filtering happens server-side. Component folders and presets are left untouched because their MAPI endpoints are not paginated.

Adds direct unit tests for `fetchAllPages` and multi-page regression tests for `fetchComponents` and `fetchComponentInternalTags`.

Fixes WDX-397

### DIFF
--- a/packages/cli/src/commands/datasources/pull/actions.ts
+++ b/packages/cli/src/commands/datasources/pull/actions.ts
@@ -2,41 +2,10 @@ import { handleAPIError, handleFileSystemError } from '../../../utils';
 import { getMapiClient } from '../../../api';
 import { join, resolve } from 'pathe';
 import { resolvePath, sanitizeFilename, saveToFile } from '../../../utils/filesystem';
+import { fetchAllPages } from '../../../utils/pagination';
 import type { DatasourceEntry, SpaceDatasource } from '../constants';
 import { DEFAULT_DATASOURCES_FILENAME } from '../constants';
 import type { SaveDatasourcesOptions } from './constants';
-
-/**
- * Generic pagination helper that fetches all pages of data
- * @param fetchFunction - Function that fetches a single page
- * @param extractDataFunction - Function that extracts data array from response
- * @param page - Current page number
- * @param collectedItems - Previously collected items
- * @returns Array of all items across all pages
- */
-async function fetchAllPages<T, R>(
-  fetchFunction: (page: number) => Promise<{ data: T; response: Response }>,
-  extractDataFunction: (data: T) => R[],
-  page = 1,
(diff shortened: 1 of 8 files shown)


packages/cli/src/utils/pagination.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Fetches every page of a list endpoint and returns the combined items.
+ * Stops when the `total` response header is reached, when a page returns
+ * no items, or immediately if the header is absent (treats the response
+ * as un-paginated).
+ *
+ * @param fetchFunction - Fetches a single page; receives the 1-based page number
+ * @param extractDataFunction - Extracts the items array from a page's response data
+ */
+export async function fetchAllPages<T, R>(
```
alexjoverm, line 10: praise: great refactor here!

packages/cli/src/commands/datasources/pull/actions.ts
alexjoverm: suggestion(non-blocking): @maoberlehner might be worth to add pagination here as well, similar to the `fetchComponent`. It's very unlikely, but as this endpoint uses `search`, it could happen it return 25+ results

alexjoverm: Similarly, across different commands in `src/commands/**`, there are 3 extra endpoints not being paginated: `componentFolders.list`, `presets.list`, `assetFolders.list`.

Might be worth to fix them on the go, but also fine if you prefer to do it separately

alexjoverm (approved):
LGTM

packages/cli/src/commands/datasources/pull/actions.ts
maoberlehner: Added `fetchAll` for datasources. It seems other endpoints are not paginated.

Claude: Backend check confirmed componentFolders.list, presets.list, and assetFolders.list are not paginated server-side — all use .all / unscoped queries. Only datasources.list calls paginate(...).

PR #576, opened 2026-04-21
feat(angular): improve angular SDK

dipankarmaikap:
This PR adds comprehensive tests for the Angular SDK.

Rich text rendering now uses the new static rendering approach from the richtext package.

Blok components are now auto-resolved, and every node/mark can be fully customized.

Example mark component:

```ts
@Component({
  selector: 'app-link',
  changeDetection: ChangeDetectionStrategy.OnPush,
  template: `
    <a [href]="data().attrs?.href" target="_blank" rel="noopener">
      <ng-content></ng-content>
    </a>
  `,
  host: { style: 'display: inline-block' },
})
export class LinkComponent {
  readonly data = input.required<RichTextComponentProps<'link'>>();
}
```

Example node component:

```ts
@Component({
  selector: 'app-heading',
  changeDetection: ChangeDetectionStrategy.OnPush,
  template: `
    <p>
      @for (doc of data().content ?? []; track doc) {
        <sb-rich-text [sbDocument]="doc" />
      }
    </p>
  `,
  host: { style: 'display: contents' },
  imports: [SbRichTextComponent],
})
export class HeadingComponent {
  readonly data = input.required<RichTextComponentProps<'heading'>>();
}
```

Also added a new component:

```html
<sb-component [sbBloks]="bloks()" />
```

Previously, users had to write:

```html
@for (blok of bloks(); track blok._uid) {
  <ng-container [sbBlok]="blok" />
}
```

This is still supported, but the new approach is much cleaner and more ergonomic.

Also includes general cleanup and refactoring across the SDK.


alexjoverm (approved):
@dipankarmaikap great work overall here. The new sb-component is a nice DX addition. A couple of notes:

- **Important:** we need to fix the static richtext regressions downstream that I pointed out internally. I approve as here we're in a new package, and still at v0.x.x, but we need them fixed before a v1.0.0. Up to you if you prefer to do it within this PR, or merge this and do it as a follow up 👍 
- Minor: this PR is actually a `feat` as it introduces new functionality.

PR #577, opened 2026-04-21
chore(release): gate packages by per-package release.branches

maoberlehner:
Teach tools/release.mjs to read each package's release.branches and skip packages whose list does not include the current branch. Unify the publish workflow to always derive the publish list from the release commit body, so the versioning step is the single source of truth for what gets published.

- tools/release.mjs: filter --projects by release.branches; error when every requested project is ineligible on the current branch, warn when some are dropped.
- packages/capi-client, packages/mapi-client, packages/migrations: drop "main" from release.branches so stable releases are disabled until these alpha packages are promoted.
- .github/workflows/publish.yaml: remove the main-branch shortcut that published every package; always publish only packages listed in the release commit body. Also harden by moving action outputs into env:.
- RELEASING.md: document the branch-gating rule and mark the three packages as alpha-only.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -37,13 +37,12 @@ jobs:
       - name: Print Environment Info
         run: pnpm nx report
 
+      # Only publish packages that were bumped in the latest release commit.
       # Workaround for Nx bugs:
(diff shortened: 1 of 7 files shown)


dipankarmaikap (approved):
LGTM 🚀

PR #566, opened 2026-04-14
fix(cli): print story identification in push error output

maoberlehner:
## Summary
- When stories fail during `storyblok stories push`, the terminal now prints which story failed (slug, UUID, filename) before the generic error message
- Failed stories are listed in the final push summary with `⚠️ Failed stories:` 
- Failed story labels are included in the JSON report via `reporter.addMeta('failedStories', ...)`

Closes #561

## Test plan
- [x] All 28 existing push tests pass
- [x] New assertions verify story identification appears in error output for all error phases (scan, create, read, map, update)
- [x] New assertions verify `Failed stories:` summary appears when errors occur
- [ ] Manual test: create story files with invalid slugs, run `storyblok stories push`, confirm terminal shows which story failed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #578, opened 2026-04-21
fix(cli): stop pushing empty content and give verbose errors request context

maoberlehner:
## Summary
- Fixes the push bug where stories whose `uuid` contains an underscore were created in Storyblok with empty placeholder content. `readLocalStoriesStream` no longer reconstructs uuids from filenames; the push command resolves `filename → uuid` through the scan index (which reads uuids from file contents).
- Makes missing-uuid local stories fail loudly in the scan instead of collapsing onto an empty-string key in `maps.stories` and silently producing empty remote content.
- Adds request context (`url`, `method`) to `FetchError` and surfaces it in `APIError.getInfo()`, so `--verbose` finally tells you which request produced a 404.

## Context
Customer report: migrating content via `storyblok stories push` created stories with empty content and surfaced no error. Reproduced locally — root cause was `basename(filename).split('_').at(-1)` in `streams.ts`, which returns only the last fragment when the uuid itself contains underscores. Pass 2's filter then missed the uuid in `maps.stories` and silently dropped the file before the real content was uploaded.

The three commits map to the three issues from the report and are independently revertible:
1. `fix(cli): push uploads content for stories whose uuid contains underscores` — primary bug.
2. `fix(cli): fail loudly when a local story is missing a uuid` — the "missing ID" case the customer also sent.
3. `feat(cli): include request url and method in verbose api errors` — better `--verbose` output.

## Overlap with WDX-391
`bugfix/WDX-391-push-stories-missing-story-identification` touches the same callback sites in `push/index.ts`. This PR was written against `main`; if WDX-391 lands first, this will need a small rebase around `push/index.ts`.

## Caveats
- The upstream `ClientError` in `@storyblok/management-api-client` does not currently carry request metadata, so stories-push errors still fall back to the existing `action` message ("Failed to create story") for the URL/method portion. `FetchError`-based paths (login/signup/assets) get the full `request` payload immediately. Extending the mapi-client is tracked separately.

## Test plan
- [x] `pnpm test` in `packages/cli` — 634 tests pass, 1 todo.
- [x] `pnpm test:types` in `packages/cli` — clean.
- [x] `pnpm lint:fix` in `packages/cli` — clean.
- [ ] Manual push against a real space:
  - [x] Story with `"uuid": "legacy_id_123"` lands with full content (not the placeholder).
  - [x] Story missing `uuid` is reported as a failed story and not created.
  - [x] A forced 404 in `--verbose` shows `request: { url, method }` in the error output (for direct-fetch commands; stories-push will still show the action name).

### DIFF
--- a/packages/cli/src/commands/stories/push/index.ts
+++ b/packages/cli/src/commands/stories/push/index.ts
@@ -233,13 +233,16 @@ pushCmd
        * We need to run the second pass in a separate pipeline because we need a
        * complete map of all stories.
        */
+      const uuidByFilename = new Map(storyIndex.map(entry => [entry.filename, entry.uuid]));
+
       await pipeline(
         // Read local stories from `.json` files.
         readLocalStoriesStream({
           directoryPath: storiesDirectoryPath,
-          fileFilter({ uuid }) {
+          fileFilter({ filename }) {
             // Only load files that were successfully created and mapped.
-            return Boolean(maps.stories.get(uuid));
+            const uuid = uuidByFilename.get(filename);
+            return uuid != null && Boolean(maps.stories.get(uuid));
           },
           setTotalStories(total) {
             summary.processResults.total = total;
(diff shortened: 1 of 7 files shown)


dipankarmaikap (approved):
Looks good as per the issue. As you already mentioned, we also need to sync with #566 regarding the error message.

PR #579, opened 2026-04-21
feat(cli): add schema command

maoberlehner:
Implements the `storyblok schema` CLI command with three subcommands:
- `schema push`: Push local schema definitions to Storyblok with diffing, snapshots, and migration generation
- `schema pull`: Pull remote schemas and generate TypeScript definitions
- `schema rollback`: Rollback to a previous schema snapshot

Includes ADRs, test scenarios, playground components, and OpenAPI spec fixes.

Fixes WDX-263


packages/openapi/resources/mapi/components/main.yaml
```diff
@@ -15,8 +15,6 @@ paths:
       summary: Retrieve Multiple Components
       parameters:
         - $ref: ../shared/parameters.yaml#/space_id
-        - $ref: ../shared/pagination.yaml#/page
-        - $ref: ../shared/pagination.yaml#/per_page
```
alexjoverm, line 19: (blocking): @maoberlehner what's the reason to remove these params? Without them, components endpoints won't be able to be paginated

packages/cli/src/utils/pagination.ts
```diff
@@ -0,0 +1,35 @@
...
+): Promise<R[]> {
+  const { data, response } = await fetchFunction(page);
+  const totalHeader = (response.headers.get('total'));
+  const total = Number(totalHeader);
+
+  const fetchedItems = extractDataFunction(data);
+  const allItems = [...collectedItems, ...fetchedItems];
+
+  if (!totalHeader || Number.isNaN(total)) {
+    // No valid 'total' header — assume not paginated, return all collected items plus current page
```
alexjoverm: nitpick: is this comment accurate/needed?

packages/cli/src/commands/schema/push/actions.ts
```diff
@@ -0,0 +1,455 @@
...
+}
+
+/** Fetches remote components, component folders, and datasources from the MAPI. */
+export async function fetchRemoteSchema(spaceId: string) {
+  const client = getMapiClient();
+  const spaceIdNum = Number(spaceId);
+
+  const [componentsRes, foldersRes, rawDatasources] = await Promise.all([
+    client.components.list({ path: { space_id: spaceIdNum }, throwOnError: true }),
+    client.componentFolders.list({ path: { space_id: spaceIdNum }, throwOnError: true }),
```
alexjoverm: question(blocking): @maoberlehner don't we need to paginate these 2 calls as well?

packages/schema/playground/qwik/scripts/cleanup-space.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+const client = createManagementApiClient({
+  personalAccessToken: token,
+  spaceId: Number(spaceId),
+  throwOnError: true,
+});
+
+async function deleteStories() {
+  const { data } = await client.stories.list({ query: { per_page: 100 } });
```
alexjoverm: question(non-blocking): it's a minor as it's a playground, but do we need pagination here?

packages/schema/playground/qwik/scripts/cleanup-space.ts
```diff
@@ -0,0 +1,86 @@
...
+  for (const story of stories) {
+    if (story.id) {
+      await client.stories.delete(story.id);
+      console.info(`  Deleted: ${story.full_slug}`);
+    }
+  }
+}
+
+async function deleteAssets() {
+  const { data } = await client.assets.list({ query: { per_page: 100 } });
```
alexjoverm: same here (pagination)

packages/schema/playground/qwik/scripts/cleanup-space.ts
```diff
@@ -0,0 +1,86 @@
...
+  console.info(`Deleting ${assets.length} asset(s)...`);
+  const ids = assets.map(a => a.id).filter((id): id is number => id !== undefined);
+  if (ids.length > 0) {
+    await client.assets.deleteMany({ body: { ids } });
+    console.info(`  Deleted ${ids.length} asset(s).`);
+  }
+}
+
+async function deleteDatasources() {
+  const { data } = await client.datasources.list({ query: { per_page: 100 } });
```
alexjoverm: suggestion: there are multiple functions (`deleteAssets`, `deleteDatasources`, etc) that they all call `.list` on the resources (unpaginated!).

Aside from fixing the pagination - It's more of an "architectural hygiene" and separation of concerns, but I'd rather suggest the `deleteX` function only does the delete, and the `.list` happens outside of the call, passing the list as an argument to each function.

packages/cli/src/commands/schema/pull/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+import { colorPalette, commands } from '../../../constants';
+import { CommandError, handleError, requireAuthentication, toError } from '../../../utils';
+import { getLogger } from '../../../lib/logger/logger';
+import { getReporter } from '../../../lib/reporter/reporter';
+import { getUI } from '../../../utils/ui';
+import { session } from '../../../session';
+import { schemaCommand } from '../command';
+import type { SchemaPullOptions } from './constants';
+import { fetchRemoteSchema } from '../push/actions';
```
alexjoverm: suggestion(non-blocking): just a minor, but the pull action depending on the push action is not ideal architecturally, could lead to future circular deps. I'd rather extract into a common file (kind of schema/utils.ts, or schema/service.ts) that both import

alexjoverm (changes requested):
Huge effort here @maoberlehner! I added a few comments to polish few things. 

Also, dropping here some general related feedback:

**1. Removed public-api exports in capi-client**

In `main` branch there was in `capi-client/src/index.ts`:

```js
export { createThrottle, parseRateLimitPolicyHeader } from './utils/rate-limit';
export type { RateLimitConfig };
```

Might be intended as we're still under 1.0.0 - but worth mentioning to be conscious as this would mean a breaking change.

**2. Tests in Schema pkg - specially schema/generate.ts**

While schema/generate.ts is part of the build pipeline rather than src, I feel it can be fragile as it's core to the package output. Just one example: if the output of `openapi-zod-client` or anything in the file changes, the pipeline would silently fail.

If you see it useful, we could add at least a few low-mantainance snapshot tests to ensure the generated output stays as expected.

Similarly, the helpers are well covered in unit tests, but the lib is lacking some integration/e2e coverage that ensures it all in combination works together. While we shouldn't test Zod, it could be valuable that the we test "in integration" that a few generated schemas work as expected, using a fixture of an example schema file with all `defineX` functions, mostly checking input/output.


**3. `urlToAsset`/`urlToAssetField` changed default `id` from 0 to 1**

Same as above, maybe intended, but just noticing that this could mean a breaking change. 

**(question) 4. Schema diffing in `schema pull`**

What about schema pull also detecting changes? If the answer is "because it's intended user only do a full code-driven approach and never touch the UI" -> what's the role of schema pull? What about scenarios and workflows where multiple users are defining the schemas architecture, or they're creating/restoring new versions of the space schemas? (as a simple example -> imagine user A pushes a different version than what user B pushed before)

Could conflicts happen? If so, what would be the best way to avoid/tackle them?

packages/openapi/resources/mapi/components/main.yaml
maoberlehner: Components are not paginated: https://github.com/storyblok/storyrails/blob/dc35c27b3e8ea4379d100815d2aa7c8ec2d68756/doc/rswag/v1/mapi.openapi.json#L11161-L11250

packages/cli/src/commands/schema/push/actions.ts
maoberlehner: No, endpoints are not paginated.

packages/schema/playground/qwik/scripts/cleanup-space.ts
maoberlehner: Good point! Added.

packages/schema/playground/qwik/scripts/cleanup-space.ts
maoberlehner: Added.

packages/schema/playground/qwik/scripts/cleanup-space.ts
maoberlehner: Done.

packages/cli/src/commands/datasources/push/README.md
```diff
@@ -1,6 +1,6 @@
 # Datasources Push Command
 
-The `datasources push` command allows you to upload datasources and their entries to your Storyblok space.
+The `datasources push` command allows you to upload datasource entries from local files to your Storyblok space. Datasource definitions themselves are typically owned by `storyblok schema push` in code-driven projects.
```
alexjoverm: heads up (non-blocking): @maoberlehner I believe this change would belong to the Docs updates we need to do as part of this project

packages/cli/test/scenarios/SCENARIOS.md
```diff
@@ -19,3 +19,4 @@ bash .claude/skills/qa-engineer-manual/scripts/seed-scenario.sh --scenario has-s
 | `has-nested-stories` | 1 story folder and 2 nested stories inside that folder (4 components, 3 stories). |
 | `has-private-assets` | 1 public asset and 1 private asset (4 components, 2 assets). |
 | `has-nested-asset-folders` | 2 nested asset folders (A → B) and 3 assets: 1 at root, 1 in Folder A, 1 in Folder B (4 components, 2 folders, 3 assets). |
+| `has-diverse-components` | 5 components (1 root, 4 nestable) using text, textarea, richtext, number, datetime, boolean, option, options, asset, multilink, bloks field types, plus 1 datasource. Good baseline for `schema pull` and round-trip testing. |
```
alexjoverm:
```suggestion
| `has-diverse-components` | 5 components (1 root, 4 nestable) using text, textarea, richtext, number, datetime, boolean, option, options, asset, multilink, bloks field types, plus 1 datasource. Good baseline for `schema init` and round-trip testing. |
```

packages/cli/test/GUIDE.md
```diff
@@ -27,6 +27,59 @@ bash .claude/skills/qa-engineer-manual/scripts/seed-scenario.sh \
...
+
+The `schema` command uses TypeScript entry files (not JSON). When test files live outside a workspace package (e.g. `.claude/tmp/`), they can't resolve `@storyblok/schema` — rewrite the imports to absolute source paths before pushing.
+
+### Test patterns
+
+**Round-trip (pull → push idempotency):**
+```bash
+# 1. Seed space with has-default-components
+# 2. Pull
+node ./dist/index.mjs schema pull --space "$STORYBLOK_SPACE_ID" --out-dir ../../.claude/tmp/schema-test
```
alexjoverm:
```suggestion
**Round-trip (init → push idempotency):**
```bash
# 1. Seed space with has-default-components
# 2. Init
node ./dist/index.mjs schema init --space "$STORYBLOK_SPACE_ID" --out-dir ../../.claude/tmp/schema-test
```

packages/cli/src/commands/schema/rollback/actions.ts
```diff
@@ -0,0 +1,404 @@
+import { readFile } from 'node:fs/promises';
+import { join } from 'pathe';
+import chalk from 'chalk';
+import { createTwoFilesPatch } from 'diff';
+
+import type { ChangesetData, ChangesetEntry, RemoteSchemaData } from '../types';
+import { applyDefaults, COMPONENT_DEFAULTS } from '../utils';
+import { serializeComponent, serializeComponentFolder, serializeDatasource } from '../push/serialize';
```
alexjoverm: suggestion(non-blocking): better to extract them to a common place rather than cross-referencing

packages/cli/src/commands/schema/rollback/actions.ts
```diff
@@ -0,0 +1,404 @@
...
+import { getMapiClient } from '../../../api';
+import { handleAPIError } from '../../../utils';
+import { fileExists, readDirectory } from '../../../utils/filesystem';
+import {
+  toComponentCreate,
+  toComponentFolderCreate,
+  toComponentUpdate,
+  toDatasourceCreate,
+  toDatasourceUpdate,
+} from '../push/actions';
```
alexjoverm: suggestion(non-blocking): better to extract them to a common place rather than cross-referencing

packages/cli/src/commands/schema/push/changeset.ts
```diff
@@ -0,0 +1,24 @@
+import { mkdir, writeFile } from 'node:fs/promises';
+import { join } from 'pathe';
+
+import type { ChangesetData } from '../types';
+
+/** Creates a filesystem-safe timestamp string for file names. */
+function fileTimestamp(iso: string): string {
+  return iso.replace(/[:.]/g, '-');
+}
```
alexjoverm: minor(non-blocking): this function is re-implemented in https://github.com/storyblok/monoblok/pull/579/changes#diff-1539ef9a2e25adb780136567c3e9334744be411d2e5b64f6e51e950476c4b590R120

Better to extracted into a util, with tests

packages/cli/src/commands/schema/push/actions.ts
```diff
@@ -0,0 +1,429 @@
...
+      catch (error) {
+        handleAPIError('push_component', error as Error, `Failed to create component ${diff.name}`);
+      }
+    }
+    else if (diff.action === 'update' && localComp) {
+      const existing = remote.components.get(diff.name);
+      if (existing?.id) {
+        const payload = toComponentUpdate(localComp);
+        try {
+          await client.components.update(existing.id, {
```
alexjoverm, line 171: @maoberlehner do you see here an opportunity to apply parallelization? No need for complicated streams IMO for components, just executing the promises in parallel and using `allSettled` could suffice. The mapi-client will handle rate limits so not even a think to care here

packages/cli/src/commands/schema/push/actions.ts
```diff
@@ -0,0 +1,429 @@
...
+      catch (error) {
+        handleAPIError('push_datasource', error as Error, `Failed to create datasource ${diff.name}`);
+      }
+    }
+    else if (diff.action === 'update' && localDs) {
+      const existing = remote.datasources.get(diff.name);
+      if (existing?.id) {
+        const payload = toDatasourceUpdate(localDs, existing);
+        try {
+          await client.datasources.update(existing.id, {
```
alexjoverm: same here https://github.com/storyblok/monoblok/pull/579/changes#r3220228099


For other calls (deleting non-existing components, stale datasources, etc) I see it a little less important as I would predict bulk operations there being smaller. But worth consider to apply some DX nicety/helper to handle in parallel promises to all calls here

packages/schema/playground/qwik/.gitignore
alexjoverm, line 1: @maoberlehner why the `qwik` playground shows in the diff? I guess it's already existing in `alpha`?

packages/schema/src/zod/index.ts
alexjoverm, line 1: Being realistic, I think all Zod-related part can be a project of its own which will take considerable time to iterate, review and document. Might make more sense to do it afterwards as an extra feature to reduce the scope. Let's discuss separatedly

DOCS_SCHEMA.md
```diff
@@ -0,0 +1,245 @@
+# `@storyblok/schema` Reference
```
alexjoverm, line 1: reminder - Just a comment to be resolved when the PR is approved to removed this file

DOCS_SCHEMA_CLI.md
```diff
@@ -0,0 +1,146 @@
+# `storyblok schema` Reference
```
alexjoverm, line 1: reminder - Just a comment to be resolved when the PR is approved to removed this file

alexjoverm (commented):
Looking great - added a few comments but are mostly minors.

Not approving yet as still need to go through deeper manual testing.

packages/schema/playground/qwik/.gitignore
maoberlehner: Yes, it was already in the alpha branch. Currently, only I work on alpha. Should we rewrite history?

alexjoverm: Not too sure. If it's about renaming some non-release commits, I guess should be fine. But not release commits as they're associated with release tags.

 But IMO, no need to rewrite

packages/schema/src/zod/index.ts
maoberlehner: Outsourced in feat/WDX-419-zod-schemas

alexjoverm (approved):
Congrats @maoberlehner, it's working great! Dropping below a couple of minors, feel free to move it forward once tackled.

- When you run `schema push` and a migration is created as a result of a breaking change, nothing regarding that is added to the logs. In other words: if you try to reverse engineer what happened by looking at the logs, you have no clue that migration files were created (example of log below).
- When there are breaking changes, it's not too easy to spot them based on the CLI output. For example, in the screenshot below, it's not too clear for me which breaking changes those are (which component? was a field renamed? did some property change?). Not sure if we should care too much, but if there would a more clear way to spot them would be great

Image


```
{"timestamp":"2026-05-13T06:44:48.966Z","level":"INFO","message":"Schema push started","context":{"runId":1778654688964,"command":"storyblok schema push","options":{"dryRun":false,"delete":false,"migrations":true,"writeComponents":true,"space":"292171738121885","verbose":false,"apiMaxRetries":3,"apiMaxConcurrency":6,"logConsoleEnabled":false,"logConsoleLevel":"info","logFileEnabled":true,"logFileLevel":"info","logFileMaxFiles":10,"uiEnabled":true,"reportEnabled":true,"reportMaxFiles":10},"cliVersion":"4.17.2","entryFile":"composability/schema.ts","space":"292171738121885"}}
{"timestamp":"2026-05-13T06:44:58.054Z","level":"INFO","message":"Changeset saved","context":{"runId":1778654688964,"command":"storyblok schema push","options":{"dryRun":false,"delete":false,"migrations":true,"writeComponents":true,"space":"292171738121885","verbose":false,"apiMaxRetries":3,"apiMaxConcurrency":6,"logConsoleEnabled":false,"logConsoleLevel":"info","logFileEnabled":true,"logFileLevel":"info","logFileMaxFiles":10,"uiEnabled":true,"reportEnabled":true,"reportMaxFiles":10},"cliVersion":"4.17.2","path":"FAKE_PATH"}}
{"timestamp":"2026-05-13T06:44:58.888Z","level":"INFO","message":"Wrote local component files","context":{"runId":1778654688964,"command":"storyblok schema push","options":{"dryRun":false,"delete":false,"migrations":true,"writeComponents":true,"space":"292171738121885","verbose":false,"apiMaxRetries":3,"apiMaxConcurrency":6,"logConsoleEnabled":false,"logConsoleLevel":"info","logFileEnabled":true,"logFileLevel":"info","logFileMaxFiles":10,"uiEnabled":true,"reportEnabled":true,"reportMaxFiles":10},"cliVersion":"4.17.2","space":"292171738121885","componentsWritten":11,"groupsWritten":3}}
{"timestamp":"2026-05-13T06:44:58.888Z","level":"INFO","message":"Schema push finished","context":{"runId":1778654688964,"command":"storyblok schema push","options":{"dryRun":false,"delete":false,"migrations":true,"writeComponents":true,"space":"292171738121885","verbose":false,"apiMaxRetries":3,"apiMaxConcurrency":6,"logConsoleEnabled":false,"logConsoleLevel":"info","logFileEnabled":true,"logFileLevel":"info","logFileMaxFiles":10,"uiEnabled":true,"reportEnabled":true,"reportMaxFiles":10},"cliVersion":"4.17.2","summary":{"total":13,"succeeded":13,"failed":0}}}

```

PR #584, opened 2026-04-27
fix: preserve local template names and add static fallback

dipankarmaikap:
This PR improves template resolution in the create command by ensuring locally defined template metadata takes priority over dynamically fetched repository data.

It also adds a fallback to the static template list when fetching the dynamic template list fails, improving reliability and preventing broken template resolution.

## Changes

- Fix Next.js naming in the create command
- Preserve locally defined template names and metadata over dynamically fetched repository details
- Add fallback to static template definitions if dynamic template fetching fails
- Ensure templates like `Next.js` keep their intended naming instead of being replaced by repo names

### DIFF
--- a/packages/cli/src/commands/create/actions.test.ts
+++ b/packages/cli/src/commands/create/actions.test.ts
@@ -451,7 +451,19 @@ describe('repositoryToTemplate', () => {
     });
   });
 
-  it('should fallback to default port if no port topic', () => {
+  it('should fallback to default port if no port topic and no static template match', () => {
+    const repo = {
+      name: 'blueprint-core-unknown',
+      topics: [],
+      clone_url: 'https://github.com/storyblok/blueprint-core-unknown.git',
+      description: 'An unknown starter',
+      updated_at: '2024-01-01T00:00:00Z',
+    };
+    const template = repositoryToTemplate(repo);
+    expect(template.location).toBe('https://localhost:3000/');
+  });
+
+  it('should use static template location when available', () => {
     const repo = {
       name: 'blueprint-core-react',
       topics: [],
@@ -460,7 +472,8 @@ describe('repositoryToTemplate', () => {
       updated_at: '2024-01-01T00:00:00Z',
     };
     const template = repositoryToTemplate(repo);
-    expect(template.location).toBe('https://localhost:3000/');
+    // React static template has location https://localhost:5173/
+    expect(template.location).toBe('https://localhost:5173/');
   });
 });
 
@@ -521,12 +534,12 @@ describe('fetchBlueprintRepositories', () => {
     expect(blueprints?.[0]?.name).toBe('React');
     expect(blueprints?.[1]?.name).toBe('Vue');
 
-    // Verify blueprint structure
+    // Verify blueprint structure (React uses static template location)
     expect(blueprints?.[0]).toEqual({
       name: 'React',
       value: 'react',
       template: 'https://github.com/storyblok/blueprint-core-react.git',
-      location: 'https://localhost:3000/',
+      location: 'https://localhost:5173/', // From static templates
       description: 'A React starter',
       updated_at: '2024-01-02T00:00:00Z',
       stars: 75,
@@ -548,5 +561,23 @@ describe('fetchBlueprintRepositories', () => {
       octokitError,
       'Failed to fetch blueprints from GitHub',
     );
+  });
+
+  it('should return static templates as fallback when GitHub API fails', async () => {
+    // Mock createOctokit to throw an error
+    const octokitError = new Error('GitHub API error');
+    mockedCreateOctokit.mockImplementation(() => {
+      throw octokitError;
+    });
+
+    const blueprints = await fetchBlueprintRepositories();
+
+    // Should return static templates as fallback
+    expect(blueprints).toBeDefined();
+    expect(blueprints.length).toBeGreaterThan(0);
+    // Check that known static templates are present
+    expect(blueprints.some(bp => bp.value === 'react')).toBe(true);
+    expect(blueprints.some(bp => bp.value === 'vue')).toBe(true);
+    expect(blueprints.some(bp => bp.value === 'nuxt')).toBe(true);
   });
 });
--- a/packages/cli/src/commands/create/actions.test.ts
+++ b/packages/cli/src/commands/create/actions.test.ts
@@ -6,7 +6,6 @@ import open from 'open';
 import { createEnvFile, extractPortFromTopics, fetchBlueprintRepositories, generateProject, generateSpaceUrl, handleEnvFileCreation, openSpaceInBrowser, repositoryToTemplate } from './actions';
 import * as filesystem from '../../utils/filesystem';
 
-// Mock external dependencies
 vi.mock('node:child_process');
 vi.mock('node:fs/promises', () => ({
   default: {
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/create/actions.test.ts
```diff
@@ -451,7 +451,19 @@ describe('repositoryToTemplate', () => {
...
+  it('should fallback to default port if no port topic and no static template match', () => {
+    const repo = {
+      name: 'blueprint-core-unknown',
+      topics: [],
+      clone_url: 'https://github.com/storyblok/blueprint-core-unknown.git',
+      description: 'An unknown starter',
+      updated_at: '2024-01-01T00:00:00Z',
+    };
+    const template = repositoryToTemplate(repo);
+    expect(template.location).toBe('https://localhost:3000/');
```
alexjoverm, line 439: (non-blocking): I'm unsure if this test is really valuable. If the template doesn't exist, either statically or remotely in Github, I think it should fail and err rather than providing a default port and localhost url

packages/cli/src/commands/create/actions.ts
```diff
@@ -195,27 +199,53 @@ export const extractPortFromTopics = (topics: string[]): string => {
...
+
 /**
  * Converts a GitHub repository to the format expected by the CLI
- * @param repo - GitHub repository data
+ * Uses static template name and location as priority if available
+ * @param repo - GitHub repository data from search API
  * @returns Formatted blueprint object
  */
-export const repositoryToTemplate = (repo: any): DynamicTemplate => {
+export const repositoryToTemplate = (repo: SearchRepoItem): DynamicTemplate => {
```
alexjoverm: praise: nice improvement on the type here!

packages/cli/src/commands/create/actions.ts
```diff
@@ -236,7 +266,9 @@ export const fetchBlueprintRepositories = async () => {
     return blueprints;
   }
   catch (error) {
-    // Fallback to hardcoded blueprints if GitHub search fails
+    // Fallback to static templates if GitHub search fails
     handleAPIError('fetch_blueprints', error as Error, 'Failed to fetch blueprints from GitHub');
+    konsola.warn('Using offline template list');
+    return getStaticTemplatesAsFallback();
```
alexjoverm: (blocking): these 2 lines of code must be unreachable, as the previous `handleAPIError` always throws. You don't need the `handleAPIError` anymore, as now you're using static templates as a fallback. You can safely remove it.

suggestion (non-blocking): Also, it's a great chance to replace the old `konsola` by the newer UI module, and even introduce the Logger (see [this for reference](https://github.com/storyblok/monoblok/blob/main/packages/cli/src/commands/migrations/run/actions.ts#L68-L71)):

```js
catch (error) {
      getUI().warn('Failed to fetch blueprints from GitHub. Using offline template list.');
      getLogger()...
      return getStaticTemplatesAsFallback();
  }
```

alexjoverm (approved):
Great work overall @dipankarmaikap. Just a small change and it's all done

packages/cli/src/commands/create/actions.test.ts
maoberlehner: I think the `repositoryToTemplate` does not know if the template exists or not, and IMO it shouldn't check it (check should happen before calling it). The function either has to choose some default or not accept a repo without topics. So I think choosing `3000` as a fallback is reasonable.

packages/cli/src/commands/create/actions.test.ts
```diff
@@ -460,7 +472,8 @@ describe('repositoryToTemplate', () => {
       updated_at: '2024-01-01T00:00:00Z',
     };
     const template = repositoryToTemplate(repo);
-    expect(template.location).toBe('https://localhost:3000/');
+    // React static template has location https://localhost:5173/
+    expect(template.location).toBe('https://localhost:5173/');
```
maoberlehner, line 451: nitpick: The comment does not add much value, as it is the logical conclusion.

packages/cli/src/commands/create/actions.test.ts
```diff
@@ -549,4 +562,22 @@ describe('fetchBlueprintRepositories', () => {
       'Failed to fetch blueprints from GitHub',
     );
   });
+
+  it('should return static templates as fallback when GitHub API fails', async () => {
+    // Mock createOctokit to throw an error
```
maoberlehner: nitpick: also more noise than signal, I can see this in the lines below.

packages/cli/src/commands/create/actions.ts
```diff
@@ -195,27 +199,53 @@ export const extractPortFromTopics = (topics: string[]): string => {
...
     stars: repo.stargazers_count,
   };
 };
 
-export const fetchBlueprintRepositories = async () => {
+/**
+ * Converts static templates to DynamicTemplate format for fallback
+ * @returns Array of DynamicTemplate objects
+ */
+const getStaticTemplatesAsFallback = (): DynamicTemplate[] => {
```
maoberlehner: (non-blocking, pre-existing): What makes `DynamicTemplate`s dynamic?

Why not:

```ts
type TemplateTechnology = 'REACT' | ...;

export interface Template {
  name: string;
  value: string;
  template: string;
  location: string;
  description?: string | null;
  updated_at?: string;
  stars?: number;
}

export const templates: Record<TemplateTechnology, Template> = {
  REACT: {
    name: 'React',
    value: 'react',
    template: 'https://github.com/storyblok/blueprint-core-react',
    location: 'https://localhost:5173/',
  },
  // ...
} as const;
```

And below:

```ts
konsola.warn('Using offline template list');
return Object.values(templates);
```

If you feel motivated, I think this could be improved/simplified.

packages/cli/src/commands/create/actions.ts
```diff
@@ -195,27 +199,53 @@ export const extractPortFromTopics = (topics: string[]): string => {
...
- * @param repo - GitHub repository data
+ * Uses static template name and location as priority if available
+ * @param repo - GitHub repository data from search API
  * @returns Formatted blueprint object
  */
-export const repositoryToTemplate = (repo: any): DynamicTemplate => {
+export const repositoryToTemplate = (repo: SearchRepoItem): DynamicTemplate => {
   const technology = repo.name.replace('blueprint-core-', '');
   const port = extractPortFromTopics(repo.topics || []);
+  const staticTemplate = findStaticTemplate(technology);
```
maoberlehner: Is there a particular reason why not `templates[technology]`?

dipankarmaikap: Thank you for pointing that out. We only needed to update the keys in the template list, specifically for Next.js. This has now been fixed by switching to `templates[technology]`.

packages/cli/src/commands/create/actions.ts
dipankarmaikap: Change it to be called `Template`

packages/cli/src/commands/create/actions.test.ts
dipankarmaikap: Removed all the inline comments.

packages/cli/src/commands/create/actions.ts
dipankarmaikap: Thank you for pointing that out, @alexjoverm. Switched the create command to use `getUI` and removed `konsola`.

alexjoverm (approved):
LGTM, great one @dipankarmaikap !

PR #588, opened 2026-04-28
fix: sb component input handling

dipankarmaikap:
This PR improves `<sb-component>` input handling by making it more flexible and resilient.

It now supports `SbBlokData`, `SbBlokData[]`, `null`, and `undefined`, and normalizes all values internally into a consistent array format.

## Changes

* Support single and multiple blok inputs
* Handle `null` and `undefined` safely
* Normalize input internally before rendering
* Reduce need for manual guards in templates

## Why

Improves DX and makes the component safer and easier to use.

### DIFF
--- a/packages/angular/src/lib/blok/sb-component.component.ts
+++ b/packages/angular/src/lib/blok/sb-component.component.ts
@@ -1,17 +1,27 @@
-import { Component, input } from '@angular/core';
+import { Component, computed, input } from '@angular/core';
 import { SbBlokDirective } from './sb-blok.directive';
 import { SbBlokData } from '../types';
 
 @Component({
   selector: 'sb-component',
   standalone: true,
+  imports: [SbBlokDirective],
   template: `
-    @for (blok of sbBloks(); track blok._uid) {
+    @for (blok of bloks(); track blok._uid) {
       <ng-container [sbBlok]="blok" />
     }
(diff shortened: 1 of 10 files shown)


packages/angular/src/lib/blok/sb-component.component.ts
```diff
@@ -1,17 +1,27 @@
...
-    @for (blok of sbBloks(); track blok._uid) {
+    @for (blok of bloks(); track blok._uid) {
       <ng-container [sbBlok]="blok" />
     }
   `,
-  imports: [SbBlokDirective],
 })
 export class StoryblokComponent {
-  readonly sbBloks = input.required<SbBlokData[]>();
+  readonly sbBlok = input<SbBlokData | SbBlokData[] | null | undefined>();
```
alexjoverm, line 16: @dipankarmaikap shall we can add a test? Something simple, just input/output based to ensure it works with all formats (object, array, null, etc)

alexjoverm (approved):
LGTM with one comment

packages/angular/src/lib/blok/sb-component.component.ts
alexjoverm: and btw, questioning here if we want to name it `block` (see team internal message)

PR #589, opened 2026-04-28
feat(nuxt): Add reactivity to useAsyncStoryblok composable https://git…

fedetorre:
Closes https://github.com/storyblok/monoblok/discussions/586

**What changed**
The api option in UseAsyncStoryblokOptions now accepts MaybeRefOrGetter<ISbStoriesParams> instead of a plain ISbStoriesParams object. This allows passing reactive values (ref, computed, or getter functions) as API parameters, so the composable can read the latest params at fetch time.

**How it works**
toValue(api) is used wherever the params are consumed — in the fetch call, the uniqueKey function, and when deriving bridgeOptions — so all three read the current value of the reactive source.
uniqueKey is now a function (instead of a pre-computed string) so useAsyncData derives a fresh key from the current params on each execution.
bridgeOptions still reads toValue(api) for resolveRelations / resolveLinks, but it is evaluated once when the composable runs (it is not reactive); the bridge is registered once per story id.

Note on refetching: useAsyncData does not watch the key getter, so changing a reactive api value alone does not auto-refetch. Trigger a re-fetch by calling refresh() or by passing a watch option (e.g. watch: [version]). This is documented in the composable's JSDoc with a reactive usage example.

The bridge watcher was simplified: a single watch with { immediate: true } replaces the previous split between an eager check and a lazy watcher, and a registeredId guard prevents re-registering the bridge for the same story. The registration block was further flattened with guard clauses.

**Breaking change**
UseAsyncStoryblokOptions.api now has type MaybeRefOrGetter<ISbStoriesParams>. Passing a plain object continues to work as before (it is a valid MaybeRefOrGetter), so existing usage is unaffected.

### DIFF
--- a/packages/nuxt/src/runtime/composables/useAsyncStoryblok.ts
+++ b/packages/nuxt/src/runtime/composables/useAsyncStoryblok.ts
@@ -1,7 +1,7 @@
+import type { AsyncData, AsyncDataOptions, NuxtError } from '#app';
+import { useAsyncData } from '#app';
 import { type ISbResult, type ISbStoriesParams, type StoryblokBridgeConfigV2, useStoryblokApi, useStoryblokBridge } from '@storyblok/vue';
 import { computed, type ComputedRef, type Ref, watch } from 'vue';
-import { useAsyncData } from '#app';
-import type { AsyncData, AsyncDataOptions, NuxtError } from '#app';
 
 /**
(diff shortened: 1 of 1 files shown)


alexjoverm (approved):

alexjoverm:
Thanks for the contribution @fedetorre & @maoberlehner !

PR #593, opened 2026-04-30
chore: adopt Vite+ toolchain and unify package scripts

maoberlehner:
- Migrate the monorepo to the **Vite+ (`vp`) toolchain**: `vp pack`/`vp build` for bundling, `vp test` for testing, `vp lint` (oxlint) for linting, `vp fmt` (oxfmt) for formatting. Framework-specific builders (Angular CLI, Nuxt module builder, `svelte-package`) are kept where Vite+ doesn't apply.
- Replace `@storyblok/eslint-config` with `@storyblok/lint-config` (oxlint-based) across all packages. The old `eslint-config` package is **removed**, not deprecated: leaving two lint configs in the tree is what caused the drift in the first place.
- **Formatting becomes an enforced step, where before there was none.** No project had a `format` script and there was no repo-level formatter config; what existed was residue (`prettier` as a devDependency in 4 packages, a `.prettierrc` in `region-helper`, a `prettier` block in `packages/angular/package.json`, `eslint-config-prettier` in 3 packages) that nothing ran. This is why the formatting commit is large.
- **Unify package scripts** so `nx run-many` targets work uniformly. Canonical vocabulary: `build`, `dev` (watch-build), `lint` / `lint:fix`, `test`, `test:watch`, `test:ui`, `test:unit`, `test:e2e`, `test:e2e:watch`, `test:types`, `test:coverage`. Drops drifted variants like `build:watch`, `coverage`, `test:unit:ui`, `test:e2e-watch`.
- **Formatting is owned by the repo root, not by packages.** `vp fmt` reads the `fmt` key of the Vite+ config in its own working directory and does not search parent directories, so a per-package `vp fmt` would silently fall back to oxfmt's defaults (and four projects have no Vite+ config at all). The root `vite.config.ts` is the only formatter config, `pnpm format` formats the whole repo, and CI checks it once. Packages have no `format` script.
- Update `nx.json` `targetDefaults` for the new targets (caching + `dependsOn`).
- Update `README.md`, `CONTRIBUTING.md`, `AGENTS.md`, and per-package READMEs.
- Adds ADR [`0013-adopt-viteplus.md`](adr/0013-adopt-viteplus.md).

## Commits

Two commits, deliberately split so the toolchain change is reviewable on its own:

1. **`chore: adopt Vite+ toolchain and unify package scripts`** - 164 files. Every meaningful change. No file in this commit changes only formatting.
2. **`chore: reformat repo with oxfmt`** - 1108 files. Formatting only: whitespace, quote style, line wrapping, `package.json` key order.

Review the first commit; the second is mechanical output of `pnpm format`.

## Performance

Lint runtime, `nx run-many --target=lint --exclude="@storyblok/playground-*" --skip-nx-cache`, 3 runs each, same machine, both trees prebuilt:

| Run      | Branch (vp / oxlint) | Main (eslint) |
| -------- | -------------------- | ------------- |
| 1        | 11s                  | 18s           |
| 2        | 9s                   | 16s           |
| 3        | 10s                  | 16s           |
| avg      | **~10s**             | ~16.7s        |
| projects | 20                   | 20            |

Roughly **40% faster** on the same 20 projects, and tighter run-to-run (9-11s vs 16-18s).

### Caveats

- **Not an apples-to-apples comparison.** Oxlint covers a subset of the rules ESLint did. Some `@antfu/eslint-config` and `@storyblok/eslint-config` rules have no oxlint equivalent yet; `packages/lint-config/src/base.ts` lists them as comments so the intent stays visible. Part of the gain is less work, not just a faster engine.
- Playgrounds are excluded from both runs (they invoke `next lint` interactively).
- Local benchmark on a single machine; CI numbers will differ.
- `--skip-nx-cache` measures actual lint work. Cache hits in normal dev/CI are far faster than either column.

## Things this surfaced

Adopting the toolchain exposed problems that had to be fixed for the branch to be green:

- **nx: `test`, `test:types`, `test:coverage` now depend on `^build`.** With `injectWorkspacePackages`, pnpm copies a dependency into its consumers and only refreshes that copy after the dependency's `build` runs. Without `^build` these targets raced that sync and failed intermittently in the integration-test playgrounds.
- **Oxlint's type-aware rules caught a real defect:** an `import` inside a `declare namespace` in `@storyblok/astro` (TS1147).
- **Vitest 3 -> 4** (what Vite+ pins). Vitest 4 invokes mock implementations with `new`, so a `@storyblok/js` bridge mock had to stop using an arrow function.
- **Generated and vendored files are excluded from formatting.** `**/src/generated/` and `tools/openapi-codegen/specs/` (committed, pinned by `spec.lock`) are in `.prettierignore`, matching what every `oxlint.config.ts` already ignores, so regenerating or re-pulling never fights the formatter. Same for committed tool output (coverage reports, `.svelte-kit`).
- **`proseWrap: "always"` reverses the repo's old "never hard-wrap markdown" rule**, so `.claude/rules/writing-style.md` is updated: prose in the repo is wrapped by the formatter; prose outside it (PR descriptions, issue comments) stays unwrapped, since those renderers are linebreak-sensitive.
- **oxfmt/oxlint are pinned to the exact versions `vite-plus` pins** (`oxfmt` 0.61.0, `oxlint` 1.76.0). Declaring newer ones installs a second copy that `vp` never runs.

Fixes WDX-327


dipankarmaikap (approved):
Overall, looks good to me. Just two small things I noticed:

* The root `oxlint.config.ts` lints the playground directories, but the individual package configs (for example, Vue's `oxlint.config.ts`) ignore the playground. Should we make this consistent?
* Not a requirement, but could we also fix the `publint` warning about `repository.url` needing to use the `git+https://...` format in this PR?

Neither is blocking from my side; just thought it would be nice to clean these up while we're here.

maoberlehner:
> Overall, looks good to me. Just two small things I noticed:
> 
> * The root `oxlint.config.ts` lints the playground directories, but the individual package configs (for example, Vue's `oxlint.config.ts`) ignore the playground. Should we make this consistent?
> * Not a requirement, but could we also fix the `publint` warning about `repository.url` needing to use the `git+https://...` format in this PR?
> 
> Neither is blocking from my side; just thought it would be nice to clean these up while we're here.

- Just noticed problems with the root linting behavior myself. `packages/` and `tools/*/` that have their own lint config are now excluded from root lint.
- Fixed `publint` warnings

PR #594, opened 2026-05-04
fix: enhance richTextRenderer with customizable renderers and options

dipankarmaikap:
This PR refines the `richTextRenderer` introduced earlier for static rendering.

Previously, the renderer only generated HTML with no way for users to customise the output. This update introduces full customisation support via `SbRichTextOptions`.

### **What’s new**

* **Custom renderers support**

  * Added a `renderers` option that allows overriding how nodes and marks are rendered
  * Fully typed for all supported nodes and marks

* **Feature parity improvements**

  * Supports image optimisation via `optimizeImages`, similar to `richTextResolver`
  * Improved handling of marks and node rendering
  * Exported small internal utilities for:

      * table handling
      * mark handling
        (these can be reused by other framework SDKs)
### **Usage examples**

```ts
const options: SbRichTextOptions = {
  renderers: {
    paragraph: ({ content }) =>
      `<div class="custom">${richTextRenderer(content, options)}</div>`,
  },
};

const node: SbRichTextDoc = {
  type: 'paragraph',
  content: [text('Hello')],
};

expect(richTextRenderer(node, options))
  .toBe('<div class="custom">Hello</div>');
```

```ts
const options: SbRichTextOptions = {
  renderers: {
    bold: ({ children }) => `<b class="heavy">${children}</b>`,
  },
};

expect(
  richTextRenderer(text('Bold', [{ type: 'bold' }]), options)
).toBe('<b class="heavy">Bold</b>');
```

### **Tests**

* Added comprehensive test coverage for:

  * mark merging
  * table rendering
  * image optimisation
  * renderer overrides

### **Angular SDK alignment**

* Updated Angular SDK rich text implementation to match the same output
* Reused the same test cases to ensure consistency

### **Notes / Next steps**

* This PR focuses on stabilising and validating the API surface
* A follow-up PR will:

  * make this renderer the main export
  * introduce framework-specific components for React, Vue, and Astro with renderer customisation support


packages/richtext/src/static/attribute.ts
```diff
@@ -31,35 +31,55 @@ const STYLE_MAP: StyleMap = {
...
   const rest: Record<string, unknown> = {};
+  let staticAttrs: Record<string, unknown> = {};
 
+  if (type in NODE_RENDER_MAP) {
+    const renderMap
+    = NODE_RENDER_MAP[
+      type as keyof typeof NODE_RENDER_MAP
+    ];
+    if (renderMap && 'attrs' in renderMap) {
+      const attrs = renderMap.attrs || {};
```
alexjoverm, line 66: This param is shadowed by the function `attrs` parameter. Better find a different name to eliminate ambiguity

packages/richtext/src/static/attribute.ts
```diff
@@ -31,35 +31,55 @@ const STYLE_MAP: StyleMap = {
...
+
+      const rawStyle
+      = 'style' in attrs && typeof attrs.style === 'string'
+        ? attrs.style
+        : '';
+      const { style: _style, ...rest } = attrs as Record<string, unknown>;
```
alexjoverm, line 72: same here, this `rest` variable is shadowed by the `const rest` in line 57. Find different names or refactor

packages/richtext/src/static/attribute.ts
```diff
@@ -94,10 +114,16 @@ export function processAttrs(
     // DEFAULT PASS THROUGH
     rest[attrName] = value;
   }
-  // Special handling for Storyblok links to add anchor to href
+  // Special handling for Storyblok links
   if (type === 'link' && attrs.linktype === 'story') {
-    rest.href = `${attrs.href ?? ''}#${attrs.anchor ?? ''}`;
+    rest.href = `${attrs.href ?? ''}${attrs.anchor ? `#${attrs.anchor}` : ''}`;
```
alexjoverm, line 119: nitpick: I find this line unnecessarily hard to read and grasp. I'd suggest to make it more readable, even if it ends in 2-3 lines instead of one

packages/richtext/src/static/node-helpers.ts
```diff
@@ -0,0 +1,154 @@
...
+ * @param a - First link mark
+ * @param b - Second link mark
+ * @returns True if the marks have identical attributes
+ */
+export function areLinkMarksEqual(a: PMMark | null, b: PMMark | null): boolean {
+  if (!a || !b) {
+    return false;
+  }
+
+  const aa = (a.attrs ?? {}) as Record<string, unknown>;
```
alexjoverm, line 38: Can we find better names rather than `aa` or `ab`?

packages/richtext/src/static/node-helpers.ts
```diff
@@ -0,0 +1,154 @@
...
+  const ba = (b.attrs ?? {}) as Record<string, unknown>;
+
+  return (
```
alexjoverm, line 41: blocking: a link type can be set to accept custom attrs (see image below). We need to add those custom attrs to the comparison for this to work correct, using deep equality (we have a `deepEqual` in `richtext/src/utils/index.ts`)


Image

packages/richtext/src/static/attribute.ts
```diff
@@ -31,35 +31,55 @@ const STYLE_MAP: StyleMap = {
...
 const DEFAULT_ATTR_MAP: AttrMap = {
   fallbackImage: 'src',
-  meta_data: 'data-meta_data',
   body: 'data-body',
   colspan: 'colSpan',
   rowspan: 'rowSpan',
 };
 
-export const EXCLUDED_ATTRS = new Set(['level', 'linktype', 'uuid', 'custom', 'anchor']);
+export const EXCLUDED_ATTRS = new Set(['level', 'linktype', 'uuid', 'custom', 'anchor', 'meta_data', 'copyright', 'source']);
```
alexjoverm, line 39: blocking (regression): similarly to [this comment](https://github.com/storyblok/monoblok/pull/594/changes#r3194360173), I realize there might be a regression here. As actually the current richtext spreads the `custom` attributes in the `<a>` tag, and by being excluded that will be dropped

packages/richtext/src/static/richtext-renderer.ts
```diff
@@ -1,166 +1,326 @@
...
+  return `<${tag}${htmlAttrs}>${content}</${tag}>`;
+}
+
+/** Renders an image node with optimization applied. */
+function renderOptimizedImage(node: SbRichTextDoc, options: SbRichTextOptions): string {
+  const attrs = node.attrs as Record<string, unknown> | undefined;
+  const src = attrs?.src as string | undefined;
+
+  if (!src) {
+    return buildHtmlAttrs('image', attrs) ? `<img${buildHtmlAttrs('image', attrs)}>` : '<img>';
```
alexjoverm, line 105: `buildHtmlAttrs` is unnecessarily called twice here, which could have potential future side effects. Better to refactor it:

```ts
if (!src) {
  const htmlAttrs = buildHtmlAttrs('image', attrs);
  return `<img${htmlAttrs}>`;
}
```

packages/richtext/src/static/attribute.ts
```diff
@@ -94,10 +114,16 @@ export function processAttrs(
...
-  // Special handling for Storyblok links to add anchor to href
+  // Special handling for Storyblok links
   if (type === 'link' && attrs.linktype === 'story') {
-    rest.href = `${attrs.href ?? ''}#${attrs.anchor ?? ''}`;
+    rest.href = `${attrs.href ?? ''}${attrs.anchor ? `#${attrs.anchor}` : ''}`;
+  }
+  if (type === 'link'
+    && attrs.linktype === 'email'
+    && typeof attrs.href === 'string') {
+    rest.href = `mailto:${attrs.href.replace('mailto:', '') ?? ''}`;
```
alexjoverm, line 124: also there is unnecessary nullish coalescing here. Can be `mailto:${attrs.href.replace('mailto:', '')}`

alexjoverm (changes requested):
@dipankarmaikap Added a few comments to polish further

An important note -> keep in mind this PR is a `feat`, not a `fix`. It's adding new functionality, new exports, new public api (fix, by definition, is when you fix a behavior without changing the public api at all). Could you rebase the PR accordingly?

As a side note - notice some publicly exported types were changed. "By the book", that will mean a breaking change. That said, usage of the SDK is still low and it's just types, not affecting runtime, so I think we can safely move forward without a major

dipankarmaikap:
@alexjoverm Closing this PR in favor of [https://github.com/storyblok/monoblok/pull/595](https://github.com/storyblok/monoblok/pull/595) since the commit history here became a bit messy and the commit format should follow `feat`. The new PR provides a cleaner history. I’ve also incorporated all the feedback you shared on this PR into the new one.

PR #595, opened 2026-05-06
feat: enhance richTextRenderer with customisable renderers and options

dipankarmaikap:
This PR refines the `richTextRenderer` introduced earlier for static rendering.

Previously, the renderer only generated HTML with no way for users to customise the output. This update introduces full customisation support via `SbRichTextOptions`.

### **Custom renderers support**

* Added a `renderers` option to override how nodes and marks are rendered
* Fully typed for all supported nodes and marks

### **Feature parity improvements**

* Supports image optimisation via `optimizeImages`, similar to `richTextResolver`
* Improved handling of:

  * mark merging
  * table rendering (`thead` / `tbody`)
  * nodes with null tags but valid content
* Exported small internal utilities for:

  * table handling
  * mark handling
    (can be reused by other framework SDKs)

### **API and type improvements**

* Introduced `SbRichTextDoc` as the primary RichText document type
* Renamed:

  * `StoryblokRichTextJson` → `SbRichTextDoc`
  * `SBAngularComponentMap` → `SbAngularComponentMap`
* Improved custom richtext component typing across Angular SDK
* These renames are considered safe as the package is not yet v1

## **Usage examples**

```ts
const options: SbRichTextOptions = {
  renderers: {
    paragraph: ({ content }) =>
      `<div class="custom">${richTextRenderer(content, options)}</div>`,
  },
};

const node: SbRichTextDoc = {
  type: 'paragraph',
  content: [text('Hello')],
};

expect(richTextRenderer(node, options))
  .toBe('<div class="custom">Hello</div>');
```

```ts
const options: SbRichTextOptions = {
  renderers: {
    bold: ({ children }) => `<b class="heavy">${children}</b>`,
  },
};

expect(
  richTextRenderer(text('Bold', [{ type: 'bold' }]), options)
).toBe('<b class="heavy">Bold</b>');
```

## **Tests**

* Added comprehensive test coverage for:

  * mark merging
  * table rendering
  * image optimisation
  * renderer overrides
* Improved and cleaned up existing tests
* Ensured consistent output across implementations

## **Angular SDK alignment**

* Updated Angular rich text implementation to match static renderer output
* Reused the same test cases to ensure parity
* Improved custom component type mapping

## **Notes / Next steps**

* This PR focuses on stabilising and validating the API surface
* A follow-up PR will:

  * make this renderer the main export
  * introduce framework-specific components for React, Vue, and Astro with renderer customisation support

## **Open question**

The docs team suggested renaming `richTextRenderer` → `renderRichText`.

Since:

* this API is not yet documented publicly
* it is not exposed yet

We can safely rename it without introducing breaking changes.

👉 Feedback welcome on whether we should proceed with this rename in the next PR.

### DIFF
--- a/packages/angular/src/lib/richtext/rich-text.component.ts
+++ b/packages/angular/src/lib/richtext/rich-text.component.ts
@@ -119,10 +119,8 @@ export class SbRichTextComponent implements OnDestroy {
       if (this.shouldAbort(version)) return;
 
       if (group.linkMark) {
-        // Render grouped text nodes under a single link
         this.renderLinkGroup(group.nodes, group.linkMark, parent, version);
       } else {
-        // Render single node normally
         this.renderNode(group.nodes[0], parent, version);
       }
     }
@@ -150,10 +148,8 @@ export class SbRichTextComponent implements OnDestroy {
       return;
     }
 
-    // Create the link element
     const tag = resolveTag(linkMark);
     if (!tag) {
-      // No tag, render text nodes directly
       for (const node of nodes) {
         this.renderTextNode(node as SbRichTextDoc & { type: 'text' }, parent, version);
       }
@@ -170,7 +166,6 @@ export class SbRichTextComponent implements OnDestroy {
       const innerMarks = getInnerMarks(node);
       let current: Node = this.renderer.createText(node.text || '');
 
-      // Apply inner marks
       for (const mark of innerMarks) {
         const CustomMark = this.resolver.getSync(mark.type);
 
@@ -218,7 +213,6 @@ export class SbRichTextComponent implements OnDestroy {
     const syncComponent = this.resolver.getSync(node.type);
 
     if (syncComponent) {
-      // Custom component available synchronously
       const anchor = this.renderer.createComment('sb-node');
       this.renderer.appendChild(parent, anchor);
       this.mountComponent(syncComponent, { data: node }, parent, anchor);
@@ -227,7 +221,6 @@ export class SbRichTextComponent implements OnDestroy {
 
     // Check if this type has an async loader registered
     if (this.resolver.has(node.type)) {
-      // Create placeholder and resolve async
       const anchor = this.renderer.createComment('sb-node');
       this.renderer.appendChild(parent, anchor);
       this.resolveAndMount(node, parent, anchor, version);
@@ -270,7 +263,6 @@ export class SbRichTextComponent implements OnDestroy {
     }
 
     if (!isSelfClosing(tag)) {
-      // Special handling for tables: group rows into thead/tbody
       if (node.type === 'table' && node.content) {
         this.renderTableContent(node.content, el, version);
       } else {
@@ -295,7 +287,6 @@ export class SbRichTextComponent implements OnDestroy {
   private renderTableContent(rows: SbRichTextDoc[], tableEl: HTMLElement, version: number): void {
     const { headerRows, bodyRows } = splitTableRows(rows);
 
-    // Render thead if there are header rows
     if (headerRows.length > 0) {
       const thead = this.renderer.createElement('thead');
       for (const row of headerRows) {
@@ -304,7 +295,6 @@ export class SbRichTextComponent implements OnDestroy {
       this.renderer.appendChild(tableEl, thead);
     }
 
-    // Render tbody if there are body rows
     if (bodyRows.length > 0) {
       const tbody = this.renderer.createElement('tbody');
       for (const row of bodyRows) {
@@ -321,7 +311,6 @@ export class SbRichTextComponent implements OnDestroy {
     anchor: Node,
     version: number,
   ): Promise<void> {
-    // node.type is guaranteed to not be 'text' since text nodes are handled separately
     const CustomNode = await this.resolveCached(node.type as SbRichTextElement);
     if (this.shouldAbort(version, anchor)) return;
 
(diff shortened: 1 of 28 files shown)


packages/angular/src/lib/richtext/rich-text.component.ts
```diff
@@ -101,18 +103,110 @@ export class SbRichTextComponent implements OnDestroy {
...
+    const groups = groupLinkNodes(children);
+
+    for (const group of groups) {
+      if (this.shouldAbort(version)) return;
+
+      if (group.linkMark) {
+        // Render grouped text nodes under a single link
+        this.renderLinkGroup(group.nodes, group.linkMark, parent, version);
+      } else {
+        // Render single node normally
```
alexjoverm: nitpick: these 2 comments aren't much needed

packages/angular/src/lib/richtext/rich-text.component.ts
```diff
@@ -101,18 +103,110 @@ export class SbRichTextComponent implements OnDestroy {
...
+      for (const node of nodes) {
+        this.renderTextNode(node as SbRichTextDoc & { type: 'text' }, parent, version);
+      }
+      return;
+    }
+
+    // Create the link element
+    const tag = resolveTag(linkMark);
+    if (!tag) {
+      // No tag, render text nodes directly
```
alexjoverm: nitpick: these 2 comments aren't much needed

packages/richtext/src/static/richtext-renderer.ts
```diff
@@ -1,166 +1,327 @@
...
+ *   content: [{ type: 'paragraph', content: [{ type: 'text', text: 'Hello' }] }]
  * });
- *
- * console.log(html);
- * // <p>Hello World</p>
+ * // => '<p>Hello</p>'
  * ```
  */
-export function richTextRenderer(document: StoryblokRichTextJson): string {
+export function richTextRenderer(
```
alexjoverm: naming: `renderRichText`

packages/richtext/src/static/richtext-renderer.ts
```diff
@@ -1,166 +1,327 @@
...
+  // Static children (e.g., pre > code)
+  const staticChildren = getStaticChildren(node);
   if (staticChildren) {
-    const innerContent = renderStaticChildren(
-      staticChildren,
-      attrs,
-      childContent,
-    );
-    return `<${tag}${htmlAttrs}>${innerContent}</${tag}>`;
+    const inner = renderStaticStructure(staticChildren, node.attrs, content);
```
alexjoverm: by Claude: renderStaticStructure bypasses processAttrs (packages/richtext/src/static/richtext-renderer.ts:92-94) -- When a node has static children (e.g., code_block = pre > code), raw node.attrs are passed directly to attrsToHtmlString without going through processAttrs. The outer `<pre>` tag also gets no attributes at all. This works today because code_block's only attr is class, but it is architecturally fragile.

packages/richtext/src/static/richtext-renderer.ts
```diff
@@ -1,166 +1,327 @@
...
-  }
-  return parts.join('');
+  const nodes = document.type === 'doc' ? document.content : [document];
+  return nodes?.length ? renderChildren(nodes, options) : '';
 }
 
 /** Renders a single node to HTML. */
-function renderNode(node: StoryblokRichTextJson): string {
+function renderNode(node: SbRichTextDoc, options?: SbRichTextOptions): string {
+  // Text nodes: apply marks and escape content
```
alexjoverm: nitpick: comment doesn't add much value

packages/richtext/src/static/render-richtext.ts
```diff
@@ -1,166 +1,327 @@
...
 /** Renders a single node to HTML. */
-function renderNode(node: StoryblokRichTextJson): string {
+function renderNode(node: SbRichTextDoc, options?: SbRichTextOptions): string {
+  // Text nodes: apply marks and escape content
   if (node.type === 'text') {
-    return renderText(node);
+    return renderTextNode(node as TextNode, node.marks, options);
   }
+
+  // Custom renderer takes full control
```
alexjoverm, line 50: nitpick: comment doesn't add much value

packages/richtext/src/static/richtext-renderer.ts
```diff
@@ -1,166 +1,327 @@
...
+  // Image nodes: apply optimization if configured
+  if (node.type === 'image' && options?.optimizeImages) {
+    return renderOptimizedImage(node, options);
+  }
+
+  const htmlAttrs = buildHtmlAttrs(node.type, node.attrs);
 
-  if (selfClosing) {
-    return `<${tag}${htmlAttrs} />`;
+  // Self-closing tags
```
alexjoverm: nitpick: In general I see multiple comments like these across the code that doesn't add much value, as the code is already quite self explanatory. Some JSDocs-like explaining more complex areas are valuable, but I'd recommend to review one-liners like this and remove as they can rather add noise to the code

alexjoverm (approved):
@dipankarmaikap great work here overall, it's pretty solid. There are just few minor comments, non blocking. Feel free to move it forward once tackled, no need to re-review

packages/richtext/src/static/richtext-renderer.ts
dipankarmaikap: Thanks for pointing this out @alexjoverm . I’ve fixed it now and it correctly goes through `processAttrs`.

Regarding the attributes, currently all attributes are applied to the inner element and there is no way to control inner vs outer assignment. I think we can keep this as is for now since users can fully override the renderer and decide themselves where attributes should be applied.

I’ve also added test examples for both the Angular and HTML renderers.

PR #572, opened 2026-04-15
fix(js): enable tree-shaking by externalizing dependencies

maoberlehner:
## Summary

- Mark `@storyblok/richtext` and `storyblok-js-client` as external in the `@storyblok/js` Vite build config
- Add `"sideEffects": false` to `package.json` so bundlers can tree-shake unused exports
- Consumers who don't use richtext features drop from **~1 MB to ~32 KB** (Vite, minified)

## Context

`@storyblok/js` v5 bundles `@storyblok/richtext` (including 20+ tiptap extensions) and `storyblok-js-client` directly into its dist file, resulting in a 1.1 MB `storyblok-js.mjs`. This means even consumers who only use `storyblokInit` and `apiPlugin` pay the full cost — bundlers cannot tree-shake the unused richtext code because everything is pre-bundled into a single blob.

This partially addresses #471 (the richtext package itself is still large when used, but consumers who don't need it are no longer affected).

### Bundle size comparison (Vite, terser-minified)

| Scenario | Before | After |
|---|---|---|
| Full import (init + richtext + js-client) | 1,087 KB | 1,284 KB |
| **Partial import (init + apiPlugin only)** | **1,080 KB** | **32 KB (-97%)** |

## Test plan

- [x] `@storyblok/js` unit tests (19/19), type-check, lint — all pass
- [x] Export parity verified: all 16 exports present, types match main
- [x] Functional tests: `storyblokInit`, `apiPlugin`, `renderRichText`, `richTextResolver`, `storyblokEditable`, `useStoryblokBridge`, enums, `StoryblokClient` — all work
- [x] Downstream packages build and pass tests: vue (1/1), react (33/33), svelte (9/9), astro (67/67 unit + 4/4 e2e)
- [x] Vite bundle test confirms tree-shaking works end-to-end

Fixes WDX-310
Fixes #476

Issue #471, opened 2026-03-03
🐛 [Bug]: Bundle sizes have massively increased over past releases.


herre-ixxi:
### Package

@storyblok/richtext (Rich Text Renderer)

### Bug Description

In @storyblok/richtext 3.5.0 the total size of the package was still 37,5kb. Overall really nice for the features provided.

In 3.6.0 this increased to 168kb
In 3.7.0 this became 419kb
In 4.0.0 this became 735kb

Going by the changelogs this seems to have been caused by adding increasingly complex parsers for first Markdown and then HTML?
- https://github.com/storyblok/monoblok/pull/216 - Introduces markdown-it
- https://github.com/storyblok/monoblok/pull/246 - Introduces node-html-parser
- https://github.com/storyblok/monoblok/pull/369 - Replaces both by tiptap with a large amount of extensions.

I realize parsing these formats is complex but I do have to wonder if it's worth increasing the size of these packages by this much? To be completely honest 735kb feels genuinely like a bug for a package that's intended to be exposed client-side. Going by our bundle-analyzer this ends up making @storyblok/richtext a multitude times the size of react-dom.

Please reconsider the usage of these parsers. This doesn't feel like something that should be parsed client-side. Ideally this would be parsed within Storyblok's API itself so that only parsed data is exposed and these packages never have to deal with any of these concerns.

Alternatively consider making these parsers an optional configuration when initializing the Storyblok SDK and providing documentation on which options can and can not be used within Storyblok if these aren't provided so it's possible to skip these to keep a much more optimized bundle.

### Steps to Reproduce

1. Add the latest version of any Storyblok package
2. Run a bundle analyzer

### Expected Behavior

Storyblok's API provides a clean and parsed JSON format that never includes data that must be parsed client-side. The richtext package only contains rendering logic and is not responsible for parsing data.

### Actual Behavior

The richtext package is responsible for dealing with any number of unvalidated and unstructured data formats that it must parse, including with a parser that's sufficiently advanced to account for all sorts of edge-cases.

### Code Sample

```javascript

```

### Environment

```shell
Not related to a specific environment
```

### Error Logs

```shell
Not related to specific error logs
```

mmeester:
Our team ran into the exact same issue. I also believe parsing should happen server‑side, not in the client. Pulling in Tiptap as a dependency for client-facing packages feels unnecessary and makes the setup needlessly heavy.

adambrudzynski:
[...]

PR #600, opened 2026-05-08
perf(astro): externalize deps and split entry points for tree shaking

### LINKED ISSUE #599: [Bug]: `@storyblok/astro` package root pulls full pre-bundle (dev toolbar, bridge loader, ProseMirror) into consumer output
### Package

@storyblok/astro

### Bug Description

`@storyblok/astro` ships as a single pre-bundled ESM file (`dist/storyblok-astro.es.js`, \~1.1 MB minified). Importing **any** symbol from the package root drags the entire bundle — including the dev toolbar, the live-preview message handler, `@storyblok/js`, `@storyblok/richtext` (TipTap + ProseMirror), `storyblok-js-client`, and `morphdom` — into the consumer’s worker/SSR output.

This happens because the Vite library-mode build emits a single `storyblok-astro.es.js` and the rollup `external` list is `['astro', '@astrijs/check', 'typescript']` (the middle entry is a typo of `@astrojs/check` — harmless, since `@astrojs/check` isn’t actually imported). Every other dependency (`@storyblok/js`, `@storyblok/richtext`, `storyblok-js-client`, `morphdom`, `camelcase`) is inlined. Even `dist/components/StoryblokComponent.astro`, which is shipped as a separate file, transitively pulls the pre-bundle through its `import { toCamelCase } from '@storyblok/astro'` line.

This is the same bug class as storyblok/monoblok#471 (richtext bundle bloat) and storyblok/monoblok#583 (`@storyblok/react/rsc` barrel pulls live-editing + ProseMirror into client bundles), in a different package. PR storyblok/monoblok#572 addressed it for `@storyblok/js` by externalizing deps and adding `"sideEffects": false`. This issue asks for the same treatment to be applied one layer up, in `@storyblok/astro`.

### Steps to Reproduce

```bash
mkdir sb-astro-bundle-test && cd sb-astro-bundle-test
npm init -y && npm pkg set type=module
npm i @storyblok/astro@9.0.5 esbuild

cat > entry.js <<'EOF'
import { storyblokEditable } from '@storyblok/astro';
globalThis.x = storyblokEditable;
EOF

npx esbuild entry.js --bundle --minify --format=esm \
  --platform=browser --conditions=import,browser,module \
  --outfile=out.js
ls -lh out.js
```

### Reproduction URL

*No response*

### Expected Behavior

`storyblokEditable` is a pure data-formatting helper that returns HTML attributes — it has no runtime requirement on the live-preview bridge, the dev toolbar, or `@storyblok/richtext`. A consumer importing only `storyblokEditable` should see ≤ 1 KB of `@storyblok/astro` code in their output, plus the externalized `@storyblok/js` chunk (which is itself \~32 KB minified after storyblok/monoblok#572 and tree-shakes to a few hundred bytes when only `storyblokEditable` is reached).

### Actual Behavior

Importing **any** symbol from `@storyblok/astro` produces a bundle that includes the entire pre-bundled file. Measured against `@storyblok/astro@9.0.5` with esbuild (browser ESM, minified, gzipped):

| Scenario | Bundle (gz) |
| -- | -- |
| `import { storyblokEditable }` | 170.4 KB |
| `import { storyblok }` (integration) | 176.1 KB |
| `import { renderRichText }` | 172.2 KB |
| `import * as sb` | 184.7 KB |

### Environment

```shell
- `@storyblok/astro@9.0.5`
- `astro@6.3.1`
```

### Asks

There are three concrete sub-asks, in priority order:

1. **Switch to per-file ESM output.** Apply the same treatment as storyblok/monoblok#572 to `@storyblok/astro`’s `vite.config.ts`: externalize `@storyblok/js`, `@storyblok/richtext`, `storyblok-js-client`, `morphdom`, `camelcase`; set `preserveModules: true`; emit one ESM file per source file. Add `"sideEffects": false` to `package.json` so consumer bundlers can tree-shake.
2. **Make** `dist/components/StoryblokComponent.astro` import internally. Change its top-of-file `import { toCamelCase, type SbBlokData } from '@storyblok/astro'` to relative imports of the source files (`../utils/toCamelCase`, `../types`). Otherwise consumers who use `<StoryblokComponent>` still pull the whole package root.
3. **Add a** `@storyblok/astro/richtext` subpath export. A thin entry that re-exports `renderRichText` (and the surrounding types/helpers) from `@storyblok/richtext` directly, skipping `@storyblok/js`. Lets consumers who want richtext opt in by path rather than relying on the bundler to tree-shake across re-export chains. (Alternative: drop the root-level `renderRichText` re-export and document `@storyblok/richtext` as the canonical entry — but that’s a breaking change; the subpath is additive.)

### Additional Context

Cross-package issues with the same root cause:

* storyblok/monoblok#471 — `@storyblok/richtext` bundle size regression (TipTap)
* storyblok/monoblok#583 — `@storyblok/react/rsc` barrel pulls live-editing + ProseMirror into client bundles
* storyblok/monoblok#572 — already-merged fix for `@storyblok/js` (the pattern this issue asks to extend)

I have a working patch and bundle-size measurements for all three sub-asks combined. I’ll open it as a PR shortly.

adamchal:
## Summary

Externalizes runtime deps, switches `@storyblok/astro` to multi-entry `preserveModules` ESM output, and adds a `@storyblok/astro/richtext` subpath. Mirrors the pattern from #572 (which did the same for `@storyblok/js`), applied one layer up.

After this change, importing utilities like `storyblokEditable` or `storyblok` no longer drags the dev toolbar, live-preview message handler, `@storyblok/js`, `@storyblok/richtext` (TipTap + ProseMirror), `storyblok-js-client`, or `morphdom` into the consumer bundle. Routes that don’t use rich text are completely free of `@storyblok/richtext`.

Fixes #599.

## Problem

`packages/astro/vite.config.ts` produces a single bundled `dist/storyblok-astro.es.js` (~1.1 MB) with only `astro`, `@astrijs/check`, `typescript` in the external list. Every other dependency is inlined. Even `dist/components/StoryblokComponent.astro` transitively pulls the pre-bundle through its `import { toCamelCase } from '@storyblok/astro'` line. Consumers can’t tree-shake across the boundary.

(Side note: `@astrijs/check` in the existing external list is a typo of `@astrojs/check`; harmless because `@astrojs/check` isn’t actually imported. This PR fixes it as a free-rider — happy to split it out if you’d prefer.)

## Changes

**`packages/astro/vite.config.ts`**

- Multi-entry input (`index`, `richtext`, `types`, `utils/toCamelCase`) + `formats: ['es']`.
- Externalize `@storyblok/js`, `@storyblok/richtext`, `storyblok-js-client`, `morphdom`, `camelcase` (plus existing `astro`, `typescript`, and corrected `@astrojs/check`).
- `preserveModules: true` + `preserveModulesRoot: 'src'` — one output file per source file.

**`packages/astro/package.json`**

- `"sideEffects": false`.
- New `./richtext` subpath export.
- `main` / `module` / `exports["."]` repointed at `./dist/index.js` (per-file entry instead of the bundled blob).
- Dropped the UMD / `require` paths. The package is `"type": "module"` and consumed via Astro’s ESM module graph; the UMD output was unused in practice.
- Added `@storyblok/richtext` to `dependencies` (now used directly by `src/richtext.ts`; was previously transitive through `@storyblok/js`).

**`packages/astro/src/components/StoryblokComponent.astro`**

- Imports `toCamelCase` and `SbBlokData` from internal relative paths (`../utils/toCamelCase`, `../types`) instead of `@storyblok/astro`. After `preserveModules`, those resolve to `dist/utils/toCamelCase.js` and `dist/types.d.ts` (type-only) — both exist as standalone outputs.

**`packages/astro/src/richtext.ts`** (new)

- Thin entry that re-exports `renderRichText`, `richTextResolver`, `segmentStoryblokRichText`, and the rich-text types directly from `@storyblok/richtext` (skips `@storyblok/js`).

## Bundle measurements

esbuild, browser ESM, minified, gzipped — against `@storyblok/astro@9.0.5` (before) vs this branch (after). Reproduce with the snippet in #599.

| Scenario | Before (gz) | After (gz) | Delta |
|---|---:|---:|---|
| `import { storyblokEditable }` | 170.4 KB | **0.2 KB** | **−99.9%** |
| `import { storyblok }` | 176.1 KB | **5.7 KB** | **−97%** |
| `import { renderRichText }` (root) | 172.2 KB | 192.9 KB | +12% (¹) |
| `import { renderRichText } from '@storyblok/astro/richtext'` | n/a | 192.9 KB | new |
| `import * as sb` | 184.7 KB | 207.3 KB | +12% (¹) |

(¹) Routes that pull `@storyblok/richtext` grow slightly because the externalized `@storyblok/richtext@4.3.0` is heavier raw than what 9.0.5 inlined — the regression tracked in #471, orthogonal to this PR. Routes that don’t use rich text (the common case) are now free of it entirely.

## Test plan

- [x] `pnpm --filter @storyblok/astro build` — clean. Total dist JS output is ~22 KB (per-file ESM) versus 1.2 MB ES + 880 KB UMD before.
- [x] `pnpm --filter @storyblok/astro test:unit` — 67/67 passing.
- [x] `pnpm --filter @storyblok/astro test:types` — clean.
- [x] Functional smoke test against a real Astro 6 SSG site: installed the patched tarball over `@storyblok/astro@9.0.5`, exercised every public surface (`storyblok`, `storyblokEditable`, `storyblokInit`, `apiPlugin`, `loadStoryblokBridge`, `toCamelCase`, `renderRichText` from both root and the new `/richtext` subpath, plus the type re-exports), and type-checked clean under `astro/tsconfigs/strictest` with `verbatimModuleSyntax`.
- [x] CI green.
- [x] `test:e2e` (Cypress against the playground) — please run, I don’t have the env set up locally.

## Notes for reviewers

- This is a **non-breaking** change for the documented public API. Consumers doing `import { ... } from '@storyblok/astro'` keep working, and existing subpath exports (`./StoryblokComponent.astro`, `./FallbackComponent.astro`, `./StoryblokServerData.astro`, `./middleware.ts`, `./toolbarApp.ts`, `./client`) are preserved.
- The new `./richtext` subpath is additive. The root-level `renderRichText` re-export is preserved for back-compat; it just routes through `@storyblok/js` instead of skipping straight to `@storyblok/richtext`.

### DIFF
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -21,29 +21,31 @@
     "astro-sdk",
     "withastro"
   ],
+  "sideEffects": false,
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
-      "import": "./dist/storyblok-astro.es.js",
-      "require": "./dist/storyblok-astro.umd.js"
+      "import": "./dist/index.js"
+    },
+    "./richtext": {
+      "types": "./dist/richtext.d.ts",
+      "import": "./dist/richtext.js"
     },
     "./middleware.ts": {
       "types": "./dist/live-preview/middleware.d.ts",
-      "import": "./dist/live-preview/middleware.ts",
-      "require": "./dist/live-preview/middleware.ts"
+      "import": "./dist/live-preview/middleware.ts"
     },
     "./toolbarApp.ts": {
       "types": "./dist/dev-toolbar/toolbarApp.d.ts",
-      "import": "./dist/dev-toolbar/toolbarApp.ts",
-      "require": "./dist/dev-toolbar/toolbarApp.ts"
+      "import": "./dist/dev-toolbar/toolbarApp.ts"
     },
     "./client": "./dist/lib/client.ts",
     "./FallbackComponent.astro": "./dist/components/FallbackComponent.astro",
     "./StoryblokComponent.astro": "./dist/components/StoryblokComponent.astro",
     "./StoryblokServerData.astro": "./dist/components/StoryblokServerData.astro"
   },
-  "main": "./dist/storyblok-astro.es.js",
-  "module": "./dist/storyblok-astro.js",
+  "main": "./dist/index.js",
+  "module": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "files": [
     "*.d.ts",
@@ -72,6 +74,7 @@
   },
   "dependencies": {
     "@storyblok/js": "workspace:*",
+    "@storyblok/richtext": "workspace:*",
     "camelcase": "^8.0.0",
     "morphdom": "^2.7.8"
   },
--- a/packages/astro/src/components/StoryblokComponent.astro
+++ b/packages/astro/src/components/StoryblokComponent.astro
@@ -1,6 +1,7 @@
 ---
 import options from 'virtual:storyblok-options';
-import { toCamelCase, type SbBlokData } from '@storyblok/astro';
+import { toCamelCase } from '../utils/toCamelCase';
+import type { SbBlokData } from '../types';
 import type { AstroComponentFactory } from 'astro/runtime/server/index.js';
 import { storyblokComponents } from 'virtual:import-storyblok-components';
 
--- a/packages/astro/src/richtext.ts
+++ b/packages/astro/src/richtext.ts
@@ -1,0 +1,36 @@
+import type {
+  StoryblokRichTextNode,
+  StoryblokRichTextOptions,
+} from '@storyblok/richtext';
+import { richTextResolver } from '@storyblok/richtext';
+
+/**
+ * Render Rich Text using the default `@storyblok/richtext` resolver.
+ *
+ * Equivalent to `richTextResolver(options).render(data)`.
+ */
+export function renderRichText<T = string>(
+  data: StoryblokRichTextNode<T>,
+  options?: StoryblokRichTextOptions<T>,
+): T | undefined {
+  return richTextResolver(options).render(data);
+}
+
+export {
+  asTag,
+  BlockTypes,
+  ComponentBlok,
+  LinkTypes,
+  MarkTypes,
+  richTextResolver,
+  segmentStoryblokRichText,
+  TextTypes,
+} from '@storyblok/richtext';
+
+export type {
+  StoryblokRichTextDocumentNode,
+  StoryblokRichTextImageOptimizationOptions,
+  StoryblokRichTextNode,
+  StoryblokRichTextNodeTypes,
+  StoryblokRichTextOptions,
+} from '@storyblok/richtext';
--- a/packages/astro/vite.config.ts
+++ b/packages/astro/vite.config.ts
@@ -8,12 +8,31 @@ export default defineConfig(() => {
   return {
     build: {
       lib: {
-        entry: path.resolve(__dirname, 'src/index.ts'),
-        name: 'storyblokAstro',
-        fileName: format => `storyblok-astro.${format}.js`,
+        entry: {
+          'index': path.resolve(__dirname, 'src/index.ts'),
+          'richtext': path.resolve(__dirname, 'src/richtext.ts'),
+          'types': path.resolve(__dirname, 'src/types.ts'),
+          'utils/toCamelCase': path.resolve(__dirname, 'src/utils/toCamelCase.ts'),
+        },
+        formats: ['es'],
       },
       rollupOptions: {
-        external: ['astro', '@astrijs/check', 'typescript'],
+        external: [
+          'astro',
+          '@astrojs/check',
+          'typescript',
+          '@storyblok/js',
+          '@storyblok/richtext',
+          'storyblok-js-client',
+          'morphdom',
+          'camelcase',
+        ],
+        output: {
+          preserveModules: true,
+          preserveModulesRoot: 'src',
+          entryFileNames: '[name].js',
+          chunkFileNames: 'chunks/[name]-[hash].js',
+        },
       },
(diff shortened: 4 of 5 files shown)


derekolson:
Bump. Working on a new Astro site now and this fix seems like a no-brainer.

adamchal:
@dipankarmaikap any feedback for this one?

dipankarmaikap:
Hey @adamchal ,

Thank you for the pull request. At this moment, we're not planning to move forward with this approach. However, we do see value in the ideas presented, and we've already started internal discussions around whether this is a pattern we should adopt going forward.

Regarding the original concern about the Rich Text package being too heavy, we currently have an upcoming solution in the works. You can see the draft PR here: https://github.com/storyblok/monoblok/pull/608.

The goal is to significantly improve the Rich Text developer experience while keeping the package size to around 3 KB. This would allow us to continue shipping everything as a single package, without requiring users to install additional dependencies.

We're aiming to release this in the coming weeks.

Thank you again for the contribution and for taking the time to explore this approach.

adamchal:
@dipankarmaikap thanks for the quick and candid feedback!

Once #608 lands I’ll take another look at where the bytes are going, and see if there’s still an opportunity to make `@storyblok/astro` fully tree-shakeable, so that a Cloudflare Workers or other size-constrained edge-runtime project can use just the integration and `<StoryblokComponent>` without the bridge / api / init runtime ending up in the production bundle. If a clean path shows up, I'll update #599 with the fresh numbers and draft a new, more focused PR against post-#608 main.

Thanks again for the consideration, and for moving #608 forward.

PR #606, opened 2026-05-13
chore(ci): harden publish workflow

maoberlehner:
## Summary

- Make the publish workflow a no-cache path by adding a `cache` input to `.github/actions/setup-node` and calling it with `cache: "false"` from `publish.yaml`. This skips the shared pnpm/Nx cache composite and disables `actions/setup-node` package-manager caching while still running `pnpm install --frozen-lockfile`.
- Keep normal CI and preview workflows on the default cached setup path.
- Keep `pkg.pr.new.yml` from running on `main`, since protected release-branch commits are handled by the real publish workflow.
- Add `persist-credentials: false` to checkout steps that do not push, and add explicit `contents: read` permissions where those workflows need checkout.
- Add CODEOWNERS coverage for `.github/workflows/` and `.github/actions/` using `@storyblok/plugins`.

## Why

The publish job receives `id-token: write` for npm trusted publishing, so it should not restore or save dependency/Nx caches that could be influenced by other workflows. Strictly disabling caches in publish is simpler and safer than maintaining a separate cache namespace for the release path.

## Test plan

- [x] YAML parse for `.github/workflows/*` and `.github/actions/*`.
- [x] `git diff --check origin/main..HEAD`.
- [x] Confirmed `.github/actions/cache-monorepo/action.yaml` has no branch diff after dropping the obsolete cache namespace commit.
- [x] Confirmed `publish.yaml` keeps `contents: read` and `id-token: write`, uses `persist-credentials: false`, and calls setup-node with `cache: "false"`.
- [x] Confirmed CI and pkg.pr.new still use the default cached setup path.

Third-party action SHA pinning is intentionally left out of this pass.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,4 @@
-.github/workflows/ @storyblok/plugins
-.github/actions/ @storyblok/plugins
+.github/workflows/ @storyblok/dx
+.github/actions/ @storyblok/dx
 
 packages/region-helper @storyblok/plugins
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,4 @@
+.github/workflows/ @storyblok/plugins
+.github/actions/ @storyblok/plugins
+
 packages/region-helper @storyblok/plugins
--- a/.github/actions/setup-node/action.yaml
+++ b/.github/actions/setup-node/action.yaml
@@ -1,18 +1,32 @@
 name: Setup Node
 description: Setup node
+inputs:
+  cache:
+    description: Restore/save pnpm and Nx caches
+    required: false
+    default: "true"
 runs:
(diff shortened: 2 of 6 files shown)


.github/CODEOWNERS
```diff
@@ -1 +1,4 @@
+.github/workflows/ @storyblok/plugins
+.github/actions/ @storyblok/plugins
```
alexjoverm: @maoberlehner I think here you mean `@storyblok/dx` 😄  plugins teams is from PE

alexjoverm (approved):
Thanks @maoberlehner ! Just a small thing to change but no re-review required

.github/CODEOWNERS
maoberlehner: Updated ;)

PR #609, opened 2026-05-14
fix: consolidate rate limit, introduce new name and deprecate old one

alexjoverm:
Renames `api.maxConcurrency` to `api.rateLimit` across the CLI — the old name was ambiguous and bled into stream internals via ad-hoc `batchSize` parameters that bypassed the global config entirely.

Done as an **alias** -> old `maxConcurrency` still works but it's deprecated.

### What's included

- **`ApiConfig.rateLimit`** replaces `maxConcurrency` as the canonical field (default: `6`). The old name is kept as an optional deprecated field for backward compatibility.
- **`--api-rate-limit`** is the new CLI flag; `--api-max-concurrency` and `api.maxConcurrency` in config files still work but emit a deprecation warning and are forwarded to `rateLimit` at resolve time. Both are marked `@todo(next-major)` for removal.
- **`createConcurrencyLock()`** (`src/utils/concurrency.ts`) — shared utility that reads `getActiveConfig().api.rateLimit` at call time. Replaces the hardcoded `new Sema(12)` module-level constants in `assets/streams.ts` and `stories/streams.ts`, which were initialised before config resolution.
- **`batchSize` removed** from `StoriesStream` and `UpdateStream` — it was never passed by any caller and silently bypassed the global config.

Before/after for config files and CLI:

```diff
- api:
-   maxConcurrency: 10
+ api:
+   rateLimit: 10
```

```diff
- --api-max-concurrency 10
+ --api-rate-limit 10
```

Fixes WDX-302

### DIFF
--- a/packages/cli/src/commands/components/push/graph-operations/index.ts
+++ b/packages/cli/src/commands/components/push/graph-operations/index.ts
@@ -30,7 +30,7 @@ export async function pushWithDependencyGraph(
   space: string,
   spaceState: SpaceComponentsDataState,
 
-  maxConcurrency: number = getActiveConfig().api.maxConcurrency,
+  maxConcurrency: number = getActiveConfig().api.rateLimit,
 ): Promise<PushResults> {
   // Build and validate the dependency graph with colocated target data
   const context: GraphBuildingContext = { spaceState };
--- a/packages/cli/src/commands/components/push/graph-operations/index.ts
+++ b/packages/cli/src/commands/components/push/graph-operations/index.ts
@@ -30,7 +30,7 @@ export async function pushWithDependencyGraph(
   space: string,
   spaceState: SpaceComponentsDataState,
 
-  maxConcurrency: number = getActiveConfig().api.rateLimit,
+  backpressure: number = getActiveConfig().api.rateLimit,
 ): Promise<PushResults> {
   // Build and validate the dependency graph with colocated target data
   const context: GraphBuildingContext = { spaceState };
@@ -38,7 +38,7 @@ export async function pushWithDependencyGraph(
   validateGraph(graph);
 
   // Process all resources using the dependency graph
-  const results = await processAllResources(graph, space, maxConcurrency);
+  const results = await processAllResources(graph, space, backpressure);
 
   return results;
 }
--- a/packages/cli/src/utils/backpressure-lock.ts
+++ b/packages/cli/src/utils/backpressure-lock.ts
@@ -3,6 +3,9 @@ import { getActiveConfig } from '../lib/config';
 
 // 2× the per-second rate limit so the pipeline always has work queued for the
 // next API window without over-loading memory.
+// In edge-case scenarios like ~2 s request duration: 12 / 2 s = 6 req/s, saturating the
+// default 6 req/s rate limit. If requests take longer (e.g. 2.5 s →
+// 12 / 2.5 s = 4.8 req/s) the pipeline under-utilises the limit slightly
 const PIPELINE_BACKPRESSURE_MULTIPLIER = 2;
 const DEFAULT_PIPELINE_BACKPRESSURE = 12;
 
--- a/packages/mapi-client/src/utils/rate-limit.ts
+++ b/packages/mapi-client/src/utils/rate-limit.ts
@@ -1,130 +1,47 @@
-const DEFAULT_MAX_CONCURRENT = 6;
+import { RateLimit } from 'async-sema';
+
+const DEFAULT_REQUESTS_PER_SECOND = 6;
 const MAX_RATE_LIMIT = 1_000;
 
 export interface RateLimitConfig {
   /**
-   * Maximum number of concurrent in-flight requests.
+   * Maximum number of MAPI requests to start per second.
    * Defaults to 6. Capped at 1000.
    */
-  maxConcurrency?: number;
+  requestsPerSecond?: number;
   /**
-   * Dynamically adjust the rate limit based on the `X-RateLimit-Policy`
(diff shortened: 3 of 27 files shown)


dipankarmaikap (approved):
LGTM 🚀

maoberlehner (changes requested):
After digging deeper into how this maps to Storyrails, I want to challenge the current implementation.

The original misleading naming originated from me: I named the MAPI client's option `rateLimit`, but implemented it as an in-flight concurrency semaphore, not a real rate limiter. I also introduced `batchSize` on the migration streams to mean "pipeline backpressure" (gating how many calls reach the mapi client simultaneously) rather than anything batch related.

How I understand it:

- Storyrails' default MAPI throttle is a per-second rate limit: 6 req/s for paid spaces, 3 req/s for free spaces.
- Storyrails also has separate opt-in concurrency limits, but those are not what the 3/6 MAPI throttle values represent.
- The MAPI client `rateLimit` option currently behaves as a fixed-size in-flight semaphore, so `rateLimit: 6` means "6 simultaneous requests," not "6 requests per second."
- The stream-level `Sema` controls how many items the pipeline processes in parallel. That is pipeline backpressure, not API rate limiting.

Proposed direction for this PR:

1. Keep `rateLimit` as the public CLI flag/config field only if this PR makes it an actual rate limiter. `--api-rate-limit 6` should start at most 6 MAPI requests per second.
2. Change the MAPI client implementation from an in-flight semaphore to a fixed one-second request-rate limiter that matches Storyrails' MAPI throttle. `rateLimit: 6` should start at most 6 requests in the current wall-clock second and queue additional requests until the next second.
3. Rename the stream-level `Sema` helpers to reflect pipeline backpressure, for example `createPipelineBackpressureLock()` for the factory and `pipelineSlot` / `acquirePipelineSlot()` at call sites. Avoid names like `apiRateLimit`, `apiConcurrency`, or `createConcurrencyLock()` for this layer, because it controls stream work-item flow rather than MAPI request pacing. I'm currently running a performance benchmark to see if `backpressure` should be higher than `rateLimit` (as it was before), but looks like it does not matter. I'll add a comment with the results.

maoberlehner:
Benchmark:

## Results @ 1000 stories

Each timed run started with ~1000 stories already in the space (cleanup 1000 + push 1000).

| Run | Main (s) | PR (s) |
|---|---|---|
| #1 | 514 | 555 |
| #2 | 554 | 571 |
| #3 | 575 | 568 |
| Best | 514 | 555 |
| Median | 554 | 568 |
| Worst | 575 | 571 |

Intra-branch spread: main ±5.5%, PR ±1.5%. Median delta: PR is ~2.5% slower — within noise.

So, performance-wise, it does not matter much whether the back pressure limit (previously `batchSize`) is slightly higher or equal to the rate limit. Yet I'd recommend doing something like `rateLimit * 2` to be on the safe side. The only downside of a higher back pressure limit is that we need to keep 12 stories in memory, although only 6 can be processed anyway. But a potential downside of using the rate limit for back pressure limit too is that there might be overlapping situations where the mapi client might already be ready again to process the next story, but the back pressure semaphore is lagging a couple of milliseconds behind before it lets the pipeline proceed.

alexjoverm:
@maoberlehner feedback applied! I also made a couple of extra changes:
- I replaced our own custom made rate limit for async-sema in mapi-client as that part is fragile
- I removed the logic related to rate limit headers -> those don't exist yet, and we don't know yet how they'll look like so it's just dead logic we had in the client

packages/cli/src/commands/components/push/graph-operations/index.ts
```diff
@@ -30,7 +30,7 @@ export async function pushWithDependencyGraph(
   space: string,
   spaceState: SpaceComponentsDataState,
 
-  maxConcurrency: number = getActiveConfig().api.maxConcurrency,
+  maxConcurrency: number = getActiveConfig().api.rateLimit,
```
maoberlehner: Should we rename `maxConcurrency` here (and below) too?

packages/cli/src/utils/backpressure-lock.ts
```diff
@@ -0,0 +1,13 @@
+import { Sema } from 'async-sema';
+import { getActiveConfig } from '../lib/config';
+
+// 2× the per-second rate limit so the pipeline always has work queued for the
+// next API window without over-loading memory.
+const PIPELINE_BACKPRESSURE_MULTIPLIER = 2;
```
maoberlehner, line 9: Idea: as I was wondering: add a comment explaining that `12 / 2 sec = 6 req/s` (where `2 sec` = request duration), so if requests take longer than `2 sec`, `12` backpressure would be too low (because `12 / 2.5 sec = 4,8 req/s`, not utilizing the full 6 req/s rate limit), but we don't expect requests to take that long.

packages/cli/src/utils/backpressure-lock.ts
```diff
@@ -0,0 +1,13 @@
+import { Sema } from 'async-sema';
```
maoberlehner, line 1: nitpick: maybe rename to `backpressure-lock.ts`

packages/mapi-client/src/utils/rate-limit.ts
```diff
@@ -1,130 +1,47 @@
-const DEFAULT_MAX_CONCURRENT = 6;
+import { RateLimit } from 'async-sema';
```
maoberlehner, line 1: Nice improvement! Looked up the package size and it is `1.1kb`, well worth it!

maoberlehner (changes requested):

maoberlehner (approved):

PR #611, opened 2026-05-15
feat: allow configure cv management

alexjoverm:
### Problem

Both storyblok-js-client and @storyblok/api-client automatically track the cv (content version) returned by the CDN API and attach it to every subsequent published request. This is hardcoded with no opt-out.

This causes problems in scenarios where the cv should not be managed by the client:

SSR with edge/CDN caching — attaching a tracked cv defeats edge-cache reuse; every new deployment or publish changes the cv, making all CDN-cached responses stale.

Webhook-driven invalidation — when cache purging is triggered externally, having the client auto-track cv adds noise and unpredictability.

External cv control — teams that set cv manually per-request cannot prevent the client from overriding or overwriting the tracked value.

### Solution

Add a cv option to the cache configuration in both clients.

```js
// storyblok-js-client
new StoryblokClient({
  cache: { cv: 'manual' },
})
// @storyblok/api-client
createApiClient({
  cache: { cv: 'manual' },
})
```

Fixes WDX-428

### DIFF
--- a/packages/capi-client/src/index.ts
+++ b/packages/capi-client/src/index.ts
@@ -51,6 +51,17 @@ interface CacheConfig {
   /** Time-to-live in milliseconds for cached entries. @default 60_000 */
   ttlMs?: number;
   /**
+   * Controls how the `cv` (content version) query parameter is managed.
+   *
+   * - `'auto'` (default): automatically attach the tracked `cv` to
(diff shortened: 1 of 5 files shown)


dipankarmaikap (approved):
LGTM 🚀 Clean implementation, and the added test coverage looks solid.

PR #613, opened 2026-05-18
feat: add patch method

alexjoverm:
With new upcoming `patch` endpoints (a/b testing experiments), we need to add support in js-client for patch methods


dipankarmaikap (approved):
LGTM, I'm approving this PR. However, it's missing tests in the `index.test.ts` file. We already have tests for POST, PUT, and DELETE (Lines 758 to 819), and I think we should also add a similar test for PATCH here.

That said, this is not a blocker.

alexjoverm:
You're right, it was added but somehow didn't push it. Good catch!

PR #614, opened 2026-05-19
fix: simplify live preview typing and update Angular integration for generic API

dipankarmaikap:
This PR removes the dependency on `ISbStoryData` and simplifies the typing of `onStoryblokEditorEvent`.

The live preview package now uses a generic `unknown`-based story internally, and consumers define their own `Story` type.

We also enforce an **ESM-only build for the live preview package**, since it is browser-only.

### Changes

* Removed `ISbStoryData` dependency
* Made `onStoryblokEditorEvent` fully generic
* Angular SDK now defines `ISbComponentType` locally
* Adapted `onStoryblokEditorEvent<Story>((story) => ...)` usage
* Live preview package is now ESM-only (browser usage only)

### Usage

```ts
onStoryblokEditorEvent<Story>((story) => {
  console.log(story);
});
```

### DIFF
--- a/packages/live-preview/src/onStoryblokEditorEvent.ts
+++ b/packages/live-preview/src/onStoryblokEditorEvent.ts
@@ -1,14 +1,15 @@
 import type { BridgeParams } from '@storyblok/preview-bridge';
-import type { ISbComponentType, ISbStoryData } from 'storyblok-js-client';
-
 import { loadStoryblokBridge } from './loadStoryblokBridge';
 import { canUseStoryblokBridge } from './utils/canUseStoryblokBridge';
 
 /**
+ * Base story type used internally.
+ */
+/**
  * Internal listener registry for Storyblok `input` events.
  * Each listener receives the updated story data from the Visual Editor.
  */
-const inputListeners = new Set<(story: ISbStoryData) => void>();
+const inputListeners = new Set<(story: unknown) => void>();
 
 /**
  * Tracks whether the Storyblok Preview Bridge event listeners
@@ -48,7 +49,7 @@ async function initializeBridge(bridgeOptions?: BridgeParams): Promise<void> {
 
       if (event.action === 'input' && event.story) {
         for (const listener of inputListeners) {
-          listener(event.story as ISbStoryData);
+          listener(event.story);
         }
         return;
       }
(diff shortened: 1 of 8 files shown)


maoberlehner:
nitpick: the branch name does not follow our naming guidelines and should be a little bit more descriptive: At least: `fix/live-preview-types`

packages/live-preview/src/onStoryblokEditorEvent.ts
```diff
@@ -1,14 +1,15 @@
 import type { BridgeParams } from '@storyblok/preview-bridge';
-import type { ISbComponentType, ISbStoryData } from 'storyblok-js-client';
-
 import { loadStoryblokBridge } from './loadStoryblokBridge';
 import { canUseStoryblokBridge } from './utils/canUseStoryblokBridge';
 
+/**
+ * Base story type used internally.
+ */
```
maoberlehner, line 7: I don't quite understand why this comment is where it is. Seems unrelated to the code/comment following it.

packages/live-preview/tsdown.config.ts
```diff
@@ -1,10 +1,10 @@
 import { defineConfig } from 'tsdown';
 
 export default defineConfig({
-  format: ['esm', 'cjs'],
+  format: ['esm'],
   exports: true,
   sourcemap: true,
   dts: true,
-  attw: true,
+  attw: false,
```
maoberlehner, line 8: I think we can use `esm-only` profile instead of disabling it: https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/packages/cli/README.md#profiles

packages/live-preview/src/onStoryblokEditorEvent.ts
```diff
@@ -99,15 +87,15 @@ async function initializeBridge(bridgeOptions?: BridgeParams): Promise<void> {
  * ```
  */
 export async function onStoryblokEditorEvent<
-  T extends ISbComponentType<string> = ISbComponentType<string>,
+  T = unknown,
```
maoberlehner, line 90: Why do we make this a generic? We _know_ (can look up) what to expect from editor events.

The `storyblok-bridge` repo is not very specific, unfortunately.

But a somewhat honest type would be:

```ts
export type LivePreviewStory<TContent = Record<string, unknown>> = {
  id: number;
  uuid?: string;
  content?: TContent;
  [key: string]: unknown;
};

export async function onStoryblokEditorEvent<
  TContent = Record<string, unknown>,
>(
  callback: (story: LivePreviewStory<TContent>) => void,
  bridgeOptions?: BridgeParams,
): Promise<() => void> {
  // ...
}
```

packages/live-preview/src/onStoryblokEditorEvent.ts
```diff
@@ -1,14 +1,15 @@
...
 
+/**
+ * Base story type used internally.
+ */
 /**
  * Internal listener registry for Storyblok `input` events.
  * Each listener receives the updated story data from the Visual Editor.
  */
-const inputListeners = new Set<(story: ISbStoryData) => void>();
+const inputListeners = new Set<(story: unknown) => void>();
```
maoberlehner, line 12: See type below

maoberlehner (changes requested):

packages/live-preview/src/onStoryblokEditorEvent.ts
dipankarmaikap: @maoberlehner With this, we would still have the type mismatch issue, for example when using the Story type from the api-client.

Also, with your PR https://github.com/storyblok/monoblok/pull/625, do you think we should continue with this approach?

maoberlehner: Yeah, the shape of what is returned here is fundamentally incompatible with the `Story` coming from the CAPI. I need to change this in my PR, where I did it wrong. Go ahead with your approach here.

dipankarmaikap:
Closing this in favour of https://github.com/storyblok/monoblok/pull/631

PR #615, opened 2026-05-20
fix(cli): map asset internal_tag_ids between spaces on push

### LINKED ISSUE #509: 🐛 [Bug]: Pulling and pushing assets results in an "error"
### Package

storyblok (CLI)

### Bug Description

When pulling assets from one space and pushing to another space, the assets are created, but a failed status is returned and the process exits with code 1. The error log indicates an HTTP 422 error. Repeated attempts with trimmed down asset JSON files indicate the `internal_tags_ids` and `internal_tags_list` members may be the culprits; while the assets upload and are created successfully, the tags are not applied (they do exist before pull/push).

### Steps to Reproduce

1. Pull assets from one space. Assets are tagged with at least one tag.
2. Push assets to another space. Tags should already exist in that space.

### Expected Behavior

Assets should be uploaded to storage, created, and tagged. The process should exit successfully with exit code 0.

### Actual Behavior

Assets are uploaded to storage and created; however, they are not tagged. The process fails with exit code 1.

### Code Sample

```javascript

```

### Environment

```shell
storyblok components pull --separate-files --space=$STORYBLOK_SPACE_ID
storyblok datasources pull --separate-files --space=$STORYBLOK_SPACE_ID
```

### Error Logs

```shell
{"timestamp":"2026-03-16T14:02:24.131Z","level":"INFO","message":"Pushing assets started","context":{"runId":1773669744129,"command":"storyblok assets push","options":{"updateStories":false,"from":"1023424","space":"290824385118662","verbose":false,"apiMaxRetries":3,"apiMaxConcurrency":6,"logConsoleEnabled":false,"logConsoleLevel":"info","logFileEnabled":true,"logFileLevel":"info","logFileMaxFiles":10,"uiEnabled":true,"reportEnabled":true,"reportMaxFiles":10},"cliVersion":"4.16.2"}}
{"timestamp":"2026-03-16T14:02:27.541Z","level":"ERROR","message":"The request was well-formed but was unable to be followed due to semantic errors","context":{"runId":1773669744129,"command":"storyblok assets push","options":{"updateStories":false,"from":"1023424","space":"290824385118662","verbose":false,"apiMaxRetries":3,"apiMaxConcurrency":6,"logConsoleEnabled":false,"logConsoleLevel":"info","logFileEnabled":true,"logFileLevel":"info","logFileMaxFiles":10,"uiEnabled":true,"reportEnabled":true,"reportMaxFiles":10},"cliVersion":"4.16.2","error":{"name":"API Error","message":"The request was well-formed but was unable to be followed due to semantic errors","httpCode":422,"httpStatusText":"Unprocessable Content","stack":"API Error: The request was well-formed but was unable to be followed due to semantic errors\n    at handleAPIError (file:///.../nextjs-site/node_modules/.pnpm/storyblok@4.16.2_@tiptap+pm_e30e530c8c1414fab7670443669288b4/node_modules/storyblok/dist/index.mjs:722:11)\n    at updateAsset (file:///.../nextjs-site/node_modules/.pnpm/storyblok@4.16.2_@tiptap+pm_e30e530c8c1414fab7670443669288b4/node_modules/storyblok/dist/index.mjs:7118:5)\n    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at async createAsset (file:///.../nextjs-site/node_modules/.pnpm/storyblok@4.16.2_@tiptap+pm_e30e530c8c1414fab7670443669288b4/node_modules/storyblok/dist/index.mjs:7136:26)\n    at async processAsset (file:///.../nextjs-site/node_modules/.pnpm/storyblok@4.16.2_@tiptap+pm_e30e530c8c1414fab7670443669288b4/node_modules/storyblok/dist/index.mjs:7697:22)\n    at async file:///.../nextjs-site/node_modules/.pnpm/storyblok@4.16.2_@tiptap+pm_e30e530c8c1414fab7670443669288b4/node_modules/storyblok/dist/index.mjs:7723:43"},"errorCode":"422","context":{"assetId":154600769927051}}}
{"timestamp":"2026-03-16T14:02:27.558Z","level":"INFO","message":"Pushing assets finished","context":{"runId":1773669744129,"command":"storyblok assets push","options":{"updateStories":false,"from":"1023424","space":"290824385118662","verbose":false,"apiMaxRetries":3,"apiMaxConcurrency":6,"logConsoleEnabled":false,"logConsoleLevel":"info","logFileEnabled":true,"logFileLevel":"info","logFileMaxFiles":10,"uiEnabled":true,"reportEnabled":true,"reportMaxFiles":10},"cliVersion":"4.16.2","summary":{"assetFolderResults":{"total":0,"succeeded":0,"failed":0},"assetResults":{"total":1,"succeeded":0,"failed":1,"skipped":0}}}}
```

### Additional Context

CLI output:

Image

maoberlehner:
## Summary

- `storyblok assets push --from <SRC> --space <TGT>` previously forwarded source-space `internal_tag_ids` from pulled sidecars verbatim; the target space rejected them with HTTP 422 and assets were created but untagged (exit code 1).
- The push command now fetches the target space's asset internal tags, builds a `name → id` map, and translates each asset's `internal_tag_ids` via the sidecar's `internal_tags_list` names before the create/update call.
- **Tag names missing from the target space are now created there** (mirroring `components push`) so pushed assets keep their tags. Before the upsert pipeline the command gathers source tag names from the sidecars, diffs them against the target's existing asset tags, and creates the missing ones via `createAssetInternalTag` (`object_type: 'asset'`). Created IDs are merged into the `name → id` map.
- Creation is **best-effort**: a failed create (or a tag ID with no sidecar name to create from) falls back to the existing drop-and-warn, so behavior is never worse than before. Same-space pushes and dry runs skip creation.
- `internal_tags_list` is stripped from outgoing payloads (read-only server-side). Tagless sources omit `internal_tag_ids` entirely so the mapi-client skips the follow-up metadata PUT for assets without tag metadata.

Fixes WDX-332
Fixes #509

## Test plan

- [x] Reproduced 422 against real spaces (source asset tagged, target tag exists with same name, different ID).
- [x] After fix: `assets push` exits 0; target asset carries the target-space tag ID, not the source-space ID.
- [x] Cross-space push with a tag absent from the target creates the tag and maps the asset to the new ID.
- [x] `pnpm nx lint:fix storyblok` passes.
- [x] `pnpm nx run storyblok:test:types` passes.
- [x] `pnpm --filter storyblok exec vitest run` — 706 passed, 1 todo.
- [x] New tests cover: source→target ID mapping by name, creating a missing tag and mapping to the new ID, graceful degradation when a tag create fails (reference dropped, push still exits 0), dropped ID without sidecar name, omitted `internal_tag_ids` for tagless sources; plus unit tests for the gather/ensure helpers.
- [x] QA: pull tagged assets from one space, push to another; verify matching-name tags map to existing IDs and absent tags are created in the target, and process exits 0.

### DIFF
--- a/packages/cli/src/commands/assets/push/index.test.ts
+++ b/packages/cli/src/commands/assets/push/index.test.ts
@@ -59,6 +59,10 @@ const server = setupServer(
     { message: 'Not Found' },
     { status: 404 },
   )),
+  http.get('https://mapi.storyblok.com/v1/spaces/:spaceId/internal_tags', () => HttpResponse.json(
+    { internal_tags: [] },
+    { headers: { 'Total': '0', 'Per-Page': '100' } },
+  )),
 );
 
 const preconditions = {
@@ -199,6 +203,14 @@ const preconditions = {
       http.put(`https://mapi.storyblok.com/v1/spaces/${space}/asset_folders/:folderId`, () => HttpResponse.json(
         { message: 'Update failed' },
         { status: 500 },
+      )),
+    );
+  },
+  hasTargetAssetInternalTags(tags: Array<{ id: number; name: string }>, { space = DEFAULT_SPACE }: { space?: string } = {}) {
+    server.use(
+      http.get(`https://mapi.storyblok.com/v1/spaces/${space}/internal_tags`, () => HttpResponse.json(
+        { internal_tags: tags },
+        { headers: { 'Total': String(tags.length), 'Per-Page': '100' } },
       )),
     );
   },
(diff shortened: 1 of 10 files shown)


packages/cli/src/commands/assets/push/index.test.ts
```diff
@@ -1392,4 +1419,177 @@ describe('assets push command', () => {
...
+
+    await assetsCommand.parseAsync(['node', 'test', 'push', '--from', DEFAULT_SPACE, '--space', targetSpace]);
+
+    const createPayload = (actions.createAsset as unknown as Mock).mock.calls[0][0];
+    expect(createPayload).not.toHaveProperty('internal_tag_ids');
+    expect(createPayload).not.toHaveProperty('internal_tags_list');
+    expect(process.exitCode).not.toBe(1);
+  });
+
+  it('should create internal tags missing from the target space and map to the new tag ID', async () => {
```
maoberlehner, line 1507: What do you think, is this the right approach? Just changed it to dropping missing tags to now creating them after I saw that this is how we do it with components.

dipankarmaikap: I think your approach is perfect. As a developer using the CLI, I would want the CLI to create the missing tag.

dipankarmaikap (approved):
LGTM. 🚀

**Praise:** The test coverage added in this PR is excellent. The scenarios are comprehensive and make it much easier to reason about the behaviour.

I manually tested several scenarios, including assets with tags, without tags, and updating tags between spaces. Everything worked as expected.

**Observation:** I noticed that cross-space pushes without an existing manifest will create new assets, as there is no content-based deduplication. This appears to be existing behaviour rather than something introduced by this PR. I also assume there isn't a straightforward way to solve this without downloading and comparing assets across spaces, which would likely be quite expensive.

PR #618, opened 2026-05-21
feature(js-client): skip CORS preflight headers for browser delivery API requests

onefriendaday:
Browser delivery API requests in storyblok-js-client were sending headers that trigger CORS preflight (OPTIONS): Content-Type: application/json, SB-Agent, and SB-Agent-Version. Those headers are unnecessary for typical CDN GET calls and add an extra round trip in the browser.

This change omits those headers only when both are true:

Content Delivery API (no oauthToken)
Running in a browser (typeof window !== 'undefined')
Server-side delivery clients (Node, SSR, edge) still send the full header set. Management API behavior is unchanged.

### DIFF
--- a/packages/js-client/src/index.ts
+++ b/packages/js-client/src/index.ts
@@ -110,8 +110,14 @@ export class Storyblok {
     }
 
     const headers: Headers = new Headers();
+    const isManagementApi = !!config.oauthToken;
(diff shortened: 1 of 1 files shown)


alexjoverm:
Following up in https://github.com/storyblok/monoblok/pull/619

PR #624, opened 2026-05-26
feat(richtext)!: unify richtext rendering across SDKs with new static renderer API

dipankarmaikap:
This PR introduces the new `renderRichText` renderer API across all SDKs and aligns framework implementations with the new static richtext rendering architecture.

### Changes

* introduce `renderRichText` as the main richtext renderer API
* remove legacy `richTextResolver` and `segmentStoryblokRichText` APIs
* align React, Vue, Angular, and Svelte SDKs with the new renderer architecture
* improve type safety and customization support across framework integrations
* add/update `StoryblokRichText` implementations for framework parity
* re-export renderer utilities from JS SDK

### Breaking Changes

* `renderRichText` is now the only supported richtext rendering API
* legacy richtext renderer APIs have been removed

Migration guide: https://storyblok.com/mp/migrating-from-storyblok-richtext-v4-to-v5


packages/react/package.json
```diff
@@ -68,7 +68,8 @@
     }
   },
   "dependencies": {
-    "@storyblok/js": "workspace:*"
+    "@storyblok/js": "workspace:*",
+    "@storyblok/richtext": "workspace:*"
```
maoberlehner, line 72: `packages/react/vite.config.ts` externalizes `@storyblok/js` but not `@storyblok/richtext` leading to `@storyblok/richtext` being bundled into `@storyblok/react` while also remaining a dependency of the externalized `@storyblok/js`. Add `@storyblok/richtext` to the React Rollup external list to avoid duplicate richtext code and keep module resolution consistent.

packages/react/src/core/richtext-hoc.tsx
```diff
@@ -0,0 +1,332 @@
...
+}
+function resolveComponent<K extends SbRichTextElement>(
+  type: K,
+  components?: SbReactComponentMap,
+): ComponentType<SbReactRichTextProps<K>> | undefined {
+  return components?.[type] as ComponentType<SbReactRichTextProps<K>> | undefined;
+}
+export function createRichTextHook(StoryblokComponent: ElementType, _options?: CreateRichTextHookOptions) {
+  return function useRichText({ document, optimizeImage, components }: StoryblokRichtextProps) {
+    const render = useStoryblokRichText({
```
maoberlehner: The `useMemo` in `useStoryblokRichText` is mostly defeated because here we build a fresh merged components object on every render before passing it to `useStoryblokRichText`. Even if callers memoize their components, we wrap them in a new object, so `[optimizeImage, components]` changes and `createStoryblokRenderer` runs again.

I think this could be a potential fix:

```ts
export function createRichTextHook(StoryblokComponent: ElementType, _options?: CreateRichTextHookOptions) {
  const BlokRenderer: SbReactRichTextComponent<'blok'> = ({ attrs }) => (
    <>
      {Array.isArray(attrs?.body) && attrs.body.map((blok, index) => (
        <StoryblokComponent blok={blok} key={blok._uid || index} />
      ))}
    </>
  );

  return function useRichText({ document, optimizeImage, components }: StoryblokRichtextProps) {
    const mergedComponents = useMemo<SbReactComponentMap>(() => ({
      blok: BlokRenderer,
      ...components,
    }), [components]);

    const render = useStoryblokRichText({
      optimizeImage,
      components: mergedComponents,
    });

    return render(document);
  };
}
```

packages/richtext/src/static/normalize-nodes.ts
```diff
@@ -0,0 +1,45 @@
...
+  includeKeys = false,
+): SbRichTextNode[] {
+  if (!input) {
+    return [];
+  }
+  if (Array.isArray(input)) {
+    return input;
+  }
+  const nodes = Array.isArray(input)
+    ? input
```
maoberlehner: This is unreachable as we already check in line 15 if it is an array.

packages/vue/src/composables/useStoryblokRichText.ts
```diff
@@ -1,27 +1,360 @@
...
+export interface StoryblokRichTextProps extends SbVueRichTextRenderContext {
+  document: MaybeRefOrGetter<SbRichTextNode | SbRichTextNode[] | null | undefined>;
+}
+
+interface CreateRichTextHookOptions {
+  isServerContext?: boolean;
+}
+
+/**
+ * Creates a richtext hook factory with a custom StoryblokComponent.
```
maoberlehner: We don't use the `hook` naming in Vue world. Functions dealing with reactivity prefixed with `use` are called `composables`. So I think we should avoid the term `hook` across the whole Vue codebase.

packages/vue/src/composables/useStoryblokRichText.ts
```diff
@@ -1,27 +1,360 @@
...
+ */
+export function createRichTextHook(
+  _options?: CreateRichTextHookOptions,
+) {
+  return function useRichText({ document, optimizeImage, components }: StoryblokRichTextProps) {
+    const render = useStoryblokRichText({
```
maoberlehner: General note: I find it very hard to understand the code because I need to take a lot of hops to get somewhere where stuff is happening.

I've started with `createRichTextHook` > `useRichtext` > `useStoryblokRichText` > `createStoryblokRenderer` > `renderChildren` and only here really interesting stuff starts happening.

Can we remove a few layers here maybe?

packages/vue/src/composables/useStoryblokRichText.ts
```diff
@@ -1,27 +1,360 @@
...
+
+    return computed(() => render(toValue(document)));
+  };
+}
+
+/**
+ * Vue composable for rendering rich text content.
+ * Returns a render function that converts rich text JSON to VNodes.
+ */
+export function useStoryblokRichText(options: SbVueRichTextRenderContext = {}) {
```
maoberlehner: What makes this a composable? No reactive logic here as far as I can see.

packages/vue/src/composables/useStoryblokRichText.ts
```diff
@@ -1,27 +1,360 @@
...
+
+interface CreateRichTextHookOptions {
+  isServerContext?: boolean;
+}
+
+/**
+ * Creates a richtext hook factory with a custom StoryblokComponent.
+ */
+export function createRichTextHook(
+  _options?: CreateRichTextHookOptions,
```
maoberlehner: What are the options for?

packages/vue/src/components/StoryblokRichText.vue
```diff
@@ -1,27 +1,38 @@
...
 
-const renderedDoc = ref();
-const root = () => renderedDoc.value;
+  setup(props) {
+    const useStoryblokRichText = createRichTextHook();
+    const rendered = useStoryblokRichText({
+      optimizeImage: props.optimizeImage,
+      components: toRaw(props.components),
+      document: props.document,
+    });
```
maoberlehner, line 32: There is a practical problem with reactivity passing props here like that. But my larger concern is regarding the use of a composable for rendering in general. I don't think a composable should return VNodes. Components are for rendering. Composables are for reactive data logic.

I think you can vastly simplify this by writing it like this:

```ts
export default defineComponent({
  name: 'StoryblokRichText',
  props: {
    document: {
      type: Object as PropType<SbRichTextNode | SbRichTextNode[] | null | undefined>,
      required: false,
    },
    optimizeImage: {
      type: [Boolean, Object] as PropType<boolean | SbRichTextImageOptions>,
      required: false,
    },
    components: {
      type: Object as PropType<SbVueRichTextComponentMap>,
      required: false,
    },
  },
  setup(props) {
    const render = createStoryblokRenderer({
      optimizeImage: props.optimizeImage,
      components: { blok: BlokRenderer, ...props.components },
    });
    return () => render(props.document);
  },
});
```

maoberlehner: If you decide to go that route, you can delete the composable and move `createStoryblokRenderer` and related functions into a more fitting file (are no composables).

packages/react/src/server/server-storyblok-richtext-component.tsx
```diff
@@ -1,24 +1,33 @@
...
-    const formattedHtml = convertAttributesInElement(html as React.ReactElement);
+type UnwrappedProps = StoryblokRichtextProps & {
+  wrapper: false;
+};
 
+type Props = WrappedProps | UnwrappedProps;
+
+const StoryblokRichText = forwardRef<HTMLDivElement, Props>(
+  ({ document, optimizeImage, components, wrapper = true, ...rest }, ref) => {
+    const html = useStoryblokServerRichText({ document, optimizeImage, components });
```
maoberlehner: nitpick: `html` isn't really true here. Maybe find a more honest name like `children` or `nodes` or `richtext`?

maoberlehner (changes requested):
I have not finished reviewing yet, but I have already found a couple of things that have a large surface area, so here is my first part of the review.

In general: I know that file naming was already chaotic before, yet we should work on improving it, and a large PR like this is the perfect opportunity. E.g. `useStoryblokRichText.ts` should be `use-storyblok-rich-text.ts`, 

For React components it's also kind of a mix. E.g. `server-storyblok-richtext-component.tsx` -> `component` is redundant, and other React components are PascalCase. The component is named `StoryblokRichText` but in the file name, it is `richtext` (instead of `richt-text`.

I think we should consistently name all regular files kebab-case and components depending on the convention we've followed so far. With React and Vue, we mostly have PascalCase, Angular kebap. And except when it's about the package name itself (which is `richtext`, we should stick to writing `richt text`/`richt-text`/`RichText`.

packages/vue/src/composables/useStoryblokRichText.ts
```diff
@@ -1,27 +1,360 @@
...
+/**
+ * Builds Vue props from node/mark type and attrs.
+ * Handles class attribute (Vue uses 'class' not 'className').
+ */
+function buildVueProps(
+  type: SbRichTextElement,
+  attrs: Record<string, unknown> | undefined,
+): Record<string, unknown> {
+  // Vue uses 'class' directly, unlike React which needs 'className'
+  const processed = processAttrs(type, attrs);
```
maoberlehner: Why not use `processAttrs` directly?

packages/react/src/core/richtext-hoc.tsx
```diff
@@ -0,0 +1,332 @@
...
+): Record<string, unknown> {
+  const processed = processAttrs(type, attrs, {
+    class: 'className',
+  });
+
+  // Convert style object to React CSSProperties format
+  // The style from processAttrs is already an object with camelCase keys
+  if (processed.style && typeof processed.style === 'object') {
+    // Style is already in the correct format from processAttrs
+    return processed as Record<string, unknown>;
```
maoberlehner: Casting `processed` to `Record<string, unknown>` does nothing here.

packages/react/src/core/richtext-hoc.tsx
```diff
@@ -0,0 +1,332 @@
...
+  type: SbRichTextElement,
+  attrs: Record<string, unknown> | undefined,
+): Record<string, unknown> {
+  const processed = processAttrs(type, attrs, {
+    class: 'className',
+  });
+
+  // Convert style object to React CSSProperties format
+  // The style from processAttrs is already an object with camelCase keys
+  if (processed.style && typeof processed.style === 'object') {
```
maoberlehner: I don't get this `if`. If it is true, exactly the same happens as if it is not (we return `processed` as is). `Convert style object to React CSSProperties format` -> this is a lie.

maoberlehner: I think `buildReactProps` can just be removed and replaced by using `processAttrs(type, attrs, { class: 'className' });` directly.

packages/richtext/src/render-richtext.ts
```diff
@@ -26,47 +19,41 @@ type TextNode = PMNode & { type: 'text' };
...
+function renderNode(node: SbRichTextNode, context?: SbRichTextRenderContext): string {
   if (node.type === 'text') {
-    return renderTextNode(node as TextNode, node.marks, options);
+    return renderTextNode(node, node.marks, context);
   }
-
+  const content = node.content ? renderChildren(node.content, context) : '';
   // Custom renderer takes full control
-  const customRenderer = options?.renderers?.[node.type as keyof typeof options.renderers];
+  const customRenderer = context?.renderers?.[node.type];
```
maoberlehner: I was liking the cast here more as it keeps the line below easier to read.

Maybe we can introduce:
```ts
type AnyRenderer = (props: SbRichTextNode & { children: string; context?: SbRichTextRenderContext }) => string

// allowing
const customRenderer = context?.renderers?.[node.type] as AnyRenderer;
```

packages/angular/src/lib/richtext/rich-text.component.ts
```diff
@@ -85,39 +86,25 @@ export class SbRichTextComponent implements OnDestroy {
   // --------------------------------------------------
   // Render entry
   // --------------------------------------------------
-  private render(sbDocument: SbRichTextDoc | SbRichTextDoc[] | null | undefined): void {
+  private render(sbDocument: SbRichTextNode | SbRichTextNode[] | null | undefined): void {
```
maoberlehner: Is the param name `sbDocument` still correct when it is now a Node?

packages/angular/src/lib/types.ts
```diff
@@ -1,10 +1,10 @@
...
+import { SbRichTextElement, SbRichTextProps } from '@storyblok/richtext';
 
 export type SbBlokKeyDataTypes = string | number | object | boolean | undefined;
 
 export interface SbBlokData extends ISbComponentType<string> {
   [index: string]: SbBlokKeyDataTypes;
 }
 
-export type RichTextComponentProps<T extends SbRichTextElement> = SbRichTextProps<T>;
+export type SbAngularRichTextProps<T extends SbRichTextElement> = SbRichTextProps<T>;
```
maoberlehner: Not sure why this is necessary. Why not use `SbRichTextProps` directly?

packages/astro/src/components/richtext/RenderLinkInner.astro
```diff
@@ -0,0 +1,18 @@
+---
+import type { SbAstroRichTextRenderContext, SbRichTextNode, SbRichTextTextNode } from '../../';
+import { getInnerMarks } from '../../';
+import RenderTextNodeWithMarks from './RenderTextNodeWithMarks.astro';
```
maoberlehner, line 4: nitpick: I think the `Render` prefix is redundant. The fact that it is a component already tells us that it renders. Or to put it differently: If we add `Render` prefix here, why not `RenderLinkInner` too?

packages/astro/src/components/richtext/LinkInner.astro
```diff
@@ -0,0 +1,18 @@
+---
+import type { SbAstroRichTextRenderContext, SbRichTextNode, SbRichTextTextNode } from '../../';
+import { getInnerMarks } from '../../';
+import RenderTextNodeWithMarks from './RenderTextNodeWithMarks.astro';
+
+interface Props {
+  nodes: SbRichTextNode[];
```
maoberlehner: Either this should be `SbRichTextTextNode` or the `as` cast below should be replaced with an actual type check.

packages/astro/src/components/richtext/RenderElement.astro
```diff
@@ -0,0 +1,34 @@
+---
```
maoberlehner: Same nitpick as above: does `Render` prefix add value?

packages/astro/src/components/richtext/RenderElement.astro
```diff
@@ -0,0 +1,34 @@
...
+  type SbRichTextNode,
+} from '../../';
+import DynamicElement from './DynamicElement.astro';
+interface Props {
+  node: SbRichTextNode | SbRichTextMark;
+}
+const { node } = Astro.props;
+
+const Tag = resolveTag(node);
+const tagName = Tag || '';
```
maoberlehner: What is this for? Why not use `<${Tag}${attrsToHtmlString(processedAttrs)}>`?

packages/astro/src/components/richtext/RenderElement.astro
maoberlehner: (same question for all `Render` prefixes)

packages/astro/src/components/richtext/DynamicElement.astro
```diff
@@ -0,0 +1,9 @@
+---
```
maoberlehner: Name might be confusing as below `RenderElement` means a rich text _element_ while here it means a HTML _element_, if I understand it correctly.

maoberlehner: Not quite sure how much value this component adds in general, as below we could:

```diff
-<DynamicElement tag={Tag} attrs={processedAttrs}>
-  <slot />
-</DynamicElement>
+<Tag attrs={processedAttrs}>
+  <slot />
+</Tag>
```

packages/astro/src/components/richtext/RenderLinkGroup.astro
```diff
@@ -0,0 +1,26 @@
+---
+import { type SbRichTextNode, type SbAstroRichTextRenderContext, type SbRichTextMark, isValidAstroComponent } from '../../';
+import LinkInner from './LinkInner.astro';
+import RenderElement from './RenderElement.astro';
+
+interface Props {
+  nodes: SbRichTextNode[];
+  linkMark: SbRichTextMark;
```
maoberlehner: We should either name it `mark` or only allow link mark: `Extract<SbRichTextMark, { type: 'link' }>`.

packages/astro/src/components/richtext/DynamicElement.astro
maoberlehner: Ah, after reading more code, I realize that both `DynamicElement` and `RenderElement` refer to HTML _elements_. So what confused me about the naming was that I thought `RenderElement` renders a `SbRichTextElement`. I think we shouldn't use the two concepts interchangeably. When an "element" in this codebase is either a `node` or a `mark`, using it here to mean HTML _element_ is confusing.

Naming suggestion below.

packages/astro/src/components/richtext/RenderNode.astro
```diff
@@ -0,0 +1,48 @@
...
+        node.attrs.body.map((blok, index) => <StoryblokComponent blok={blok} key={blok._uid || index} />)}
+    </Fragment>
+  ) : node.type === 'table' ? (
+    <RenderTable node={node} options={options} />
+  ) : getStaticChildren(node) ? (
+    <RenderStaticChildren node={node} options={options} />
+  ) : (
+    <RenderElement node={node}>
+      {node.content && <RenderChildren nodes={node.content} options={options} />}
+    </RenderElement>
```
maoberlehner: Seeing those two components, and the `SbRichTextElement` where `Element` means either `node` or `mark`, I would expect that I can use `RenderElement` to render any `node` or `mark`. Yet I now understand the `Element` here refers to rendering an HTML _element_. I propose that instead we rename `RenderElement` to `ElementTag`.

packages/astro/src/components/richtext/RenderStaticChildren.astro
```diff
@@ -0,0 +1,24 @@
+---
```
maoberlehner, line 1: nitpick: naming doesn't seem quite right as we're rendering not only the children.

packages/astro/src/components/richtext/RenderStaticChildren.astro
```diff
@@ -0,0 +1,24 @@
...
+import { type SbAstroRichTextRenderContext, type SbRichTextNode, getStaticChildren, resolveTag } from '../../';
+import DynamicElement from './DynamicElement.astro';
+import RenderChildren from './RenderChildren.astro';
+import RenderStaticStructure from './RenderStaticStructure.astro';
+
+interface Props {
+  node: SbRichTextNode;
```
maoberlehner: nitpick: we could make the types more precise here and only allow `node.type !== 'text'`

maoberlehner: Looking closer, only `code_block` currently has `children`, so we probably should only allow nodes of type `code_block` here? Or make it future safe like this:

```ts
import { NODE_RENDER_MAP } from './render-map.generated';
  import type { SbRichTextElementByType } from './types.generated';

  /** Node types whose render-map entry declares a static `children` structure (e.g. code_block → pre>code). */
  export type StaticChildrenNodeType = {
    [K in keyof typeof NODE_RENDER_MAP]:
      (typeof NODE_RENDER_MAP)[K] extends { children: unknown[] } ? K : never;
  }[keyof typeof NODE_RENDER_MAP];

  /** Richtext nodes that carry a static children structure. */
  export type SbRichTextStaticChildrenNode = SbRichTextElementByType[StaticChildrenNodeType];

  Today StaticChildrenNodeType resolves to 'code_block'; the day codegen emits another node with children, this widens automatically — no edits. (null entries like doc/text and plain { tag } entries fail extends {
  children: unknown[] } → never and drop out.)

  Use it in the component

  // RenderStaticChildren.astro
  import type { SbRichTextStaticChildrenNode } from '../../';

  interface Props {
    node: SbRichTextStaticChildrenNode;
    options: SbAstroRendererOptions;
  }
```

packages/astro/src/components/richtext/RenderStaticStructure.astro
```diff
@@ -0,0 +1,36 @@
...
+    const processedAttrs = buildAstroAttrs(type, mergedAttrs);
+
+    if (isSelfClosing(Tag)) {
+      return <Fragment set:html={`<${tag}${attrsToHtmlString(processedAttrs)}>`} />;
+    }
+    return (
+      <DynamicElement tag={Tag} attrs={processedAttrs}>
+        {children ? (
+          <Astro.self type={type} specs={children} parentAttrs={parentAttrs}>
+            <slot />
```
maoberlehner, line 27: So if `specs` has length `3`, we show three times the same `slot`? Does not seem like something we want.

Is this really a generic case or should we just have a `CodeBlock` component? What other nodes might have children? And might there be nodes with more than one child?

packages/astro/src/components/richtext/RenderTable.astro
```diff
@@ -0,0 +1,16 @@
+---
+import { type SbRichTextNode, type SbAstroRichTextRenderContext, splitTableRows } from '../../';
+import RenderElement from './RenderElement.astro';
+import RenderNode from './RenderNode.astro';
+interface Props {
+  node: SbRichTextNode;
```
maoberlehner: Shouldn't we only allow table nodes here?

packages/astro/src/components/richtext/RenderElement.astro
```diff
@@ -0,0 +1,34 @@
...
+  attrsToHtmlString,
+  buildAstroAttrs,
+  resolveTag,
+  isSelfClosing,
+  type SbRichTextMark,
+  type SbRichTextNode,
+} from '../../';
+import DynamicElement from './DynamicElement.astro';
+interface Props {
+  node: SbRichTextNode | SbRichTextMark;
```
maoberlehner: Should we disallow all types here that need special handling in `RenderNode`? As I understand it, passing a node with type `text`, `image`, `blok`, `table` here is not something we want?

packages/astro/src/components/richtext/RenderTextNodeWithMarks.astro
```diff
@@ -0,0 +1,34 @@
...
+import {
+  type SbRichTextTextNode,
+  type SbRichTextMark,
+  type SbAstroRichTextRenderContext,
+  isValidAstroComponent,
+} from '../../';
+import RenderElement from './RenderElement.astro';
+
+interface Props {
+  node: SbRichTextTextNode;
```
maoberlehner, line 11: Limiting allows types possible/beneficial here?

packages/astro/src/components/richtext/RenderTextNodeWithMarks.astro
```diff
@@ -0,0 +1,34 @@
...
+  marks: SbRichTextMark[] | undefined;
```
maoberlehner: nitpick: `marks?: SbRichTextMark[];`

packages/astro/src/components/StoryblokRichText.astro
```diff
@@ -0,0 +1,14 @@
+---
+import { type SbRichTextNode, type SbAstroRichTextRenderContext, normalizeNodes } from '../';
+import RenderChildren from './richtext/RenderChildren.astro';
+
+interface Props extends SbAstroRichTextRenderContext {
+  document: SbRichTextNode | null | undefined | SbRichTextNode[];
```
maoberlehner: I find the naming confusing. `SbRichTextNode` can have a type `doc`. With the prop called `document` I expect only type `doc` to be allowed but it seems to work with any node(s).

packages/astro/vite.config.ts
```diff
@@ -4,53 +4,49 @@ import path from 'node:path';
...
-      },
+export default defineConfig({
+  build: {
+    lib: {
+      entry: path.resolve(__dirname, 'src/index.ts'),
+      formats: ['es'],
+      fileName: () => 'index.js',
+    },
+    rollupOptions: {
+      external: ['astro', '@astrijs/check', 'typescript'],
```
maoberlehner: typo in `@astrijs/check` but I think neither adding `@astrojs/check` or `typescript` here does anything so I think we can remove them.

packages/react/src/core/richtext-hoc.tsx
```diff
@@ -0,0 +1,332 @@
...
+    // Style is already in the correct format from processAttrs
+    return processed as Record<string, unknown>;
+  }
+
+  return processed;
+}
```
maoberlehner: The whole file feels _weird_.

- Why is it called `richtext-hoc` when a higher-order component is nowhere to be seen?
- Why `render*` functions instead of making them components e.g. `renderLinkGroup` -> `LinkGroup` (same for Vue equivalent)?
- I don't understand why we even have rich text hooks. Rendering is a component concern and not a hook concern.

packages/react/src/storyblok-rich-text.tsx
```diff
@@ -1,21 +1,33 @@
...
+  ({ document, optimizeImage, components, wrapper = true, ...rest }, ref) => {
+    const html = useStoryblokRichText({ document, optimizeImage, components });
+    if (!wrapper) {
+      return html;
+    }
     return (
-      <div ref={ref}>
-        {formattedHtml}
+      <div ref={ref} {...rest}>
+        {html}
```
maoberlehner: At the little cost of `forwardRef` and elimination of `useMemo` (which probably didn't do much to improve performance) I think we this could be much simpler and completely remove the need for a separate server variant:

```tsx
// Default `blok` renderer. Module-scoped → stable identity (no per-render
// remounts) and no client-only deps.
function DefaultBlok({ attrs }: SbRichTextProps<'blok'>) {
  if (!Array.isArray(attrs?.body)) {
    return null;
  }
  return (
    <>
      {attrs.body.map((blok, index) => (
        <StoryblokComponent blok={blok} key={blok._uid || index} />
      ))}
    </>
  );
}

type DivProps = Omit<ComponentPropsWithoutRef<'div'>, 'children'>;

type WrappedProps = StoryblokRichtextProps & DivProps & { wrapper?: true };
type UnwrappedProps = StoryblokRichtextProps & { wrapper: false };
export type StoryblokRichTextProps = WrappedProps | UnwrappedProps;

/**
  * Renders Storyblok richtext. Universal: no hooks or refs, so it works as a
  * React Server Component and on the client. Rendering is a pure transform of
  * `document`, so there is nothing to memoize across renders on the server; on
  * the client, wrap in a memoized parent if a hot re-render path needs it.
  */
function StoryblokRichText({
  document,
  optimizeImage,
  components,
  wrapper = true,
  ...rest
}: StoryblokRichTextProps): ReactNode {
  const content = createStoryblokRenderer({
    optimizeImage,
    components: { blok: DefaultBlok, ...components },
  })(document);

  if (!wrapper) {
    return content;
  }

  return <div {...rest}>{content}</div>;
}

export default StoryblokRichText;
```

packages/svelte/src/lib/richtext/ElementTag.svelte
```diff
@@ -0,0 +1,28 @@
+<script lang="ts">
```
maoberlehner, line 1: What was written about `Render` prefix above, applies here too.

packages/svelte/src/lib/richtext/RenderLinkGroup.svelte
```diff
@@ -0,0 +1,26 @@
+<script lang="ts">
+  import { type SbRichTextMark, type SbRichTextNode } from '@storyblok/richtext';
+
+  import type { SbSvelteRichTextRenderContext } from '../richtext-helpers';
+  import RenderElement from './RenderElement.svelte';
+  import LinkInner from './LinkInner.svelte';
+
+  type Props = {
+    nodes: SbRichTextNode[];
+    linkMark: SbRichTextMark;
```
maoberlehner: In general, a lot of comments from Astro files apply here too, please double-check. Thx!

packages/svelte/src/lib/richtext/RenderTable.svelte
```diff
@@ -0,0 +1,26 @@
...
+<RenderElement {node}
+  >{#if headerRows.length > 0}<thead>
+      {#each headerRows as row (row._key)}
+        <RenderNode node={row} {options} />
+      {/each}
+    </thead>{/if}{#if bodyRows.length > 0}<tbody>
+      {#each bodyRows as row (row._key)}
+        <RenderNode node={row} {options} />
+      {/each}
+    </tbody>{/if}</RenderElement
```
maoberlehner: nitpick: indentation is off

maoberlehner (changes requested):

packages/react/package.json
dipankarmaikap: Good point. Fixed

packages/richtext/src/static/normalize-nodes.ts
dipankarmaikap: Fixed.

packages/vue/src/composables/useStoryblokRichText.ts
dipankarmaikap: Good point. I was mainly following the existing API pattern. But you're right, if we're being technical, this isn't really a composable anymore, it's more of a factory function that creates a renderer.

I've kept the `useStoryblokRichText` name for now to avoid introducing a new API and to maintain backwards compatibility, but the implementation is much cleaner now.

packages/vue/src/composables/useStoryblokRichText.ts
dipankarmaikap: Thank you, updated the comment.

packages/react/src/core/richtext-hoc.tsx
dipankarmaikap: Good point. I've refactored this to make it much simpler to use. Similar to the Vue implementation, it's not technically a hook anymore either, since it doesn't rely on any React hook functionality. The new approach is essentially a renderer factory, which makes the implementation simpler and easier to reason about.

packages/vue/src/composables/useStoryblokRichText.ts
dipankarmaikap: Removed

packages/vue/src/components/StoryblokRichText.vue
dipankarmaikap: Went with your suggestion.

packages/vue/src/composables/useStoryblokRichText.ts
dipankarmaikap: Fiexed

packages/react/src/core/richtext-hoc.tsx
dipankarmaikap: Fixed

packages/react/src/core/richtext-hoc.tsx
dipankarmaikap: Fixed

packages/richtext/src/render-richtext.ts
dipankarmaikap: Improved the type here

packages/astro/src/components/StoryblokRichText.astro
dipankarmaikap: As we support recursive rendering, users can technically pass any `SbRichTextNode`, not just an `SbRichTextDoc`. I've extracted the union type `SbRichTextNode | SbRichTextNode[] | null | undefined` into a dedicated `SbRichTextInput` type to make the API clearer and improve readability throughout the codebase.

packages/react/src/core/richtext-hoc.tsx
dipankarmaikap: You're right, it's not technically a hook. The naming already exists though, so I think we can keep it for now. I've simplified the implementation quite a bit, and it should be much easier to read and follow.

That said, you're correct that we could remove it entirely. The same applies to Vue as well. We should decide whether that's something we want to do, because we already provide components that support full customization, and under the hood everything is powered by the `createRichTextRenderer` factory function. We could simply export that for users who prefer a function-based API.

From a technical perspective, neither `useStoryblokRichText` nor its Vue counterpart are really hooks/composables anymore. They're essentially thin wrappers around the renderer factory.

Regarding the `render*` prefix, it originally came from the static `renderRichText` function. I kept the same naming pattern so it's easier to map the new implementation back to the existing API and concepts.

These functions are internal implementation details, so end users won't interact with them directly. That said, I'm not strongly attached to the naming and we can remove or simplify the prefix if we decide it's unnecessary.

packages/react/src/storyblok-rich-text.tsx
dipankarmaikap: I've simplified this quite a bit. Both `StoryblokRichText` and `StoryblokServerRichText` are now much cleaner, with less duplication and a flow that's easier to read and follow.

packages/svelte/src/lib/richtext/RenderTable.svelte
dipankarmaikap: This is intentional. Otherwise, a space gets inserted between the tags. While that's not an issue from an HTML rendering perspective, it causes the test to fail.

packages/angular/src/lib/richtext/rich-text.component.ts
dipankarmaikap: This is true. Do you have a better name in mind that we could use consistently throughout the codebase?

Because we support recursive rendering, `SbRichTextDoc` is ultimately just another node type, and users can technically pass any node and have it render correctly. The original assumption was that users would primarily pass the rich text document they receive from Storyblok, which is why the naming leaned toward "document".

packages/angular/src/lib/types.ts
dipankarmaikap: This is mainly to follow the pattern used across the other SDKs. Technically, we already export `SbRichTextProps` as part of `renderRichText`, so that type is available as well.

I don't have a strong preference here, but I wanted to keep things consistent where possible. Angular can use the standard `SbRichTextProps`, but Vue and React require framework-specific variations, so a shared `SbRichTextProps` type doesn't work in the same way across all SDKs.

packages/astro/src/components/richtext/RenderElement.astro
dipankarmaikap: It's going to add an extra slash, resulting in `<br />`. That's perfectly valid HTML and doesn't affect rendering, but it causes the unified test to fail because the output no longer matches exactly.

packages/astro/src/components/richtext/RenderTextNodeWithMarks.astro
dipankarmaikap: Yes, as this could only happen in text node.

packages/astro/src/components/richtext/RenderElement.astro
dipankarmaikap: I think there is no need to limit the type here.

packages/astro/src/components/richtext/RenderStaticChildren.astro
dipankarmaikap: I think we should keep this as is for now. I've already improved the type so it excludes text nodes while remaining open to all other node types.

I agree that the approach you're suggesting would be more type-safe, and it would likely be easier to extend in the future. However, since we don't currently have any other nested node types that would benefit from those stricter constraints.

If we introduce more node types in the future, we can revisit this and make the typing more explicit and type-safe at that point.

packages/astro/src/components/richtext/RenderStaticStructure.astro
dipankarmaikap: This works correctly for the current use case. For example:

```ts
specs = [
  {
    tag: 'strong',
    children: [
      {
        tag: 'em',
      },
    ],
  },
];
```

renders as:

```html
<strong>
  <em>
    <slot />
  </em>
</strong>
```

That said, you do raise a valid point about whether we should have a more explicit structure for certain cases instead of relying on the recursive spec tree.

For now, I'd prefer to wait and see whether we introduce any future tags or rendering scenarios that would benefit from a different approach. At the moment, this works out of the box and handles the existing requirements well.

I agree that it's not the easiest implementation to read and it's not as type-safe as it could be. If we end up expanding the set of supported tags or adding more complex nesting rules, that would be a good opportunity to revisit the design and make it more explicit.

dipankarmaikap:
Hey @maoberlehner ,

I've gone through your comments and addressed the obvious issues, as well as refactored and improved things where possible. There are still a few open points where I'd appreciate your input.

**Render* prefix**

I kept the `Render*` prefix for both the static implementation and the framework-specific implementations so they all follow the same naming pattern, regardless of whether they're functions or components. As you pointed out, components already imply rendering, which is a fair point. My reasoning was mainly consistency across SDKs, making it easier to compare implementations and apply changes across frameworks in the future. I'm happy to go with a different naming convention if you feel strongly about it, especially since these are internal APIs and not something end users interact with.

**RichText document naming**

As you also pointed out, the `RichTextComponent` currently takes a `document`/`sbDocument` prop. However, we actually support rendering any rich text node, not just a document node. Do you have any suggestions for a better name that would reflect this more accurately?

**`useStoryblokRichText`**

Another good point you raised is that `useStoryblokRichText` in React and Vue isn't really a hook/composable anymore. Under the hood, it simply returns the same renderer created by `createRichTextRenderer`, which is what `StoryblokRichText` uses internally as well.

Given that, do you think we should consider deprecating `useStoryblokRichText` and encourage users to use `createRichTextRenderer` directly if needed? Or should we keep it as a convenience API for backwards compatibility?

packages/react/src/server/richtext.tsx
```diff
@@ -1,6 +1,14 @@
+import { createDefaultBlok } from '@/core/create-default-blok';
+import { createRichTextRenderer } from '../core/richtext';
+import type { SbReactRichTextRenderContext } from '../core/richtext';
 import StoryblokServerComponent from './server-component';
-import { createRichTextHook } from '../core/richtext-hoc';
 
-export const useStoryblokServerRichText = createRichTextHook(StoryblokServerComponent, {
-  isServerContext: true,
-});
+export const useStoryblokServerRichText = (props: SbReactRichTextRenderContext = {}) => {
```
maoberlehner, line 6: As far as I can see, this is no longer exported from the package.

packages/vue/src/components/StoryblokRichText.vue
```diff
@@ -1,27 +1,36 @@
...
-  immediate: true,
-  deep: true,
+  setup(props) {
+    const render = createRichTextRenderer({
+      optimizeImage: props.optimizeImage,
+      components: {
+        blok: BlokRenderer,
+        ...props.components,
+      },
+    });
```
maoberlehner, line 32: My recommendation was not quite right. We need to wrap in computed to make the render function update on props change.

```ts
const renderer = computed(() =>
    createRichTextRenderer({
      optimizeImage: props.optimizeImage,
      components: {
        blok: BlokRenderer,
        ...props.components,
      },
    }),
  );
```

This might warrant the composable, although I think it's probably the better call to leave it inline, just with added computed. Sorry for not thinking about this thoroughly enough to catch my error earlier.

packages/react/src/storyblok-rich-text.tsx
```diff
@@ -1,24 +1,4 @@
-import React, { forwardRef } from 'react';
+import StoryblokComponent from './core/storyblok-component';
+import { createStoryblokRichText } from './core/create-storyblok-richtext';
 
-import { convertAttributesInElement } from './utils';
```
maoberlehner, line 3: We don't use this anymore internally. Good to leave the implementation in the code for backward compatibility but we might consider it for deprecation and eventual removal.

packages/vue/src/composables/useStoryblokRichText.ts
maoberlehner: Much better now, appreciate the effort, thanks! 🙏

packages/angular/src/lib/richtext/rich-text.component.ts
maoberlehner: Let's stick to `sbDocument` good enough for now.

packages/angular/src/lib/types.ts
maoberlehner: Let's keep it that way! Thanks for considering it.

packages/astro/src/components/richtext/RenderStaticStructure.astro
maoberlehner: Sensible approach, maybe we can warn on `specs.length > 1`?

packages/astro/src/components/richtext/RenderElement.astro
maoberlehner: Took a closer look and realized `RenderElement` does indeed render all types of nodes.

packages/react/src/core/richtext-hoc.tsx
maoberlehner: I think your changes were already a great improvement in the right direction! Let's keep it at that.

packages/svelte/src/lib/richtext/RenderTable.svelte
maoberlehner: Ahh, makes sense. Should we add a comment like `<!-- whitespace-sensitive: do not reformat -->` so further reviewers don't trip over it again?

maoberlehner (changes requested):
Some pretty amazing improvements! Especially given the backward compatibility constraints you're facing!

> Render prefix*
>
> I kept the Render* prefix for both the static implementation and the framework-specific implementations so they all follow the same naming pattern, regardless of whether they're functions or components. As you pointed out, components already imply rendering, which is a fair point. My reasoning was mainly consistency across SDKs, making it easier to compare implementations and apply changes across frameworks in the future. I'm happy to go with a different naming convention if you feel strongly about it, especially since these are internal APIs and not something end users interact with.

Leave it like it is.

> RichText document naming
>
> As you also pointed out, the RichTextComponent currently takes a document/sbDocument prop. However, we actually support rendering any rich text node, not just a document node. Do you have any suggestions for a better name that would reflect this more accurately?

Let's stick to it for now.

> useStoryblokRichText
>
> Another good point you raised is that useStoryblokRichText in React and Vue isn't really a hook/composable anymore. Under the hood, it simply returns the same renderer created by createRichTextRenderer, which is what StoryblokRichText uses internally as well.
>
> Given that, do you think we should consider deprecating useStoryblokRichText and encourage users to use createRichTextRenderer directly if needed? Or should we keep it as a convenience API for backwards compatibility?

We definitely should leave it for now for backward compatibility. A deprecation comment might be a good idea, I leave that to your judgment. Maybe, after this PR is merged, you could write a little draft with ideas/considerations and a roadmap for the next major release so we can discuss it as a team? This is an area where we as a team in general don't have a good practice defined for us on how to plan a major release. Could be a good opportunity to shape something.

Big PR with great improvements, many of our users are looking forward to! 🎉

packages/vue/src/components/StoryblokRichText.vue
dipankarmaikap: I'm not sure if this is correct; it's working without `computed`.

packages/astro/src/components/richtext/RenderStaticStructure.astro
dipankarmaikap: I think we don’t need a warning here since this case should never occur and we fully control the render map. I’ve instead added the warning at the point where the render map is generated.

packages/react/src/server/richtext.tsx
dipankarmaikap: Thank you for pointing that out. Fixed

maoberlehner (approved):

PR #625, opened 2026-05-28
feat(live-preview): source Story / Block / BlockContent from openapi-codegen

maoberlehner:
- Add `tools/openapi-codegen/`: internal, unpublished workspace package that owns OpenAPI spec fetching and emits sliced, aliased `_internal.gen.ts` per source spec via a single `generate({ outDir, include, sdk })` API. Specs are pinned by SHA in `spec.lock` and git-ignored; consumers commit generated output.
- Migrate `@storyblok/live-preview` to source `Story`, `Block`, `BlockContent` from `@storyblok/openapi-codegen` instead of `storyblok-js-client`'s `ISbStoryData` / `ISbComponentType`. Adds `LivePreviewStory<TStory extends Story>` for the Visual Editor `input` payload: `id`/`uuid`/`content` are sourced from the supplied `Story` generic (`id` required, `uuid`/`content` optional), every other field is left as `unknown` rather than over-promising a fully typed CDN story; `onStoryblokEditorEvent` is generic over the `Story`, not just its content.
- This *technically* breaks the type-only public surface, but is intentionally shipped as a **minor**, not a breaking release: the package is pre-1.0 (`0.x`), whose API is not yet considered stable under SemVer, and the change is type-only — runtime behavior is unchanged.
- Scope is intentionally limited to the live-preview canary. `capi-client`, `mapi-client`, `schema`, and the legacy `packages/openapi/` are untouched and will migrate in follow-up PRs.

See WDX-406
See WDX-430

## Test plan

- [x] `pnpm nx run-many -t build typecheck lint test --projects=@storyblok/live-preview,@storyblok/openapi-codegen`
- [x] Consumer apps: type-check still passes against `@storyblok/live-preview`'s new generic surface (`Story`, `Block`, `BlockContent`, `LivePreviewStory`).
- [x] Manually re-run `pnpm nx generate @storyblok/live-preview` and confirm `src/generated/` output is byte-stable.


tools/openapi-codegen/specs/mapi/components/field-types/asset-field.yaml
```diff
@@ -0,0 +1,37 @@
+allOf:
```
maoberlehner, line 1: We need those because the official OpenAPI spec does not contain field specs, and they do not plan to add them. We might consider moving this to the official overlay repo, but for now, it is easier to leave it here.

Those field specs are now duplicated and also exist in the openapi package. We can't remove the openapi package with this PR because we need it for the clients still, and updating the clients happens in the alpha branch instead.

packages/live-preview/src/onStoryblokEditorEvent.ts
```diff
@@ -98,16 +98,14 @@ async function initializeBridge(bridgeOptions?: BridgeParams): Promise<void> {
  * cleanup()
  * ```
  */
-export async function onStoryblokEditorEvent<
-  T extends ISbComponentType<string> = ISbComponentType<string>,
->(
-  callback: (story: ISbStoryData<T>) => void,
+export async function onStoryblokEditorEvent<TBlock extends StoryBlock = StoryBlock>(
```
maoberlehner: Wrong type, see: https://github.com/storyblok/monoblok/pull/614#discussion_r3302534898

tools/openapi-codegen/scripts/known-types.ts
```diff
@@ -0,0 +1,157 @@
...
+  capi: resolve(CACHE_DIR, 'capi/cdn-v2.openapi.yaml'),
+  mapi: resolve(CACHE_DIR, 'mapi/management-v1.openapi.yaml'),
+  overlay: OVERLAY_PATH,
+};
+
+/**
+ * Extracts top-level schema names from `components.schemas` in an OpenAPI
+ * YAML. Relies on the upstream specs using 2-space indentation.
+ */
+export function listSchemaNames(yamlPath: string): string[] {
```
dipankarmaikap: non-blocker: Although this works why not be safe and use something like `yaml` package to be 100% safe and dont rely on 2 space indentation.

tools/openapi-codegen/scripts/fetch.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (!existsSync(src)) {
+      throw new Error(`Expected ${surface} spec at ${src} — upstream layout has changed.`);
+    }
+    const dest = resolve(CACHE_DIR, surface);
+    mkdirSync(dest, { recursive: true });
+    copyFileSync(src, resolve(dest, file));
+  }
+}
+
+function main(): void {
```
dipankarmaikap, line 83: non-blocker: Should we skip fetching when everything is already up to date?

dipankarmaikap (approved):
I left two minor, non-blocking comments.

Really like this approach. A single generated file, package-local types, and no type dependencies between packages make the setup much cleaner and easier to maintain.

tools/openapi-codegen/README.md
```diff
@@ -0,0 +1,55 @@
...
+# Verify the cache on disk matches spec.lock.
+pnpm --filter @storyblok/openapi-codegen verify
+```
+
+Both `pull` commands require `gh auth status`. Specs live in the private `storyblok/openapi-wdx` repo.
+
+After a `pull:update`, regenerate consumers and commit the diff:
+
+```sh
+pnpm nx run-many -t generate --projects=@storyblok/api-client,@storyblok/management-api-client,@storyblok/schema
```
dipankarmaikap, line 23: I think we should also document how to generate types for a single package, for example:

`pnpm nx run @storyblok/live-preview:generate`

That can be useful when working on an individual package and avoids running generation across the entire workspace.

tools/openapi-codegen/src/aliases.ts
```diff
@@ -0,0 +1,87 @@
+/**
```
alexjoverm, line 1: suggestion (blocking): the openapi-codegen tool is a foundational piece to make all this work, however doesn't have any tests ATM. So if any regression is introduced, it'll have a direct impact in most packages. 

I'd suggest we add:
- some unit test only on critical core logic parts
- mostly input/output integration tests, or snapshots, to cover a great deal of the functionality in a mantainable way

tools/openapi-codegen/src/index.ts
```diff
@@ -0,0 +1,282 @@
...
+ * does not apply here (clients consume raw schema/operation names directly).
+ */
+async function emitFullSdk(spec: 'capi' | 'mapi', outDir: string, verbose: boolean): Promise<void> {
+  const sdkDir = resolve(outDir, spec);
+  await createClient({
+    input: SPEC_PATHS[spec],
+    output: sdkDir,
+    plugins: [
+      '@hey-api/typescript',
+      '@hey-api/client-ky',
```
alexjoverm, line 117: @maoberlehner do we need to declare them in `tools/package.json` or somewhere as deps? I haven't seen them listed

tools/openapi-codegen/scripts/known-types.ts
alexjoverm: +1. Would make is less fragile. All YAML parsing here is done by regex, would be better to use a yaml parser

packages/live-preview/package.json
```diff
@@ -38,21 +38,38 @@
...
+  "nx": {
+    "targets": {
+      "generate": {
+        "inputs": [
+          "{projectRoot}/scripts/generate.ts"
+        ],
+        "outputs": [
+          "{projectRoot}/src/generated"
+        ],
+        "cache": true,
```
alexjoverm, line 58: From Claude:

**Problem:** The `generate` target's `inputs` only lists `{projectRoot}/scripts/generate.ts`. The `dependsOn: ["@storyblok/openapi-codegen:build"]` ensures the codegen tool is built first, but nx input hashing does not transitively include the codegen tool's build inputs (specs, templates, aliases). If the codegen tool's build is already cached (same compiled JS output), but a template file changes in a way that produces different generated output without changing the compiled `dist/`, the consumer's generate target would incorrectly hit cache.

**Suggestion:** Add codegen tool inputs to the consumer's generate target:
```json
"inputs": [
  "{projectRoot}/scripts/generate.ts",
  "{workspaceRoot}/tools/openapi-codegen/spec.lock",
  "{workspaceRoot}/tools/openapi-codegen/templates/**",
  "{workspaceRoot}/tools/openapi-codegen/specs/**"
]
```

alexjoverm (changes requested):
@maoberlehner the architecture and approach seems pretty solid to me. Added a few comments for some missing things before merging

tools/openapi-codegen/templates/field.ts
```diff
@@ -0,0 +1,118 @@
...
+
+type OptionalFieldKeys<T> = Exclude<keyof T, RequiredFieldKeys<T>>;
+
+/**
+ * Content object for a single block instance as returned by the Storyblok
+ * Content Delivery API. Without a `TBlock` argument, this is the loose
+ * runtime shape (any block, `_editable` optional). With a schema-typed
+ * `TBlock`, fields are narrowed per the block's schema.
+ */
+export type BlockContent<TBlock extends Block = Block, TBlocks = false> =
```
alexjoverm: question (non-blocking): I feel some types are a bit harder to read, so I wonder - do you see opportunities to make types within `templates/**.ts` more human readable and simpler?

tools/openapi-codegen/templates/field.ts
```diff
@@ -0,0 +1,118 @@
...
+  PluginFieldValue,
+  RichtextFieldValue,
+  TableFieldValue,
+} from './_sources';
+import type { Block } from './block';
+
+export type { Field };
+export type { AssetFieldValue, MultilinkFieldValue, PluginFieldValue, RichtextFieldValue, TableFieldValue };
+
+type Prettify<T> = { [K in keyof T]: T[K] } & {};
```
alexjoverm: (non-blocking): `Prettify<T>` duplicated in a few files

**Files:**
- `packages/live-preview/src/onStoryblokEditorEvent.ts:7`
- `tools/openapi-codegen/templates/block.ts:10`
- `tools/openapi-codegen/templates/field.ts:16`
- `tools/openapi-codegen/templates/story.ts:5`
- `tools/openapi-codegen/templates/mapi-story.ts:9`

**Problem:** `type Prettify<T> = { [K in keyof T]: T[K] } & {};` is defined identically in every template file and also in `onStoryblokEditorEvent.ts`. While templates are copied to each consumer (so some duplication is inherent), `onStoryblokEditorEvent.ts` could import it from the generated types rather than redefining it. For the templates, consider extracting it to a shared `_utils.ts` template that all others import from.

tools/openapi-codegen/src/known-types.ts
```diff
@@ -0,0 +1,356 @@
...
+    'Link',
+    'LinkWithFullSlug',
+    'PublicSpace',
+    'PublishedStory',
+    'RateLimitError',
+    'Story',
+    'StoryAlternate',
+    'Tag',
+    'UnauthorizedError',
+  ] as const,
```
alexjoverm, line 37: (non-blocking): I see many `as const` along the file (maybe in others). Are they necessary? Is there cleaner way?

tools/openapi-codegen/src/paths.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+import { existsSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { dirname, resolve } from 'pathe';
+
+function findSourceToolRoot(): string {
+  const moduleDir = dirname(fileURLToPath(import.meta.url));
+  let dir = moduleDir;
+  for (;;) {
```
alexjoverm: nitpick (non-blocking): I find `for(;;)` unnecessarily convoluted. Any more human-readable way to declare the loop?

tools/openapi-codegen/src/transform.ts
```diff
@@ -0,0 +1,229 @@
...
+
+function extractProperty(typeNode: ts.TypeNode, propName: string): ts.TypeNode | undefined {
+  if (ts.isTypeLiteralNode(typeNode)) {
+    for (const member of typeNode.members) {
+      if (
+        ts.isPropertySignature(member)
+        && member.name
+        && ts.isIdentifier(member.name)
+        && member.name.text === propName
+        && member.type
```
alexjoverm: nitpick (non-blocking): it'd be great to structure better conditions like this

tools/openapi-codegen/src/known-types.ts
maoberlehner: As far as I remember, I tried to remove it, but it is needed here to make the `generate()` function not fall back to strings. Although I try to avoid `as const` too if they don't add value, there is also no downside of using them AFAIK (opposed to `as SomeType`). But will double-check when addressing the other comments!

tools/openapi-codegen/templates/field.ts
maoberlehner: I don't disagree with the sentiment 😬 Yet, I think I've spent 90% of the time on writing those types and getting them as readable as I can (as Claude Code wasn't of great use there at all). There is probably a way to improve them further, but it's outside the realm of my expertise / would require another significant time investment.

I fear that clean types for our users require complex types on our end.

tools/openapi-codegen/scripts/known-types.ts
maoberlehner: +1 😅
https://github.com/storyblok/monoblok/pull/639/changes what do you think?

tools/openapi-codegen/src/aliases.ts
maoberlehner: The problem with input/output tests is that we can't run them in CI because we can't download the OpenAPI spec in CI (without exposing it).

I'll add unit tests!

tools/openapi-codegen/src/index.ts
maoberlehner: @hey-api/typescript/@hey-api/sdk/@hey-api/client-ky are built-in plugin identifiers bundled in @hey-api/openapi-ts (See e.g.: https://www.npmjs.com/package/@hey-api/typescript does not exist).

packages/live-preview/package.json
maoberlehner: Fixed (but different solution).

tools/openapi-codegen/scripts/fetch.ts
maoberlehner: done!

tools/openapi-codegen/README.md
maoberlehner: done!

tools/openapi-codegen/templates/field.ts
maoberlehner: Made some minor updates to hopefully make them somewhat more readable.

tools/openapi-codegen/templates/field.ts
maoberlehner: Extracted!

tools/openapi-codegen/src/paths.ts
maoberlehner: done!

tools/openapi-codegen/src/transform.ts
maoberlehner: Done!

dipankarmaikap (approved):
LGTM 🚀 Nice improvements and great to see the added tests.

alexjoverm (approved):

PR #626, opened 2026-05-28
feat(openapi): replace @storyblok/openapi with specs generated from the official OpenAPI spec

maoberlehner:
Replaces the in-repo `@storyblok/openapi` spec with on-demand generation from the private `storyblok/openapi-wdx`, regenerates the MAPI/CAPI clients and `@storyblok/schema` types from it, and decouples the generated type sources so consumers no longer pull unrelated types.

Instead of packages depending on one single source of truth types package, that changes (and bumps all packages depending on it) whenever there is a slight change to the OpenAPI spec or the resulting types, we let all packages generate their own types and let them specify exactly which types they need.

Yet they still depend on a single-source-of-truth-type `@storyblok/openapi-codegen` package. So why is this different? We don't need to publish and version this package at all. So changing it does not necessarily bump the packages that depend on it. The `@storyblok/openapi-codegen` package only provides the tooling and standardization that allows other packages to create just the types they need. Only if _their_ types change must we update the version of a package.

Closes WDX-406
Closes WDX-430

---

## WDX-406 — Remove @storyblok/openapi and replace with overlay

- **`@storyblok/openapi` removed from the monorepo.** No package remains; some schemas were reused in `tools/openapi-codegen/specs/overlay.openapi.yaml`.
- **MAPI & CAPI clients generated from `openapi-wdx`** via `@storyblok/openapi-codegen` (flat SDK emit).
- **`@storyblok/schema` types generated from `openapi-wdx`**.
- **Specs pulled on demand, not committed.** Pinned by SHA in `spec.lock`; `fetch`/`pull` require `gh auth` against the private repo. `.openapi-cache/` is git-ignored.
- **External-contributor implications documented** in `tools/openapi-codegen/README.md`: contributors without `openapi-wdx` access can still `pnpm install && pnpm build` (consumers commit their `src/generated/` output); spec changes surface as reviewable diffs only when type-relevant.

## WDX-430 — Break out the types dependencies

Reduces cross-type coupling that triggered unnecessary dependent releases:

- Consumers can use `generate({ include: ['Story', '...'] })` to generate a **minimal set of types it actually needs**, instead of the full generated surface — so an unrelated spec change no longer ripples a release through every dependant.


packages/capi-client/src/resources/datasource-entries.ts
```diff
@@ -1,8 +1,8 @@
-import { list as listDatasourceEntriesApi } from '../generated/datasource_entries/sdk.gen';
+import { listDatasourceEntries } from '../generated/capi/sdk.gen';
 import type {
-  ListData as DatasourceEntriesListData,
-  ListResponses as DatasourceEntriesListResponses,
-} from '../generated/datasource_entries/types.gen';
+  ListDatasourceEntriesData as DatasourceEntriesListData,
+  ListDatasourceEntriesResponses as DatasourceEntriesListResponses,
```
maoberlehner: Why `as` renaming?

packages/capi-client/src/resources/datasources.ts
```diff
@@ -1,10 +1,10 @@
...
 import type {
-  GetData as DatasourcesGetData,
-  GetResponses as DatasourcesGetResponses,
-  ListData as DatasourcesListData,
-  ListResponses as DatasourcesListResponses,
-} from '../generated/datasources/types.gen';
+  GetDatasourceByIdData as DatasourcesGetData,
+  GetDatasourceByIdResponses as DatasourcesGetResponses,
+  ListDatasourcesData as DatasourcesListData,
+  ListDatasourcesResponses as DatasourcesListResponses,
```
maoberlehner: Why `as` renaming?

packages/capi-client/src/resources/links.ts
```diff
@@ -1,5 +1,5 @@
-import { list as listLinks } from '../generated/links/sdk.gen';
-import type { ListData as LinksListData, ListResponses as LinksListResponses } from '../generated/links/types.gen';
+import { listLinks } from '../generated/capi/sdk.gen';
+import type { ListLinksData as LinksListData, ListLinksResponses as LinksListResponses } from '../generated/capi/types.gen';
```
maoberlehner: Why `as` renaming?

packages/capi-client/src/resources/spaces.ts
```diff
@@ -1,24 +1,21 @@
-import { get as getSpaceApi } from '../generated/spaces/sdk.gen';
-import type { GetResponses as SpacesGetResponses } from '../generated/spaces/types.gen';
+import { getSpace as getSpaceApi } from '../generated/capi/sdk.gen';
+import type { GetSpaceData as SpacesGetData, GetSpaceResponses as SpacesGetResponses } from '../generated/capi/types.gen';
```
maoberlehner: Why `as` renaming?

packages/capi-client/src/resources/stories.test.ts
```diff
@@ -49,6 +49,10 @@ const makeStory = (uuid: string, content: Record<string, unknown>) => {
 
 describe('stories.get()', () => {
   it('should successfully retrieve a single story', async () => {
+    server.use(
+      http.get('https://api.storyblok.com/v2/cdn/stories/*', () =>
+        HttpResponse.json({ story: makeStory('test-story', { component: 'page', _uid: 'uid-1' }) })),
+    );
```
maoberlehner, line 45: Why was this not necessary before?

packages/capi-client/src/resources/stories.ts
```diff
@@ -1,17 +1,17 @@
-import { get, list } from '../generated/stories/sdk.gen';
-import type { GetData, GetResponses, ListData, ListResponses } from '../generated/stories/types.gen';
+import { getStoryById, listStories } from '../generated/capi/sdk.gen';
+import type { GetStoryByIdData as GetData, GetStoryByIdResponses as GetResponses, ListStoriesData as ListData, ListStoriesResponses as ListResponses } from '../generated/capi/types.gen';
```
maoberlehner: Why `as` renaming?

packages/capi-client/src/resources/stories.ts
```diff
@@ -120,8 +120,21 @@ type ListResponse<
...
+  story: Story;
+}
+
+interface StoriesData extends StoryRelationData {
+  stories: Story[];
+}
+
 /** Pre-resolved to avoid TypeScript emitting deep indexed-access chains that trip up DTS bundlers. */
-type StoryIdentifier = GetData['path']['identifier'];
+type StoryIdentifier = GetData['path']['id'];
```
maoberlehner: Why this change?

packages/capi-client/src/resources/stories.ts
```diff
@@ -120,8 +120,21 @@ type ListResponse<
...
+
+interface StoryData extends StoryRelationData {
+  story: Story;
+}
+
+interface StoriesData extends StoryRelationData {
+  stories: Story[];
+}
+
 /** Pre-resolved to avoid TypeScript emitting deep indexed-access chains that trip up DTS bundlers. */
```
maoberlehner: Check if this command is still true.

packages/capi-client/src/resources/stories.ts
```diff
@@ -120,8 +120,21 @@ type ListResponse<
   stories: Array<StoryResult<TComponents, InlineRelations, ResolveRelationsRaw>>;
 };
 
+interface StoryRelationData {
```
maoberlehner, line 130: Add comment explaining why we need those three types here.

packages/capi-client/src/resources/stories.ts
```diff
@@ -167,7 +180,8 @@ export function createStoriesResource<
           return response;
         }
 
-        const resolved = await resolveRelationMap(response.data, requestQuery, { client, throttleManager });
+        const storyData = response.data as unknown as StoryData;
```
maoberlehner, line 192: Add brief comment to this and similar cast below explaining why we need to cast (nice generic types for consumers).

packages/capi-client/src/resources/tags.ts
```diff
@@ -1,5 +1,5 @@
-import { list as listTagsApi } from '../generated/tags/sdk.gen';
-import type { ListData as TagsListData, ListResponses as TagsListResponses } from '../generated/tags/types.gen';
+import { listTags as listTagsApi } from '../generated/capi/sdk.gen';
+import type { ListTagsData as TagsListData, ListTagsResponses as TagsListResponses } from '../generated/capi/types.gen';
```
maoberlehner: Why `as` renaming?

packages/capi-client/src/utils/fetch-rel-uuids.ts
```diff
@@ -64,7 +64,10 @@ export const fetchMissingRelations = async ({
           throw new Error('Failed to fetch missing relations.');
         }
 
-        return response.data.stories;
+        // SDK return shape is the raw `DraftStory | PublishedStory` union; cast
+        // to the wrapper `Story` (public type) so callers don't have to know
+        // about the spec-level discriminated form.
```
maoberlehner: ... and get the benefits of the generic `Story` with schema support.

packages/live-preview/package.json
```diff
@@ -38,21 +38,38 @@
...
   "devDependencies": {
     "@storyblok/eslint-config": "workspace:*",
+    "@storyblok/openapi-codegen": "workspace:*",
     "@types/node": "^24.11.0",
     "bumpp": "^10.4.1",
     "eslint": "^9.39.2",
-    "storyblok-js-client": "workspace:*",
     "tsdown": "^0.20.3",
     "typescript": "^5.9.3",
     "vitest": "^4.0.18"
```
maoberlehner, line 76: Undo all live-preview changes. Best dropping the commit(s) if they have only changes in here.

packages/mapi-client/playground/integration-tests/test/types/components.test-d.ts
```diff
@@ -1,9 +1,6 @@
-import { describe, expectTypeOf, it } from 'vitest';
 import { defineBlock, defineBlockCreate, defineBlockUpdate, defineField } from '@storyblok/schema';
-import { createManagementApiClient } from '../index';
-import type { Component as ComponentMapi } from '../generated/components/types.gen';
-
-// ─── Component definitions using defineX helpers ─────────────────────────────
+import { type Component as ComponentMapi, createManagementApiClient } from '@storyblok/management-api-client';
```
maoberlehner, line 2: Why `as` renaming?

packages/mapi-client/playground/integration-tests/test/types/stories.test-d.ts
```diff
@@ -17,11 +16,9 @@ const teaserComponent = defineBlock({
 const heroComponent = defineBlock({
   name: 'hero',
   is_root: true,
-  // is_nestable defaults to true — hero can appear both as a story and in bloks
   schema: [
     defineField('title', { type: 'text' }),
     defineField('count', { type: 'number' }),
-    // bloks field without a whitelist — resolves to nestable components (hero + teaser; page excluded)
```
maoberlehner, line 24: comments like these were useful, don't you think?

packages/mapi-client/playground/integration-tests/test/types/stories.test-d.ts
```diff
@@ -1,7 +1,6 @@
-import { describe, expectTypeOf, it } from 'vitest';
 import { defineBlock, defineField, defineStoryCreate, defineStoryUpdate } from '@storyblok/schema';
-import { createManagementApiClient } from '../index';
-import type { StoryMapi } from '../generated/stories/types.gen';
+import { createManagementApiClient, type Story as StoryMapi } from '@storyblok/management-api-client';
```
maoberlehner, line 2: Why `as` renaming?

packages/mapi-client/scripts/generate.ts
```diff
@@ -1,111 +1,50 @@
...
-main().catch((error) => {
-  console.error('Generation failed:', error);
-  process.exit(1);
+const packageRoot = resolve(dirname(fileURLToPath(import.meta.url)), '..');
+
+await generate({
+  outDir: resolve(packageRoot, 'src/generated'),
+  sdk: 'mapi',
+  include: [
+    // Wrapper templates (transitively pull block, field, story, mapi-story).
```
maoberlehner: IMO useless distinction in here between wrapper and aliased (also relevant in capi generate script).

packages/mapi-client/src/resources/assets.ts
```diff
@@ -1,65 +1,61 @@
...
-  FinalizeResponses,
-  GetResponses,
-  ListData,
-  ListResponses,
-  SignResponseObjectData,
-  SignResponseObjectResponses,
-  UpdateData,
-  UpdateResponses,
-} from '../generated/assets/types.gen';
+  ShowAsset as Asset,
```
maoberlehner: ShowAsset should not be exposed by the generator. Renaming should already happen in the codegen generate script.

packages/mapi-client/src/resources/assets.ts
```diff
@@ -1,65 +1,61 @@
...
+ * Return shape of `list()`.
+ *
+ * The list endpoint serializes rows with MAPI's `IndexAsset` serializer, which
+ * is NOT identical to the `ShowAsset` (= `Asset`) shape returned by `get()`:
+ * `IndexAsset` omits `file` and `permanently_deleted`. We deliberately surface
+ * list rows as `Asset` so consumers work against a single asset type. The lie:
+ * `file`/`permanently_deleted` are absent on list rows at runtime — consumers
+ * needing those must re-fetch via `get()`.
+ */
+export type ListAssetsResult = Omit<ListAssetsResponses[200], 'assets'> & { assets: Array<Asset> };
```
maoberlehner, line 37: IndexAsset should not be exposed by the codegenerator generate script but this should be already overriden at codegen.

packages/mapi-client/src/resources/assets.ts
```diff
@@ -1,65 +1,61 @@
...
+
+/**
+ * Fields for initiating an asset upload. Pass this to `upload()` or `create()`
+ * alongside a file buffer.
  *
  * Uses `short_filename` (e.g. `"hero.png"`) to clearly distinguish it from
  * `Asset.filename`, which is the full CDN URL assigned by Storyblok after upload.
```
maoberlehner: remove this part about short_filename comment here

packages/mapi-client/src/resources/datasource-entries.test.ts
```diff
@@ -21,6 +21,11 @@ afterAll(() => server.close());
 
 describe('datasourceEntries.list()', () => {
   it('should successfully retrieve datasource entries', async () => {
+    server.use(
+      http.get('https://mapi.storyblok.com/v1/spaces/:space_id/datasource_entries', () => {
+        return HttpResponse.json({ datasource_entries: [] });
+      }),
+    );
```
maoberlehner, line 18: Why was this not necessary before?

packages/mapi-client/src/resources/datasource-entries.ts
```diff
@@ -15,43 +18,65 @@ export function createDatasourceEntriesResource<DefaultThrowOnError extends bool
...
       datasourceEntryId: number,
-      options: { body: UpdateData['body']; signal?: AbortSignal; throwOnError?: ThrowOnError; fetchOptions?: FetchOptions } & SpaceIdPathOverride,
-    ): Promise<ApiResponse<void, ThrowOnError>> {
+      options: { body: PartialUpdateDatasourceEntryData['body']; signal?: AbortSignal; throwOnError?: ThrowOnError; fetchOptions?: FetchOptions } & SpaceIdPathOverride,
+    ): Promise<ApiResponse<PartialUpdateDatasourceEntryResponses[204], ThrowOnError>> {
       const { body, signal, path, throwOnError, fetchOptions } = options;
       const resolvedSpaceId = getSpaceId(path);
-      return wrapRequest<void, ThrowOnError>(() =>
-        datasourceEntriesApi.update({
+      return wrapRequest<PartialUpdateDatasourceEntryResponses[204], ThrowOnError>(() =>
```
maoberlehner, line 56: why did this get so much more complicated?

packages/mapi-client/src/resources/stories.ts
```diff
@@ -1,82 +1,68 @@
...
+  ListVersionsData,
+  ListVersionsResponses,
+  PublishStoryData,
+  PublishStoryResponses,
+  UpdateStoryData,
+  UpdateStoryRequest,
+  UpdateStoryResponses,
+} from '../generated/mapi/types.gen';
+import type { Block as Component } from '../generated/types/block';
+import type { MapiStory, StoryCreate as MapiStoryCreate, StoryUpdate as MapiStoryUpdate } from '../generated/types/mapi-story';
```
maoberlehner: Why `as` renaming?

packages/mapi-client/src/index.ts
```diff
@@ -20,66 +20,73 @@ export type {
...
 export type {
+  Asset,
   AssetCreate,
   AssetFolder,
   AssetFolderCreate,
   AssetFolderUpdate,
   AssetUpdate,
-  BlockContentInput as BlokContentInput,
-  ComponentCreate,
+  Component as RawComponent,
```
maoberlehner: What is this? Why Raw?

packages/schema/scripts/generate.ts
```diff
@@ -0,0 +1,83 @@
...
+import { resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { generate } from '@storyblok/openapi-codegen';
+
+const PKG_ROOT = resolve(fileURLToPath(import.meta.url), '../..');
+
+await generate({
+  outDir: resolve(PKG_ROOT, 'src/generated'),
+  include: [
+    // Story + content (CAPI Story, MAPI write payloads)
```
maoberlehner: separator comments are useless here IMO

packages/capi-client/playground/playground-capi-schema/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@storyblok/playground-capi-schema",
```
maoberlehner: Rename and move "playground-{capi,mapi}-schema" to "playground-tests"

packages/schema/src/helpers/define-block.ts
```diff
@@ -1,7 +1,16 @@
...
+  ComponentCreate,
+  ComponentUpdate,
+  NestableBlocks,
+  RootBlocks,
+  SchemaArrayToRecord,
+} from '../generated/types/block';
 import type { Prettify } from '../utils/prettify';
 
-export type { ComponentCreate, ComponentUpdate };
+export type { Block, BlockSchema, BlockSchemaInput, ComponentCreate, ComponentUpdate, NestableBlocks, RootBlocks, SchemaArrayToRecord };
```
maoberlehner: remove Component variants from public export surface.

packages/schema/src/helpers/define-story.ts
```diff
@@ -48,47 +41,48 @@ const MAPI_STORY_DEFAULTS = {
...
+  user_ids: [],
+  space_role_ids: [],
+  can_not_view: null,
+  is_scheduled: null,
+  scheduled_dates: null,
+  ideas: [],
+  favourite_for_user_ids: [],
 };
 
+export type { MapiStory, Story, StoryBlock, StoryCreate, StoryUpdate };
```
maoberlehner: Unsure about StoryBlock naming. Maybe RootBlock?

packages/schema/tsconfig.json
```diff
@@ -9,6 +9,6 @@
     "isolatedModules": true,
     "skipLibCheck": true
   },
-  "include": ["**/*.ts"],
+  "include": ["src"],
```
maoberlehner: why this change? I also want ts files outside of src included!

tools/openapi-codegen/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@storyblok/openapi-codegen",
+  "type": "module",
+  "version": "0.0.1",
+  "private": true,
+  "description": "Internal: owns the OpenAPI spec cache and the shared @hey-api/openapi-ts generator used by consumer packages. Specs are git-ignored; the upstream commit SHA is pinned in spec.lock. Not published.",
```
maoberlehner: details like hey-api is useless here

tools/openapi-codegen/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+    /* Linting */
+    "strict": true,
+    "noFallthroughCasesInSwitch": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noEmit": true,
+    "isolatedModules": true,
+    "skipLibCheck": true
+  },
+  "include": ["scripts/**/*.ts", "src/**/*.ts"],
```
maoberlehner: include all ts files in this repo

packages/cli/src/commands/assets/actions.ts
```diff
@@ -210,9 +210,22 @@ export const createAsset = async (
...
+      updated_at: _updatedAt,
+      file: _file,
+      deleted_at: _deletedAt,
+      content_length: _contentLength,
+      content_type: _contentType,
+      permanently_deleted: _permanentlyDeleted,
+      locked: _locked,
+      internal_tags_list: _internalTagsList,
+      ...assetBody
+    } = asset as AssetUpload & Partial<Asset>;
```
maoberlehner: This as typecast seems like a smell

packages/cli/src/commands/assets/streams.ts
```diff
@@ -673,13 +674,15 @@ const processAsset = async ({
       fileBuffer,
     );
     // updateAsset returns void; the remote asset's readonly fields are unchanged.
-    newRemoteAsset = { ...remoteAsset, ...updatePayload };
+    // Cast: `Asset` declares `is_private: boolean` (required, no null) while the
+    // spread permits `null|undefined`; the runtime value is always a boolean.
+    newRemoteAsset = { ...remoteAsset, ...updatePayload } as Asset;
```
maoberlehner: As cast seems like smell

packages/cli/src/commands/assets/types.ts
```diff
@@ -7,12 +7,14 @@ export type { Asset, AssetCreate, AssetFolder, AssetFolderCreate, AssetFolderUpd
  * CLI extension of mapi-client's AssetUploadRequest.
  * Uses `AssetUploadRequest` (which combines `AssetUpdate` metadata fields with
  * `short_filename` for uploads) rather than the schema `AssetCreate` (which has
- * `filename`). Adds an optional `id` field used for manifest-based local↔remote
- * ID mapping. The `id` is stripped before the mapi-client create/upload call.
+ * `filename`). Adds the local asset's identity (`id` and the original `filename`)
+ * used for manifest-based local↔remote mapping; both are ignored by the
+ * mapi-client create/upload call (which derives the upload name from
+ * `short_filename`).
```
maoberlehner: This is confusing. Why is filename necessary here? Has this to do with the mapi client changing using short_filename somewhere where spec says filename?

packages/cli/src/commands/components/push/tag-ids.ts
```diff
@@ -0,0 +1,17 @@
...
+ * shape. The MAPI component serializer returns tag IDs as `string[]`, while
+ * `ComponentCreate`/`ComponentUpdate` expect `number[]` — a known upstream
+ * inconsistency (every other resource uses integer tag IDs). This is the single
+ * boundary where response tag IDs are echoed back into a create/update payload.
+ *
+ * Kept dependency-free on purpose: `actions.ts` (which imports this) is part of
+ * an import cycle via `graph-operations/dependency-graph`, so this must not pull
+ * in any module that re-enters `actions.ts`.
+ */
+export function toRequestTagIds(
```
maoberlehner: is this needed anywhere else than actions? move it in actions if not

packages/cli/src/commands/stories/ref-mapper.ts
```diff
@@ -48,7 +48,7 @@ const traverseAndMapBySchema = (
 
     if (fieldRefMapper) {
       dataNew[fieldName] = fieldRefMapper(fieldValue as any, {
-        schema: fieldSchema,
+        schema: fieldSchema as Field | undefined,
```
maoberlehner: as cast seems like a smell

packages/cli/src/types/index.ts
```diff
@@ -32,6 +33,14 @@ export type {
   User,
 } from '@storyblok/management-api-client';
 
+/**
+ * `Component` mirrors the raw MAPI response shape (loose `schema`), which is what
+ * the SDK actually returns from `client.components.list()`. The SDK's public
```
maoberlehner: We need to override this in the client. Client needs to return `Block`

packages/cli/src/commands/components/push/actions.ts
```diff
@@ -20,7 +19,7 @@ function isSchemaField(value: unknown): value is Field {
 }
 
 function toWritableSchema(
-  schema: Component['schema'] | Record<string, Field> | undefined,
+  schema: Record<string, unknown> | undefined,
```
maoberlehner, line 22: Why Component['schema'] is not enough here?

packages/cli/src/commands/stories/ref-mapper.ts
```diff
@@ -47,8 +47,8 @@ const traverseAndMapBySchema = (
     const fieldRefMapper = typeof fieldType === 'string' ? fieldRefMappers[fieldType] : undefined;
 
     if (fieldRefMapper) {
-      dataNew[fieldName] = fieldRefMapper(fieldValue as any, {
-        schema: fieldSchema as Field | undefined,
+      dataNew[fieldName] = fieldRefMapper(fieldValue as Record<string, unknown>, {
```
maoberlehner, line 50: Is this typecast necessary?

maoberlehner:
FYI: Some generated files, especially `packages/mapi-client/src/generated/mapi/types.gen.ts` are huge. That's the cost we need to pay for removing the OpenAPI YAML and therefore have to commit the generated code.

.github/workflows/ci.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            projects: tag:npm:public
+            projects: "tag:npm:public,@storyblok/mapi-integration-tests,@storyblok/capi-integration-tests"
```
alexjoverm, line 42: @maoberlehner where are these packages located? Do you think we can tag them in some way to not add specific package names here?

packages/capi-client/playground/integration-tests/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "private": true,
+  "description": "Integration tests for @storyblok/api-client paired with @storyblok/schema",
+  "scripts": {
+    "test": "vitest run",
+    "test:types": "tsc --noEmit --skipLibCheck",
+    "test:e2e": "vitest run -c vitest.config.e2e.ts",
+    "lint": "eslint .",
+    "lint:fix": "eslint . --fix"
+  },
+  "devDependencies": {
```
alexjoverm, line 14: @maoberlehner what would be the main reasoning to setup the test suite as an npm package? What are the benefits and tradeoffs of this approach vs just simply having the scripts/deps within the lib itself?

packages/capi-client/playground/integration-tests/README.md
```diff
@@ -0,0 +1,24 @@
+# @storyblok/capi-integration-tests
+
+Integration tests for [`@storyblok/api-client`](../../) paired with [`@storyblok/schema`](../../../schema).
+
+This playground exists so the api-client package itself stays free of a runtime dependency on `@storyblok/schema`, while still letting us verify that the two packages line up.
```
alexjoverm, line 5: @maoberlehner wouldn't declaring it as a devDependency accomplish the same? Or it does trigger a new release here when we publish one on `@storyblok/schema`?

.github/workflows/ci.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
       matrix:
```
alexjoverm, line 39: About _"External-contributor implications documented in tools/openapi-codegen/README.md: contributors without openapi-wdx access can still pnpm install && pnpm build (consumers commit their src/generated/ output); spec changes surface as reviewable diffs only when type-relevant."_ -> should we actually add it to `CONTRIBUTING.md`?

Also @maoberlehner as now we'd need to re-generated the clients in some cases, shall we add something to do that automatically? (like some kind of post-git-pull hook?) Also, in which cases in particular would it be needed the re-generation? would be great to have it documented in contributors

packages/capi-client/playground/integration-tests/package.json
maoberlehner: One goal of the codegen package was that minor changes to one package don't cause version bumps in a lot of other packages, too. By moving the integration tests into this separate package, we can avoid tying the `@storyblok/api-client` package to `@storyblok/management-api-client` and `@storyblok/schema`.

packages/capi-client/playground/integration-tests/README.md
maoberlehner: With our current setup, `devDependencies` do trigger a new release. Claude told me that we might take a look at `updateDependents: never` scoped to only dev dependencies. Maybe something we can take a look at in the future. For now, this seemed like the most straightforward approach for still having full e2e tests of the types.

.github/workflows/ci.yaml
maoberlehner: IMO, I would remove this filter completely. Why only test public packages?

.github/workflows/ci.yaml
maoberlehner: I don't think it is relevant to external contributors. The goal was that for them, everything just works. They probably don't need/want to update the OpenAPI-based types.

I don't think we want/need to do it automatically. Also, we can't regenerate the clients fully automatically. If new endpoints are added or old ones are changed significantly, we also need to do some manual work to apply DX niceties and the schema-inferred type magic.

packages/cli/src/commands/components/push/graph-operations/dependency-graph.ts
```diff
@@ -116,7 +116,7 @@ export function buildDependencyGraph(context: GraphBuildingContext): DependencyG
 
     // Add dependencies on internal_tag_ids (direct component tag references)
     if (component.internal_tag_ids && component.internal_tag_ids.length > 0) {
-      component.internal_tag_ids.forEach((tagId) => {
+      component.internal_tag_ids.forEach((tagId: string) => {
```
dipankarmaikap: Nitpick: this type is not adding any value.

dipankarmaikap (approved):
LGTM, the migration looks solid.

One thing I noticed: a few type exports are no longer available from the client package and have been moved to the schema package. For example, `StoryAlternate`, `DatasourceEntryCreate`, and `DatasourceEntryUpdate` were previously exported from `mapi-client`.

I think this should be treated as a breaking change. We should probably mention this somewhere or keep the old exports for backwards compatibility.

.github/workflows/ci.yaml
alexjoverm: It's a way to group the "real" packages (the ones who'll end in npm), to run lint, build, test, etc on them and exclude running it on the playgrounds

alexjoverm: what would be the reason to run build, test, etc on playgrounds itself?

maoberlehner: Same as for production code: Ensuring that they're working and keep working. Yeah, the blast radius of broken playgrounds is smaller (us and contributors), yet I like my playgrounds in a working state when I need them.

If we don't care about some or all of them at all, we can just not set up linting for them and remove / don't add tests. But if they have linting and tests set up, I want those checked. Otherwise, why set them up in the first place?

packages/capi-client/playground/integration-tests/package.json
alexjoverm: I wasn't yet understanding why it'd trigger a release, but your comment here clarifies https://github.com/storyblok/monoblok/pull/626#discussion_r3411572444

I'm surprise as in the past semantic release effectively excluded devDependencies by default as expected, so I was assuming NX would do the same. If you're up to, I'd explore the `updateDependents` or some NX configuration to avoid adding an extra layer of package for tests, but fully up to you, no blocker

.github/workflows/ci.yaml
alexjoverm: Thing is, from my side is still not clear what's the full workflow here. Imagine that, after a week or two working on my branch, I go back to `main` (where a lot of changes happened in the meanwhile):
- in which cases should I re-pull the open api specs?
- In which cases should I regenerate the clients? how?
- is this something we can document, or can we add some sort of command so that when I run `git pull`, I execute it making sure I'm in the latest state of things?

Asking as actually, the other day trying out something on the CLI or schema, it was err and I realized I needed to regenerate the clients to make it work (but didn't fully get the workflow)

.github/workflows/ci.yaml
alexjoverm: Yeah, I fully understand you, they'll be kept up checked. My concern is more on the maintainance side, as that would imply we'd need to add and maintain tests to all playgrounds, producing a testing chain: we use playgrounds to test the packages, and we add tests to test the playgrounds. 

Maybe worth discussing in the weekly? :)

maoberlehner: > we'd need to add and maintain tests to all playgrounds

Not sure if I misunderstand what you or the CI setup means. As I understand it, if we remove `tag:npm:public` here, tests and linting would run for all projects **that have tests and linting set up.** If a playground has tests and linting set up, we probably want them to work, don't we? If they don't have linting and tests set up, then removing the flag wouldn't change anything. Those playground projects would just be skipped because there's nothing to run there, wouldn't they?

So I don't think we need to add any tests or lint setup where we don't have any yet. And if it is a blocker, we can even remove a failing playground test suite or lint setup if we don't care about it failing anyway. What's the point of tests and linting that fail, no matter if playground or not?

If the tests fail for no good reason, then they're useless. If the tests fail for a good reason and the playground is broken, then the playground is useless.

.github/workflows/ci.yaml
maoberlehner: This all comes back to the private OpenAPI specs. It limits the amount of automation we can do.

Usually, NX takes care of that. E.g., when you work on A that needs B and C built, NX will resolve the dependency graph and build B and C before building A.

But we can't automatically run codegen on the latest OpenAPI spec because we can't pull the latest OpenAPI spec in CI, and an external contributor can't do it either.

Yet, to answer your question, **you usually shouldn't have to!** That's why we commit all the generated code, so **you never need to generate code from OpenAPI.**

**Except** if you want to get the latest OpenAPI version pulled in because you need to apply fixes or add new features. But if you don't work on anything like that, everything should just work for you.

If it doesn't, it's a bug!

As I see it, `CONTRIBUTING.md` is mostly for external contributors, but for them, all of that is not relevant because for them, working on moboblok should just work, thanks to the committed generated code. For the agent and us, there is a [section in AGENTS.md](https://github.com/storyblok/monoblok/blob/feat/WDX-406-openapi-overlay/AGENTS.md#openapi-codegen) with a link to the [Codegen docs](https://github.com/storyblok/monoblok/blob/feat/WDX-406-openapi-overlay/tools/openapi-codegen/README.md).

alexjoverm (approved):

packages/cli/src/commands/components/push/graph-operations/dependency-graph.ts
maoberlehner: Removed!

PR #629, opened 2026-06-02
feat(cli): add `assets transfer` subcommand (space to org)

maoberlehner:
Adds `storyblok assets transfer <asset-id...> --folder-id <id>` to move existing assets from a space's local library into the organization's shared asset library. One-way only (space to org); there is no reverse operation.

The CLI surfaces this as `transfer` (DX-aligned wording) and maps it internally to the backend's `convert` endpoint (`POST /v1/spaces/{space_id}/assets/{id}/convert`), with `--folder-id` mapped to the required `target_asset_folder_id` query parameter.

### Changes

- **mapi-client:** new `assets.convertToShared(assetId, { query: { target_asset_folder_id } })` resource method. The endpoint is not in the generated OpenAPI spec, so it issues a raw `client.post`.
- **CLI action:** new `transferAsset(spaceId, assetId, folderId)`. A 403 surfaces a friendly authorization message instead of a raw API error, and the thrown `APIError` preserves the 403 status code.
- **CLI command:** new `assets transfer` subcommand. Required `--folder-id` (validated in-action, rejects missing/zero/negative values, no implicit root), optional `--space` and `--dry-run`. Multiple asset IDs run with bounded concurrency (`Sema(12)`); a per-ID summary prints via the UI module; exit codes are 2 (validation), 1 (any transfer failed), 0 (success or dry-run).
- **Docs:** `transfer/README.md` documenting usage, options, and the transfer-to-convert mapping note.

## Manual testing

Verified live against space qa space on the `eu` region. Help, listing, validation (missing/`0`/`-5`/`abc`), and dry-run were all confirmed; the live convert endpoint was exercised for the 404, 403, and success paths.

To reproduce the full success path you need a target folder in the org's shared asset library that the source space can write to. Without write access the transfer returns 403.

Fixes WDX-409

### DIFF
--- a/packages/cli/src/commands/assets/actions.test.ts
+++ b/packages/cli/src/commands/assets/actions.test.ts
@@ -36,5 +36,7 @@ describe('transferAsset', () => {
       message: expect.stringContaining('not available on your current plan'),
     });
     await expect(transferAsset('123', 42, 7)).rejects.toBeInstanceOf(APIError);
+    const error = await transferAsset('123', 42, 7).catch(e => e);
+    expect(error.code).toBe(403);
   });
 });
--- a/packages/cli/src/commands/assets/actions.test.ts
+++ b/packages/cli/src/commands/assets/actions.test.ts
@@ -35,11 +35,10 @@ describe('transferAsset', () => {
         HttpResponse.json({ error: 'Forbidden' }, { status: 403 })),
     );
 
-    await expect(transferAsset('123', 42, 7)).rejects.toMatchObject({
-      message: expect.stringContaining('write access'),
-    });
-    await expect(transferAsset('123', 42, 7)).rejects.toBeInstanceOf(APIError);
     const error = await transferAsset('123', 42, 7).catch(e => e);
+
+    expect(error).toBeInstanceOf(APIError);
+    expect(error.message).toContain('write access');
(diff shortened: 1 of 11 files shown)


packages/cli/src/commands/assets/actions.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    expect(asset.space_id).toBeNull();
+  });
+
+  it('should surface a friendly authorization message on 403', async () => {
+    server.use(
+      http.post('https://mapi.storyblok.com/v1/spaces/:spaceId/assets/:assetId/convert', () =>
+        HttpResponse.json({ error: 'Forbidden' }, { status: 403 })),
+    );
+
+    await expect(transferAsset('123', 42, 7)).rejects.toMatchObject({
```
dipankarmaikap: It's calling 3 times the same `transferAsset`

dipankarmaikap (approved):
The code is well-structured, follows existing patterns, has comprehensive automated tests, and all manual tests pass.

packages/cli/src/commands/assets/actions.test.ts
maoberlehner: fixed 👍

PR #630, opened 2026-06-02
feat: support global assets library in `assets push` and `assets pull`

maoberlehner:
Teaches the `storyblok` CLI to push and pull assets to either a space or the organization's shared libraries, so both kinds round-trip. A library is a top-level shared asset folder with per-space read or write access.

## How

**mapi-client (`@storyblok/management-api-client`)**

- Adds three hand-written shared resources, mirroring the WDX-409 `convertToShared` pattern (these endpoints are not in the generated SDK, since the WDX-407 overlay will not land): `sharedAssetFolders`, `sharedAssets` (full sign → S3 → finish → get upload flow), and `sharedInternalTags` (every call carries `asset_folder_id` = library root).
- Exports the existing `uploadToS3` helper for reuse.
- Endpoint request/response shapes were verified against `../storyrails` before wiring.

**CLI (`storyblok`)**

- `scope.ts`: a `Scope` discriminated union (`space` | `library`), on-disk base-dir resolution, library listing and access helpers, read-only failfast, and a library-root resolver. The existing transport-injected pipelines are reused unchanged.
- `assets push`: `--target <space|shared|all>` and `--library` flags, a scope loop, read-only-library failfast (no partial state), and library tag remap (`internal_tag_ids` are mapped to the library's shared tags, creating any that are missing). `meta_data` round-trips. Shared-asset creates default the folder to the library root, and the library root folder is skipped on bulk push (it is a shared root folder that cannot be pushed from a space).
- `assets pull`: `--target <with-referenced|all|space|shared>` (default `with-referenced`), which pulls space assets and then resolves shared-library assets referenced by already-pulled local stories. Classification is id-based, never URL-prefix-based.
- `--target` is constrained with Commander `.choices()`, so unknown values fail fast with a non-zero exit.
- Shared `shared_*` API error actions; command READMEs updated.

## On-disk layout

Library assets live under `.storyblok/assets/shared/<library_id>/`, parallel to the space subtree at `.storyblok/assets/<space_id>/`, each with its own `manifest.jsonl`.

## Testing

- mapi-client: new resource tests (msw).
- CLI: push tests (single-asset to library, folder defaulting, root-folder skip, target validation, read-only failfast, bulk `all`/`space`, tag remap, `meta_data` round-trip) and pull tests (`shared`, `space`, `all`, `with-referenced` with and without local stories, target validation), plus a `collectReferencedAssetIds` unit test.
- `lint`, `test:types`, `build`, and `test` are green across both packages.

## Notes

- Full story-content asset-reference rewriting remains out of scope (WDX-411), as does multi-region library enumeration.

Fixes WDX-408


packages/cli/src/commands/assets/push/index.ts
```diff
@@ -69,134 +113,224 @@ pushCmd
       return;
     }
 
-    const summaries: [string, Report['summary'][string]][] = [];
+    const target = (options.target as string) ?? 'space';
```
dipankarmaikap: nitpick: we can improve the type here. 
`type PushTarget = 'space' | 'shared' | 'all'`

packages/cli/src/commands/assets/pull/index.ts
```diff
@@ -55,47 +79,52 @@ pullCmd
...
     const { region } = state;
-
-    const summary = {
-      folderResults: { total: 0, succeeded: 0, failed: 0 },
-      fetchAssetPages: { total: 0, succeeded: 0, failed: 0 },
-      fetchAssets: { total: 0, succeeded: 0, failed: 0 },
-      save: { total: 0, succeeded: 0, failed: 0 },
-    };
-
+    const target = (options.target as string) ?? 'with-referenced';
```
dipankarmaikap: Like in Push, we can also improve the type here, something like `PullTarget`

dipankarmaikap (commented):
Looks good to me.

Other than the two small type improvements I mentioned, I don't see any issues. The tests follow the existing patterns and flow, and my manual testing through the CLI worked as expected.

packages/cli/src/commands/assets/push/index.ts
maoberlehner: Good point!

packages/cli/src/commands/assets/README.md
```diff
@@ -4,11 +4,15 @@ The `assets` module provides tools to manage Storyblok assets and asset folders.
...
 ## Subcommands
 
-- [`pull`](./pull/README.md): Download assets from your Storyblok space.
-- [`push`](./push/README.md): Upload assets to your Storyblok space.
+- [`pull`](./pull/README.md): Download assets from your Storyblok space, and optionally from the organization's shared libraries.
+- [`push`](./push/README.md): Upload assets to your Storyblok space, or to the organization's shared libraries.
 
 > See each subcommand for detailed usage, options, and examples.
 
+## Shared libraries
```
alexjoverm: (blocking): @maoberlehner as usual, all these docs (+ the Readme.md files under the asset sub-commands) should live under the package references on the Docs repositories.

packages/cli/src/commands/assets/push/README.md
```diff
@@ -72,5 +72,7 @@ storyblok assets push --space YOUR_SPACE_ID ./path/to/image.png \
 | `--data <data>` | Inline asset data as JSON (supports Asset fields, `metadata` maps to `meta_data`) | - |
 | `--short-filename <short-filename>` | Override the asset filename (defaults to the source filename) | - |
 | `--folder <folderId>` | Destination asset folder ID | - |
+| `--target <target>` | Push destination: `space`, `org`, or `all` | `space` (single asset), `all` (bulk) |
```
alexjoverm: `org` was renamed to `shared`, and the default seems confusing as `space` or `all` are not related to single or bulk asset moves (+ seems `space` is always the default?)

alexjoverm (changes requested):

packages/cli/src/commands/assets/actions.ts
```diff
@@ -363,3 +364,219 @@ export const transferAssets = async (
...
+    handleAPIError('push_shared_asset_update', toError(maybeError));
+  }
+};
+
+export const fetchSharedInternalTags = async ({ spaceId, libraryId }: {
+  spaceId: string;
+  libraryId: number;
+}) => {
+  try {
+    const { data } = await getMapiClient().sharedInternalTags.list({
```
alexjoverm: needs to be paginated

packages/cli/src/commands/assets/README.md
maoberlehner: See: https://github.com/storyblok/storyblok-docs-platform/pull/556

packages/cli/src/commands/assets/push/README.md
maoberlehner: Good catch! Was confusing and implementation was wrong. I think `all` should be default for bulk push. Fixed!

alexjoverm: @maoberlehner we need to be careful about a thing - if we change it to `all`, it would be a breaking change (soft but still a regression as we're changing the behaviour), plus AFAIK the shared asset library is not available in all plans (and I'd anticipate that in most cases, users might want to push them just to a space, and occasionally either to shared or to both)

Given this considerations, what do you think about keeping it as `space` and letting the user opt in for `all` or `shared` when needed?

maoberlehner: `all` does not mean that it pushes all assets to all targets, but that it pushes space assets to the space and shared assets to their libraries. So it pushes all sources to all respective targets. Therefore, I don't think it is a breaking change, as if you don't have shared assets on disk, `all` behaves the same as `space`.

I realize that it is not quite self-explanatory. If you have a recommendation for better naming, I'd be thankful for suggestions. Also cc @dipankarmaikap

dipankarmaikap: @maoberlehner I think we can keep `all` if we make the description a bit clearer so it's more obvious what it does. Another option could be `auto`, since it makes it clear that the CLI will automatically push each asset to the appropriate destination.

As you mentioned, if there are no shared assets on disk, `all` behaves the same as `space`, so I'm happy with either `all` or `auto`.

maoberlehner: I like `auto`! Will think about it if it doesn't make other cases more confusing, but if not, I will go with it. Thanks!

alexjoverm: Great suggestion!

maoberlehner: Changed to `auto`

alexjoverm (approved):
LGTM - just one thing: remember that the Docs updates should go in the Docs platform. We should remove it from the Readmes to avoid SEO duplication

packages/cli/src/commands/assets/actions.ts
```diff
@@ -363,3 +364,222 @@ export const transferAssets = async (
...
+export const fetchSharedAssets = async ({ spaceId, libraryId, params }: {
+  spaceId: string;
+  libraryId: number;
+  params?: AssetListQuery;
+}) => {
+  try {
+    const client = getMapiClient();
+    const { data, response } = await client.sharedAssets.list({
+      path: { space_id: Number(spaceId) },
+      query: { ...params, in_folder: libraryId, per_page: params?.per_page || 100, page: params?.page || 1 },
```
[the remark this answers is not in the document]
maoberlehner: Verified against a live space: not a bug. `shared_assets#filter_params` strips `in_folder` for access-granted roots, and `AssetFolders::PermittedQuery` returns the root plus all descendant folder ids, so a root-scoped listing returns nested-folder assets too. Confirmed end-to-end: pushed an asset into a child folder and `assets pull --target shared` fetched it. (Separate follow-up: with access to multiple libraries the pull is not scoped per-library.)

.agents/skills/qa-engineer-manual/scripts/cleanup-remote.sh
```diff
@@ -37,6 +57,138 @@ per_page=100
...
+
+  # 2. Shared internal tags scoped to the library (no name filter).
+  tags_resp=$(curl -s --retry 3 --retry-delay 1 "${tags_url}/?asset_folder_id=${library_id}" \
+    -H "Authorization: ${STORYBLOK_TOKEN}")
+  tag_ids=$(printf '%s' "${tags_resp}" | node -e '
+    const data = JSON.parse(require("fs").readFileSync(0, "utf8"));
+    process.stdout.write((data.internal_tags || []).map(t => t.id).join("\n"));
+  ')
+  read -r bf bd <<< "$(delete_ids "${tags_url}" "${library_id}/shared_internal_tags" "?asset_folder_id=${library_id}" "${tag_ids}")"
+  found_total=$((found_total + bf)); deleted_total=$((deleted_total + bd))
```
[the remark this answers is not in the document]
maoberlehner: Confirmed. The `--shared` cleanup is folder-scoped by design (no name filter, so it also catches transferred assets that keep their original names); the docs described it incorrectly. Fixed the docs to match in 1c6b7a138: SKILL.md and GUIDE.md now describe folder-scoped deletion and warn to run it only against a dedicated QA library.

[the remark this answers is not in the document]
maoberlehner: Confirmed: `shared_internal_tags` is a paginated endpoint, so the single unpaginated fetch left tags undeleted past the first page. Fixed in 4ca832d05 by looping over pages (delete-then-refetch), matching the shared assets and folders cleanup sections.

maoberlehner:
> LGTM - just one thing: remember that the Docs updates should go in the Docs platform. We should remove it from the Readmes to avoid SEO duplication

Docs PR was already merged. Kept the updates to the existing README in here so that the README at least is not wrong.

PR #637, opened 2026-06-05
fix(cli): use glob matching for datasources push --filter

### LINKED ISSUE #634: Datasource push --filter arg does not work as advertised
The documentation for `datasource push` describes `--filter` / `--fi` as supporting glob patterns, but the code simply performs `name.includes`. This caused harm on a system of ours where a negative glob pattern was not respected. A proper glob implementation is needed here.

Docs:
https://www.storyblok.com/docs/libraries/storyblok-cli#datasources-push

Code:
https://github.com/storyblok/monoblok/blob/fc52045edea231efa2b95a719114c5bc0e8492ad/packages/cli/src/commands/datasources/push/index.ts#L81

maoberlehner:
## Summary

The `--filter` option of `storyblok datasources push` is documented as a glob pattern but was implemented as a plain substring match (`String.prototype.includes`), so glob tokens like `*`, `?`, and `!` negation were silently ignored. This PR switches the filter to `minimatch`, matching the existing `components push` implementation.

For backward compatibility, filters without glob special characters (`* ? [ ] { } !`) keep the old substring matching, so existing usage like `--filter color` continues to match `colors`.

## Changes

- `packages/cli/src/commands/datasources/push/index.ts`: use `minimatch` when the filter contains glob special characters, fall back to substring matching otherwise
- `packages/cli/src/commands/datasources/push/index.test.ts`: regression tests for glob matching, negation patterns, substring fallback, and the no-match error
- `packages/cli/src/commands/datasources/push/README.md`: document glob semantics and the substring fallback

## Test plan

- [x] `pnpm vitest run src/commands/datasources/push/index.test.ts` (17/17 passing)
- [x] `pnpm nx run-many -t lint,test:types -p storyblok`
- [x] Manual check: `storyblok datasources push --space <id> --filter '!internal-*'` excludes matching datasources (the #634 reproduction)

Fixes WDX-442
Fixes #634

### DIFF
--- a/packages/cli/src/commands/datasources/push/README.md
+++ b/packages/cli/src/commands/datasources/push/README.md
@@ -39,7 +39,7 @@ This will upload a single datasource and its entries from:
 |--------|-------------|---------|
 | `-s, --space <space>` | (Required) The ID of the space to push datasources to | - |
 | `-f, --from <from>` | Source space ID to read datasources from | Target space ID |
-| `--fi, --filter <filter>` | Filter to apply to datasources by their name (e.g., "color" will match datasources containing "color") | - |
+| `--fi, --filter <filter>` | Glob pattern to filter datasources by their name (e.g., "color*" will match all datasources starting with "color"). Filters without glob characters match as substrings (e.g., "color" will match datasources containing "color") | - |
 | `--sf, --separate-files` | Read from separate files instead of consolidated files | `false` |
 | `--su, --suffix <suffix>` | Suffix to add to the files names | - |
 | `-p, --path <path>` | Custom path to read the files from | `.storyblok/datasources` |
@@ -72,7 +72,7 @@ Reads from:
 
 3. Push datasources with filter:
 ```bash
-storyblok datasources push --space 12345 --filter "color"
+storyblok datasources push --space 12345 --filter "color*"
 ```
 Reads from:
 ```
@@ -218,7 +218,7 @@ storyblok datasources push --space 67890 --from 12345
 ### Update specific datasources
 ```bash
 # Push only color-related datasources
-storyblok datasources push --space 12345 --filter "color"
+storyblok datasources push --space 12345 --filter "color*"
 ```
 
 ### Work with separate files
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/datasources/push/README.md
```diff
@@ -39,7 +39,7 @@ This will upload a single datasource and its entries from:
 |--------|-------------|---------|
 | `-s, --space <space>` | (Required) The ID of the space to push datasources to | - |
 | `-f, --from <from>` | Source space ID to read datasources from | Target space ID |
-| `--fi, --filter <filter>` | Filter to apply to datasources by their name (e.g., "color" will match datasources containing "color") | - |
+| `--fi, --filter <filter>` | Glob pattern to filter datasources by their name (e.g., "color*" will match all datasources starting with "color"). Filters without glob characters match as substrings (e.g., "color" will match datasources containing "color") | - |
```
maoberlehner, line 42: Not sure if it is the right call to support both. But it would be a severe change if not. So either we support both, or we only update the docs to mention that there is no glob matching.

dipankarmaikap: I think it’s fine if we support both approaches.

dipankarmaikap (approved):
Non-blocking: LGTM 🚀

I manually tested both the new glob pattern and the existing string-based approach, and everything works as expected. I think it’s fine if we want to support both.

We could either add a warning to indicate that we plan to move fully to glob patterns, or even better, normalise legacy string filters into glob patterns internally for consistency:

```ts
const isGlobPattern = /[*?[\]{}!]/.test(filter);

const pattern = isGlobPattern ? filter : `*${filter}*`;

spaceState.local.datasources = spaceState.local.datasources.filter(datasource =>
  minimatch(datasource.name, pattern),
);
```

packages/cli/src/commands/datasources/push/README.md
maoberlehner: Let's do it that way! I don't even think we need to change it to only supporting glob patterns in the future. Would only consider it if people report confusion/problems with this approach. Thanks!

PR #638, opened 2026-06-05
fix(vue): declare v-editable in GlobalDirectives for template type checking

### LINKED ISSUE #620: [Bug]: v-editable="blok" is unrecognized vue directive in nuxt/vue
### Package

@storyblok/nuxt

### Bug Description

I know this is very minor, but my OCD sense is tingeling

In webstorm, every single storyblok component gets a warning that v-editable="blok" is unrecognized. 

### Steps to Reproduce

Install nuxt and "@storyblok/nuxt": "^10.1.7"

Create any component. 

Add v-editable="blok" and see the warning

### Reproduction URL

_No response_

### Expected Behavior

No warning. It works, but it shows a warning

### Actual Behavior

It shows a warning on every single page

### Environment

```shell
System:
    OS: macOS 26.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 75.67 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.6.0 - /Users/kristoffer/.nvm/versions/node/v24.6.0/bin/node
    npm: 11.5.1 - /Users/kristoffer/.nvm/versions/node/v24.6.0/bin/npm
  Browsers:
    Safari: 26.3.1
  npmPackages:
    @storyblok/nuxt: ^10.1.7 => 10.1.7 
    nuxt: ^4.4.6 => 4.4.6 
    vue: ^3.5.34 => 3.5.34
```

### Error Logs

```shell

```

### Additional Context

_No response_

maoberlehner:
## Summary

- Add a `GlobalDirectives` augmentation to `@storyblok/vue` declaring `vEditable: Directive<HTMLElement>`, so vue-tsc/Volar (WebStorm, VS Code) recognize the globally registered `v-editable` directive under `strictTemplates`
- Add a compile-time regression assertion (`src/global-directives.type-test.ts`) checked by the existing `test:types` target, which CI already runs

## Root cause

The `editable` directive is registered at runtime (`app.directive('editable', ...)`) but was never declared to the type system. Vue 3.4+/Volar resolves global directives via the `GlobalDirectives` interface; without it, every `v-editable` usage produced `TS2339: Property 'vEditable' does not exist`.

The value is typed `Directive<HTMLElement>` to match the runtime directive's own type, avoiding new false positives. Since the Nuxt module hoists `@storyblok/vue` types, `@storyblok/nuxt` users get the fix automatically.

## Verification

- Without the fix, `test:types` fails with `TS2339: Property 'vEditable' does not exist on type 'GlobalDirectives'`
- End-to-end behavior verified manually: a fixture SFC using `v-editable` under `vue-tsc --strictTemplates` reproduces the issue error against the unfixed package and passes with the fix
- With the fix: `lint`, `test:unit`, and `test:types` pass for `@storyblok/vue`

Fixes WDX-434
Fixes #620


dipankarmaikap (approved):
LGTM 🚀

PR #639, opened 2026-06-08
chore: refine dependency rule to discourage fragile hand-rolling

maoberlehner:
## Summary

Rule 5 (`No new deps`) was binary: "avoid dependencies unless strictly necessary." With no counter-weight, agents resolve the ambiguity by hand-rolling error-prone functionality to comply, e.g. a regex-based YAML parser instead of using the `yaml` package (surfaced in #625 review comments).

This reframes the rule so it still discourages casual deps but names the opposite failure mode and gives a concrete decision procedure:

- Write trivial few-line helpers yourself.
- Don't reimplement error-prone functionality (parsing, crypto, dates, globbing).
- Prefer a dependency already in the lockfile (free); otherwise add a small well-established one, or flag the tradeoff rather than hand-rolling silently.

The examples carry the definition, so the vague "non-trivial" / "strictly necessary" wording is gone.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,7 +8,7 @@
 2. **Small diffs** - One logical change per commit
 3. **Plan first** - For non-trivial changes, write a plan before coding
 4. **Verify always** - Run lint/typecheck/tests before considering work done
-5. **No new deps** - Avoid adding dependencies unless strictly necessary
+5. **Deps: prefer none, don't hand-roll** - Write a few-line helper yourself. But don't reimplement error-prone functionality (parsing, crypto, dates, globbing). Use dependencies already in the lockfile; else add a small well-established dep or flag the tradeoff. Never hand-roll silently.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -8,7 +8,7 @@
 2. **Small diffs** - One logical change per commit
 3. **Plan first** - For non-trivial changes, write a plan before coding
 4. **Verify always** - Run lint/typecheck/tests before considering work done
-5. **No new deps** - Avoid adding dependencies unless strictly necessary
+5. **Deps: prefer none, don't hand-roll** - Write a few-line helper yourself. But don't reimplement error-prone functionality (parsing, crypto, dates, globbing). Use dependencies already in the lockfile; else add a small well-established dep or flag the tradeoff. Never hand-roll silently.
```
maoberlehner, line 11: Alternative: just remove it and flag unnecessary new deps in code review.

dipankarmaikap (approved):

PR #645, opened 2026-06-16
feat(schema): refocus on content shapes with new DSL + Zod validators

maoberlehner:
Refocuses `@storyblok/schema` on **content shapes** and adds a non-throwing **runtime validation** layer powered by Zod v4 (internal only — Zod never appears in public types or the `define*` helpers). The CLI now owns DSL→wire mapping, so the schema package does one job: defining the shape of user-generated content.

## Schema package

- **New content-shape DSL** — blocks use `fields: [...]` (was `schema:`); fields use `allow:` (block refs/names, was `component_whitelist`/`restrict_components`) and a new `datasource:` key (ref/slug, was `datasource_slug`); `source:` keeps the option-field selector meaning.
- **Surface reduction** — removed ~30 wire-format payload builders (`defineStoryCreate`, `definePreset`, `defineBlockFolder`, folder entities, …) now covered by the typed MAPI client at the call site.
- **Codegen** — adds the `@hey-api/openapi-ts` `zod` plugin (`compatibilityVersion: 4`) to emit internal content-value schemas; committed at `src/generated/overlay/zod.gen.ts`.
- **Validators** — `validateSchema`, `validateStory` (`{ ok, issues }` with full field-constraint checks — `max_length`, number min/max/decimals/steps, `bloks` allow lists, required fields, value shapes), and `createStoryValidator` (`StandardSchemaV1`).

## CLI

- **DSL→wire mapping** owned by the CLI: `fields[]`→`schema` record, `allow`→`component_whitelist`, `datasource`→`datasource_slug`.
- **Directory layout is the sole grouping mechanism** — component groups are derived from the schema directory paths (parent-first create); `schema init` emits the new DSL into a directory layout.
- **Preset and datasource-entry reconciliation** on `schema push` (inline lists are authoritative for owned children).

## Playgrounds

- astro + vanilla playgrounds migrated to the content-shape DSL; added runnable `validate` examples dogfooding the validators.

Refs WDX-419


packages/schema/scripts/generate.ts
```diff
@@ -15,6 +15,10 @@ const PKG_ROOT = resolve(fileURLToPath(import.meta.url), '../..');
 
 await generate({
   outDir: resolve(PKG_ROOT, 'src/generated'),
+  // `@storyblok/schema` now defines only content shapes, so the surface is the
+  // block/field/datasource definitions plus the story/content types derived
+  // from them. Payload builders and wire-only entities live in the typed
+  // clients, not here.
```
maoberlehner: This is phrased from the perspective of the current task (e.g., "now"), but it should be phrased so that it is useful to someone reading this in the future. So focused on explaining why it is as it is and conveying useful information to future devs, independent of the context of the current task. Review all other new comments in this PR too.

packages/schema/scripts/generate.ts
```diff
@@ -36,36 +40,10 @@ await generate({
...
-    'PresetUpdate',
-    'Space',
-    'SpaceCreate',
-    'SpaceUpdate',
-    'Tag',
-    'User',
-    'UserUpdate',
   ],
+  // Emit internal Zod v4 schemas for the overlay content-value shapes
+  // (asset, richtext, link, table, plugin, block-content). These power the
```
maoberlehner: "asset, richtext, link, table, plugin, block-content" gets stale quickly. Omit or make it "shapes like ..."

packages/schema/scripts/generate.ts
```diff
@@ -36,36 +40,10 @@ await generate({
...
+  // runtime validators; they are never imported by public types.
```
maoberlehner: "they are never imported by public types" not needed info here.

packages/schema/src/generated/types/block.ts
```diff
@@ -2,36 +2,27 @@
...
-/** Wire form: the MAPI object map keyed by field name. This is what `defineBlock` returns. */
-export type BlockSchema = Record<string, Field & { required?: boolean }>;
-
-/** Converts an array-form schema input into the wire-shape object map at the type level. */
-export type SchemaArrayToRecord<T extends BlockSchemaInput> = {
-  [F in T[number] as F['name']]: Omit<F, 'name'>;
-};
+/**
+ * Ordered array of named fields — the content-shape DSL form `defineBlock`
+ * accepts and returns. The array index becomes each field's `pos`.
```
maoberlehner: "The array index becomes each field's `pos`" for somebody not familiar with API shape, this comment is not self explanatory.

packages/schema/src/generated/types/block.ts
```diff
@@ -2,36 +2,27 @@
...
+ */
+export type BlockFields = ReadonlyArray<Field & { name: string; pos?: number; required?: boolean }>;
 
-/** A Storyblok block. */
+/** A Storyblok block (content-shape definition). Carries `fields` (DSL), not the wire `schema` record. */
```
maoberlehner: apendix to comment also only useful in the context of this change, not so much for future users

packages/schema/src/helpers/define-block.ts
```diff
@@ -42,76 +38,58 @@ type BlockInput<
...
 
 /**
- * Returns a {@link Block} with object-shape `schema` (matches the MAPI wire
- * shape). The user-facing input is an ordered array of `defineField` calls;
- * the array index becomes the field's `pos` in the returned map. Throws if
- * two fields share the same `name`.
+ * Returns a {@link Block} content-shape definition. The user-facing input is an
+ * ordered array of `defineField` calls under `fields`; the array index becomes
+ * each field's `pos`. A thin, strongly-typed helper — it does not map to the
+ * MAPI wire shape (the CLI owns that). Throws only on duplicate field names (a
```
maoberlehner: "the CLI owns that" ? the wire shape or the mapping? Not clear here.

packages/schema/src/helpers/schema-type.ts
```diff
@@ -1,41 +1,35 @@
...
 /**
  * Derives a schema types interface from a schema object.
  *
- * Accepts the `typeof` a schema object whose `blocks` property is a
- * record of `defineBlock()` results, and optionally `blockFolders`
- * and `datasources` records, and produces a type with unions of those types.
+ * Accepts the `typeof` a schema object whose `blocks` property is a record of
+ * `defineBlock()` results, and optionally a `datasources` record, and produces
+ * a type with unions of those types. Block folders are no longer schema
+ * entities — they are represented by the schema project's directory structure.
```
maoberlehner: "no longer" - useless info for future devs.

packages/schema/src/validators/shapes.ts
```diff
@@ -0,0 +1,47 @@
+/**
```
maoberlehner, line 5: Do we have the naming `shapes.ts` somewhere else too? Otherwise consider to go for `types.ts`.

packages/schema/src/validators/types.ts
```diff
@@ -0,0 +1,23 @@
+/** Severity of a {@link ValidationIssue}. `error` fails validation; `warning` does not. */
+export type ValidationSeverity = 'error' | 'warning';
+
+/** A single problem found by a validator. */
+export interface ValidationIssue {
```
maoberlehner, line 5: Is this standard schema format or what is this?

packages/schema/src/validators/validate-story.ts
```diff
@@ -0,0 +1,236 @@
...
+
+function validateFieldValue(
+  field: SchemaFieldLike,
+  value: unknown,
+  blocksByName: Map<string, SchemaBlockLike>,
+  path: (string | number)[],
+  entity: string,
+  issues: ValidationIssue[],
+): void {
+  switch (field.type) {
```
maoberlehner, line 53: Does this report a type error if not all possible field types are handled by `case`s?

packages/schema/src/index.ts
```diff
@@ -1,72 +1,47 @@
...
+  TableFieldValue,
+} from './helpers/define-field';
+export type { MapiStory, Story } from './helpers/define-story';
+
+// Schema type helper
 export type { Schema } from './helpers/schema-type';
+
+// Validators (Zod-powered, non-throwing)
+export { createStoryValidator, validateSchema, validateStory } from './validators';
+export type { ValidationIssue, ValidationResult, ValidationSeverity } from './validators';
```
maoberlehner: Remove the `./validators/index.ts` barrel file and instead export here from the original source directly.

tools/openapi-codegen/src/index.ts
```diff
@@ -26,6 +26,13 @@ export interface GenerateConfig {
   include: readonly KnownType[];
   /** Emit the full SDK on this surface in addition to types. */
   sdk?: 'mapi' | 'capi' | false;
+  /**
+   * Additionally emit Zod v4 schemas for the given spec's components into
+   * `<outDir>/<spec>/zod.gen.ts`. Internal: consumer code and public types must
+   * not import it — only runtime validators use these schemas, so the `Zod
+   * never in public types` rule is preserved.
```
maoberlehner: Remove the `Internal: ..` part.

tools/openapi-codegen/src/index.ts
```diff
@@ -98,6 +105,44 @@ export async function generate(config: GenerateConfig): Promise<void> {
...
+
+  if (config.zod) {
+    await emitZodSchemas(config.zod.spec, config.outDir, config.verbose === true);
+  }
+}
+
+/**
+ * Emit Zod v4 schemas for every component in `spec` into `<outDir>/<spec>/zod.gen.ts`.
+ * Runs a dedicated `@hey-api/openapi-ts` pass with the `zod` plugin
+ * (`compatibilityVersion: 4`). Names follow upstream OpenAPI conventions; the
```
maoberlehner: Implementation detail that might get stale and is not needed for consumers.

packages/cli/src/commands/schema/init/actions.ts
```diff
@@ -20,28 +19,31 @@ async function writeFileWithDirs(filePath: string, content: string): Promise<voi
   await writeFile(filePath, content, 'utf-8');
 }
 
-/** Writes all generated schema files to the target directory. */
+/**
+ * Writes all generated schema files to the target directory. Component groups
+ * are encoded by the directory layout: a block in group `Layout` is written to
+ * `components/Layout/<name>.ts`; nested groups nest directories. Blocks whose
```
maoberlehner: lower case/slugify dir names. `My Layout` should become `components/my-layout` double check if full slugify is necessary (spaces allowed).

packages/cli/src/commands/schema/types.ts
```diff
@@ -1,10 +1,17 @@
-import type { Component, ComponentFolder, Datasource } from '../../types';
+import type { Component, ComponentFolder, Datasource, DatasourceEntry, Preset } from '../../types';
 
 /** Local schema loaded from the user's TypeScript entry file. */
 export interface SchemaData {
   components: Component[];
+  /** Component groups derived from the directory layout (name-only; parent encoded by the hierarchy). */
   componentFolders: ComponentFolder[];
```
maoberlehner: Why is this still here?

packages/cli/src/commands/schema/types.ts
```diff
@@ -1,10 +1,17 @@
...
   datasources: Datasource[];
+  /** Block name → its group path segments (from the `components/` directory layout). */
+  folderPathByComponentName: Map<string, string[]>;
+  /** Block name → its inline `presets`, lifted off for reconciliation. */
+  presetsByComponentName: Map<string, Preset[]>;
+  /** Datasource name → its inline `entries`, lifted off for reconciliation. */
+  entriesByDatasourceName: Map<string, DatasourceEntry[]>;
```
maoberlehner: I don't want folderPathByComponentName, presetsByComponentName, entriesByDatasourceName here, why do we need those?

packages/schema/package.json
```diff
@@ -38,6 +38,10 @@
     "lint": "eslint .",
     "lint:fix": "eslint . --fix"
   },
+  "dependencies": {
+    "@standard-schema/spec": "^1.1.0",
+    "zod": "^4.3.6"
```
alexjoverm, line 43: question (blocking): @maoberlehner as the Zod-based runtime validations are an optional path, shouldn't we define these deps as `peerDependencies`? what would be the tradeoffs, considering this package can used client side?

packages/cli/src/commands/schema/utils.ts
```diff
@@ -24,8 +25,12 @@ export const DATASOURCE_DIMENSION_STRIP_KEYS = new Set(['id', 'datasource_id', '
 /** Fields to strip from ComponentFolder during init code generation (keeps uuid for identity). */
 export const FOLDER_INIT_STRIP_KEYS = new Set(['id']);
```
alexjoverm: Seems this one became dead code

packages/cli/src/commands/schema/types.ts
alexjoverm: shouldn't we remove `omponentFolder` here as well?

packages/cli/src/commands/schema/push/actions.test.ts
```diff
@@ -277,7 +277,12 @@ describe('buildChangesetEntries', () => {
   const remoteComp = { id: 2, name: 'hero', schema: {} } as unknown as Component;
   const staleComp = { id: 3, name: 'footer', schema: {} } as unknown as Component;
 
-  const baseLocal: SchemaData = { components: [localComp], componentFolders: [], datasources: [] };
+  const emptyMaps = {
+    folderPathByComponentName: new Map<string, string[]>(),
+    presetsByComponentName: new Map(),
+    entriesByDatasourceName: new Map(),
+  };
+  const baseLocal: SchemaData = { components: [localComp], componentFolders: [], datasources: [], ...emptyMaps };
```
alexjoverm: I'm actually unsure about a thing - in some parts (i.e.: `commands/schema/types`) the `componentFolders` have been removed, but in others like here is kept. Is it intentional? Would it be needed at schema level?

packages/cli/src/commands/schema/map-to-wire.ts
```diff
@@ -0,0 +1,55 @@
+import type { Component, Datasource, Field } from '../../types';
+
+function isRecord(value: unknown): value is Record<string, unknown> {
```
alexjoverm: minor, but this util is duplicated in `schema/push/load-schema` - consider consolidating it into the utils file to ensure same behavior in multiple places used

alexjoverm (commented):
@maoberlehner added my feedback from a code review perspective. Mostly a question to consider, the rest are minors. 

Moving to manual testing now, will approved/request changes when that part is finished

alexjoverm (changes requested):
- Shall we add some validation examples and cases to the `schema/playgrounds/vanilla`? Wouldn't only help with the review, but also I would ourselves often using them to test things out and we'd make the current playground more complete

- Right now, there are duplicated helpers (typed, and non-typed). Is there any way to consolidate them?


```
declare function createStoryHelpers(): {
  defineStory: typeof defineStory;
  defineMapiStory: typeof defineMapiStory;
  defineStoryCreate: typeof defineStoryCreate;
  defineStoryUpdate: typeof defineStoryUpdate;
  withTypes<T extends StoryblokTypesConfig>(): {
    defineStory: DefineStoryTyped<ResolveComponents<T>>;
    defineMapiStory: DefineMapiStoryTyped<ResolveComponents<T>>;
    defineStoryCreate: DefineStoryCreateTyped<ResolveComponents<T>>;
    defineStoryUpdate: DefineStoryUpdateTyped<ResolveComponents<T>>;
  };
};
```

- What would be the difference between defineMapiStory, defineStoryCreate, defineStoryUpdate? Isn't the MAPI using the same payload for all CRUD endpoints? Any way to consolidate/simplify it?

```
/**
 * Returns a full MAPI {@link MapiStory} with all fields populated.
 *
 * @example
 * const myStory = defineMapiStory(pageComponent, {
 *   name: 'My Page',
 *   content: { headline: 'Hello World!' },
 * });
 */
declare function defineMapiStory<const TBlock extends RootBlock, const TBlocks extends Block = never>(component: TBlock, story: MapiStoryInput<TBlock, TBlocks>): MapiStory<TBlock, TBlocks>;
/**
 * Defines a story creation payload for the MAPI.
 *
 * @example
 * const payload = defineStoryCreate(pageComponent, {
 *   name: 'My Page',
 *   content: { headline: 'Hello World!' },
 * });
 */
declare function defineStoryCreate<const TBlock extends RootBlock, const TBlocks extends Block = never>(component: TBlock, story: StoryCreateInput<TBlock, TBlocks>): StoryCreate<TBlock, TBlocks>;
/**
 * Defines a story update payload for the MAPI.
 *
 * @example
 * const payload = defineStoryUpdate(pageComponent, {
 *   content: { headline: 'Updated!' },
 * });
 */
```

- I see in the `generated/overlay/zod.get.ts` we have all zod runtime schemas. Are they customizable (like user might want to add extra rule)? Also, do they adapt to the schema? Like, imagine in the schema of a text field you have `.minLength: 20` statically, will it take it by default and add the runtime assertion for that?

- Seems the Standard Schema has only a `~starndard` readonly prop, forcing when using it to have it to write the accessor explicitly. Is it necessary? Is it something that can be avoided?

```
/** The Standard Schema interface. */
interface StandardSchemaV1<Input = unknown, Output = Input> {
    /** The Standard Schema properties. */
    readonly "~standard": StandardSchemaV1.Props<Input, Output>;
}

// usage

const pageValidator = createStoryValidator(pageBlock, schema);

// instead of this (current)
pageValidator['~standard'].validate(homeStory);

// this (expected)
pageValidator.validate(homeStory);
```

- I see we have 2 alternative API's: either `validateStory(story, schema)` or `createStoryValidator(schema) + .validate(story)`. Shall we rather settle for 1 way for simplicity? In my experience, adding more alternative options can confuse and generate more mantiainance (while users could create their own `createStoryValidator` easily on top of our helper?)

- On top of the validation helper functions: thinking about adding accompanying `validate` commands/flags in the CLI?

alexjoverm:
@maoberlehner extra point/question:

- do we/can we support cross-referencing values from different fields for connected rules? see https://www.sanity.io/docs/studio/validation#k1d8eb8fbb695

packages/cli/src/commands/schema/push/actions.test.ts
maoberlehner: Removed from here and `(Remote)SchemaData`. Only needed for ~~changesets and~~ `init` now.

packages/cli/src/commands/schema/utils.ts
maoberlehner: removed

maoberlehner:
> * Shall we add some validation examples and cases to the `schema/playgrounds/vanilla`? Wouldn't only help with the review, but also I would ourselves often using them to test things out and we'd make the current playground more complete

Did you see `packages/schema/playground/vanilla/app/validate.ts`? Anything to add there?

> * Right now, there are duplicated helpers (typed, and non-typed). Is there any way to consolidate them?
> 
> ```
> declare function createStoryHelpers(): {
>   defineStory: typeof defineStory;
>   defineMapiStory: typeof defineMapiStory;
>   defineStoryCreate: typeof defineStoryCreate;
>   defineStoryUpdate: typeof defineStoryUpdate;
>   withTypes<T extends StoryblokTypesConfig>(): {
>     defineStory: DefineStoryTyped<ResolveComponents<T>>;
>     defineMapiStory: DefineMapiStoryTyped<ResolveComponents<T>>;
>     defineStoryCreate: DefineStoryCreateTyped<ResolveComponents<T>>;
>     defineStoryUpdate: DefineStoryUpdateTyped<ResolveComponents<T>>;
>   };
> };
> ```
> 
> * What would be the difference between defineMapiStory, defineStoryCreate, defineStoryUpdate? Isn't the MAPI using the same payload for all CRUD endpoints? Any way to consolidate/simplify it?
> 
> ```
> /**
>  * Returns a full MAPI {@link MapiStory} with all fields populated.
>  *
>  * @example
>  * const myStory = defineMapiStory(pageComponent, {
>  *   name: 'My Page',
>  *   content: { headline: 'Hello World!' },
>  * });
>  */
> declare function defineMapiStory<const TBlock extends RootBlock, const TBlocks extends Block = never>(component: TBlock, story: MapiStoryInput<TBlock, TBlocks>): MapiStory<TBlock, TBlocks>;
> /**
>  * Defines a story creation payload for the MAPI.
>  *
>  * @example
>  * const payload = defineStoryCreate(pageComponent, {
>  *   name: 'My Page',
>  *   content: { headline: 'Hello World!' },
>  * });
>  */
> declare function defineStoryCreate<const TBlock extends RootBlock, const TBlocks extends Block = never>(component: TBlock, story: StoryCreateInput<TBlock, TBlocks>): StoryCreate<TBlock, TBlocks>;
> /**
>  * Defines a story update payload for the MAPI.
>  *
>  * @example
>  * const payload = defineStoryUpdate(pageComponent, {
>  *   content: { headline: 'Updated!' },
>  * });
>  */
> ```

Those are now completely removed. It was an oversight to leave them in.

> * I see in the `generated/overlay/zod.get.ts` we have all zod runtime schemas. Are they customizable (like user might want to add extra rule)? Also, do they adapt to the schema? Like, imagine in the schema of a text field you have `.minLength: 20` statically, will it take it by default and add the runtime assertion for that?

The generated Zod runtime schemas only provide validators that ensure data is consistent with the shape defined by the OpenAPI spec. Any custom behavior comes through the schema provided helpers. E.g., in this example, it will err on the too-long title: https://github.com/storyblok/monoblok/pull/645/changes#diff-d79640aca53a6d03bacccf4e5a6fb36b1ca7f2bd2c0f9f246381751ad124ae3bR31-R47

> * Seems the Standard Schema has only a `~starndard` readonly prop, forcing when using it to have it to write the accessor explicitly. Is it necessary? Is it something that can be avoided?
> 
> ```
> /** The Standard Schema interface. */
> interface StandardSchemaV1<Input = unknown, Output = Input> {
>     /** The Standard Schema properties. */
>     readonly "~standard": StandardSchemaV1.Props<Input, Output>;
> }
> 
> // usage
> 
> const pageValidator = createStoryValidator(pageBlock, schema);
> 
> // instead of this (current)
> pageValidator['~standard'].validate(homeStory);
> 
> // this (expected)
> pageValidator.validate(homeStory);
> ```

The whole point of `createStoryValidator()` is that it is Standard Schema compatible. For the opinionated and "clean" version, we already have `validateStory()`. For more info on why both, see below.

> * I see we have 2 alternative API's: either `validateStory(story, schema)` or `createStoryValidator(schema) + .validate(story)`. Shall we rather settle for 1 way for simplicity? In my experience, adding more alternative options can confuse and generate more mantiainance (while users could create their own `createStoryValidator` easily on top of our helper?)

They have different use cases:

- `validateStory()` is used directly to validate if a story matches the given schema. Returns a result in our own opinionated format.
- `createStoryValidator()` is only used to create a validator that is passed to another library that is Standard Schema compatible (e.g., tRPC):

```ts
// trpc.ts
import { initTRPC } from '@trpc/server';
import { createStoryValidator } from '@storyblok/schema';
import { pageBlock } from './blocks/page';
import { schema } from './schema';

const t = initTRPC.create();
const pageValidator = createStoryValidator(pageBlock, schema);

export const appRouter = t.router({
  savePage: t.procedure
    .input(pageValidator) // tRPC runs the validator on every call
    .mutation(async ({ input }) => {
      // `input` only reaches here if validation passed.
      // On failure tRPC throws a BAD_REQUEST with the issues.
      return savePageToStoryblok(input);
    }),
});
```

Nobody is ever going to write `pageValidator['~standard'].validate()` (I've removed examples in Playgrounds and replaced them with realistic examples). Do you think we should rename the function to something like `toStoryStandardSchema(block, schema)` or `storyStandardSchema(block, schema)` or is it enough to solve this with docs? I'm in favor of clarifying it in the docs.

> * On top of the validation helper functions: thinking about adding accompanying `validate` commands/flags in the CLI?

See: WDX-452

packages/cli/src/commands/schema/map-to-wire.ts
maoberlehner: Extracted.

packages/schema/package.json
maoberlehner: I think I oversold Zod's optionality when we were talking about it. As it turned out, we need the explicit dependency because the `validate` helpers won't work without it, and making them validation library-agnostic would have other downsides that IMO are not worth it, considering that I'd highly recommend that nobody use those helpers client-side.

The alternative would be to have new factory like `createValidateStory` and two versions of `createStoryValidator`:

```ts
// default, Zod-based helpers (same as now but dedicated export path):
import { validateStory, createStoryValidator } from '@storyblok/schema/zod';

// validator library-agnostic helpers
import { createValidateStory, createStoryValidator } from '@storyblok/schema';
const storyValidate = createStoryValidate(valibot); // Not 100% sure if we can build it so that it is compatible with all the common validator libraries
const validator = createStoryValidator(block, schema, valibot);
```

maoberlehner:
> @maoberlehner extra point/question:
> 
> * do we/can we support cross-referencing values from different fields for connected rules? see https://www.sanity.io/docs/studio/validation#k1d8eb8fbb695

How is it supposed to work? In Sanity, this is how you define the validation rules that get applied in the Sanity Studio UI.

Technically, we could implement validation functions like that and make them fail `validateStory()` but it would be completely decoupled from the Storyblok validation rules. So content editors enter valid (by the rules of the Storyblok backend and frontend) data, but running `validateStory()` will fail. What is the use case for that?

And IMO it would be rather confusing, especially for devs coming from Sanity, because it would be easy to assume that this allows you to define the validation rules that get applied within Storyblok.

packages/cli/src/commands/schema/push/actions.ts
dipankarmaikap, line 1: Should this be `delete_component` instead? (line 187)

packages/schema/src/helpers/define-field.ts
```diff
@@ -1,167 +1,92 @@
...
-  field: TField,
-): Omit<TField, 'name'> & { name: TName } => ({ ...field, name });
+  const TField extends FieldInput,
+>(name: TName, field: TField): DefinedField<TName, TField>;
+export function defineField(name: string, field: Record<string, unknown>): Record<string, unknown> {
+  const { allow, datasource, ...rest } = field;
+  const normalized: Record<string, unknown> = { ...rest, name };
+  if (allow !== undefined) {
+    const refs = Array.isArray(allow) ? allow : [allow];
+    normalized.allow = refs.map(ref => (typeof ref === 'string' ? ref : (ref as { name: string })?.name));
```
dipankarmaikap: Instead of typing `ref` as `{ name: string }`, could we use a proper type guard instead? We have same thing on line 89.

packages/cli/src/commands/schema/map-to-wire.ts
```diff
@@ -0,0 +1,52 @@
...
+ * `Component`. The ordered `fields` array becomes a `schema` record keyed by
+ * field name (each field keeps its `pos`); everything else passes through.
+ */
+export function mapBlockToWire(block: Record<string, unknown>): Component {
+  const { fields, ...rest } = block;
+
+  const schema: Record<string, Field> = {};
+  if (Array.isArray(fields)) {
+    for (const field of fields) {
+      if (!isRecord(field)) { continue; }
```
dipankarmaikap, line 37: Maybe we could add a warning log here to let users debug malformed schema?

dipankarmaikap (commented):
Hey @maoberlehner , the PR looks good to me. I've left a few minor comments, but nothing blocking. I also did multiple rounds of manual testing, and everything passed.

packages/cli/src/commands/schema/map-to-wire.ts
maoberlehner: Good catch, added.

packages/schema/src/helpers/define-field.ts
maoberlehner: Done!

packages/cli/src/commands/schema/push/actions.ts
maoberlehner: Indeed.

packages/schema/package.json
alexjoverm: IMO, no need to have separated agnostic vs zod helpers ;) I believe most people won't care a lot what they're built upon (as long as a good industry standard), but rather that they work well and are easy to use. 

Initially, I was concerned about [Zod's bundle size](https://bundlephobia.com/package/zod@4.4.3), but as you already mentioned, I'd foresee client-side they'll use mostly the generated types in the api client and user code. Would you see any use case where would make sense to use schema client side, where bundle size is at risk?

alexjoverm:
> Do you think we should rename the function to something like toStoryStandardSchema(block, schema) or storyStandardSchema(block, schema) or is it enough to solve this with docs? 

No hard opinion, you'd best to tell :) if I understand correct, they won't use `["~standard"]` accessor mostly because the intention is that they create their own abstractions (like the Hono one you have in the validate playground) yes? If that the case, def docs perfectly fine.

> How is it supposed to work? In Sanity, this is how you define the validation rules that get applied in the Sanity Studio UI.

Yeah... I can feel you there :D ideally, cross-field validation rules is something we rather had in product (and was a FR in the past that I guess never landed). On the other side, if we provide some kind of `validate` property (something equivalent of the `validation` [prop in the defineField](https://www.sanity.io/docs/studio/validation#k4dc8b38bc411)) so that users can set their own custom runtime rules, I can imagine they'd sometimes needs to cross-refer the values of other fields.

Personally, I think some kind of `validation` function can live perfectly in parallel aside from the product, as it's a custom _runtime_ validation depending on the field value and belongs more code-side. What do you think?

packages/schema/package.json
alexjoverm: [Update] re-thinking this, and based on last point in [this comment](https://github.com/storyblok/monoblok/pull/645#issuecomment-4865613355), I actually think would be a common use case to run validations in the user app land (which could be client-side, but that's on them - or on us to document best practices). AFAIK, our own FE and Website teams are doing that on the user app, as you can never trust that the values you receive from the API are the ones you expect :D 

What do you think it's the best we can do for such scenario?

maoberlehner: Only the types and using type imports like `import type { Schema } from '../schema/schema'`, which should not add to bundle size if the bundler is working correctly. Yet I'm sure some users will run into it, but for regular intended use, I don't see a good use case for importing it into the client bundle.

alexjoverm: just to make sure we didn't same a message at same time :D - what do you think about this scenario? https://github.com/storyblok/monoblok/pull/645#discussion_r3512987767

maoberlehner: Haha, definitely we were writing at the same time.
Hmm, I would recommend users do this server-side via a BFF (or server-executed code in Next.js and Nuxt). Runtime validation necessarily adds overhead that should not creep into client-side code, regardless of whether we expose Zod or not. For users who want to stick to a pure SPA architecture, I'd say they need to live with the trade-off of doing without runtime validation or a heavy client bundle.

alexjoverm: That sounds like a valid tradeoff 👍 def a think for documentation.

For the SPA or SSR-client-hidrated cases, as long as they don't import `@storyblok/schema` in their user app nothing will be added to their bundle so sounds great 👍

maoberlehner:
> No hard opinion, you'd best to tell :) if I understand correct, they won't use `["~standard"]` accessor mostly because the intention is that they create their own abstractions (like the Hono one you have in the validate playground) yes? If that the case, def docs perfectly fine.

Yes, this is the intended use case.

> On the other side, if we provide some kind of validate property (something equivalent of the validation [prop in the defineField](https://www.sanity.io/docs/studio/validation#k4dc8b38bc411)) so that users can set their own custom runtime rules, I can imagine they'd sometimes needs to cross-refer the values of other fields.
>
> Personally, I think some kind of validation function can live perfectly in parallel aside from the product, as it's a custom runtime validation depending on the field value and belongs more code-side. What do you think?

I see that this feature is important to you, and I, too, think it is elegantly solved in Sanity. Yet I remain skeptical about how well this works in the Storyblok context. And I also fear that it is confusing to users coming from or comparing us with Sanity. And still, I don't quite understand the use case.

I imagine it like this: dev specifies a `validate` function on a field and pushes the schema. A content editor adds content that does not `validate` but can save it just fine because Storyblok is not aware of `validate`. The dev's runtime validation fails on the newly created story. And then what?

Am I missing something? In which scenarios are devs using this? And how do we communicate to them that this has no effect on content editing in Storyblok at all?

But I don't think we need to find this out and decide this now, as adding it later would not break existing functionality. Would you be okay with creating a follow-up ticket where we can discuss potential use cases and how to design it in a way that clearly separates it from Storyblok native validation features?

alexjoverm:
> I imagine it like this: dev specifies a validate function on a field and pushes the schema. A content editor adds content that does not validate but can save it just fine because Storyblok is not aware of validate. The dev's runtime validation fails on the newly created story. And then what?

You had a great point here. Agree, can be a two-edged sword. And also agree we can think about it separately, might be a FR we never even receive

alexjoverm (approved):
Great work Markus!

PR #648, opened 2026-06-18
fix(qa-skill): make seed-scenario fail loudly instead of reporting phantom success

maoberlehner:
Hardens the `qa-engineer-manual` skill's `seed-scenario.sh` so it can no longer report success while pushing nothing, and documents the seeding behavior that makes that failure mode easy to misread.

## Why

While manually QAing another branch, the seed reported `Pushing stories ... done` and `Seeded: 10 stories` while `0` stories actually landed. Root cause:

- Every CLI push ran with output sent to `/dev/null` and printed `done` unconditionally, so a push that errored, or that exited `0` but created nothing (e.g. stories pushed against missing components), still looked successful.
- The parallel datasources/assets push compounded it: its multi-pid `wait` either swallowed a failure or, via the `ERR` trap, aborted the whole seed before stories with no clear message.

This produced "phantom success" that was easy to misdiagnose as the shared space "self-wiping" itself.

## Changes

- Add a `run_push` helper that captures output, retries once, and prints `done` only on real success; on failure it surfaces the captured CLI output and exits non-zero. Applied to login and all four pushes.
- Serialize the pushes (components, datasources, assets, stories), removing the brittle background-job + `ERR`-trap machinery. Assets still precede stories (stories remap asset IDs).
- Verify each staged resource actually landed remotely after pushing (remote count `>=` staged count, with a short retry for propagation lag) and fail with a clear `Verification FAILED` message otherwise.
- Factor the duplicated staged-file counting into a `count_staged` helper.
- Document in `SKILL.md` (Seeding section) that every seed wipes the space first and that a `done` log line alone does not prove data landed.

### DIFF
--- a/.agents/skills/qa-engineer-manual/scripts/seed-scenario.sh
+++ b/.agents/skills/qa-engineer-manual/scripts/seed-scenario.sh
@@ -87,6 +87,10 @@ require_space_id
 # Constants
 # ---------------------------------------------------------------------------
 FAKE_ID="qa-seed"
+# Total attempts per CLI push (1 = no retry; 2 = one retry on failure).
+PUSH_ATTEMPTS=2
+# Total attempts when verifying a resource landed remotely (absorbs read-after-write lag).
+VERIFY_ATTEMPTS=3
 repo_root="$(git rev-parse --show-toplevel)"
 cli_bin="${repo_root}/packages/cli/dist/index.mjs"
 skill_dir="${repo_root}/.claude/skills/qa-engineer-manual"
@@ -161,7 +165,7 @@ count_staged() {
 # unconditionally).
 run_push() {
   local label="$1"; shift
-  local attempts=2 attempt=1 status=0 out=""
+  local attempts="${PUSH_ATTEMPTS}" attempt=1 status=0 out=""
   printf "Pushing %s ... " "${label}"
   while [ "${attempt}" -le "${attempts}" ]; do
     out=$("$@" 2>&1) && { printf "done\n"; return 0; }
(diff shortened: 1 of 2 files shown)


alexjoverm (approved):

.agents/skills/qa-engineer-manual/scripts/seed-scenario.sh
```diff
@@ -129,6 +134,65 @@ cleanup() {
...
+}
+
+# Runs a CLI push, capturing output and retrying once on failure. Prints "done"
+# only on real success; on failure it prints the captured CLI output and exits
+# non-zero, so the seed can never report phantom success for a push that
+# actually errored (previously output was sent to /dev/null and "done" printed
+# unconditionally).
+run_push() {
+  local label="$1"; shift
+  local attempts=2 attempt=1 status=0 out=""
```
dipankarmaikap: Nitpick: We could make this a constant so it's easier to control and update in one place, for example:

`PUSH_RETRIES = 2`

.agents/skills/qa-engineer-manual/scripts/seed-scenario.sh
```diff
@@ -129,6 +134,65 @@ cleanup() {
...
+  exit 1
+}
+
+# Asserts a resource actually landed remotely. Catches pushes that exit 0 but
+# create nothing (e.g. stories pushed against missing components): the remote
+# count must be at least the staged count. Retries briefly to absorb
+# read-after-write propagation lag.
+verify_seeded() {
+  local resource="$1" expected="$2" actual="" attempt=1
+  while [ "${attempt}" -le 3 ]; do
```
dipankarmaikap: Nitpick: Like `PUSH_RETRIES`, we can extract this into a `VERIFY_RETRIES` constant and reuse it.

dipankarmaikap (approved):
LGTM, approving ✅

I left two minor, non-blocking pieces of feedback for your consideration.

.agents/skills/qa-engineer-manual/scripts/seed-scenario.sh
maoberlehner: Changed

.agents/skills/qa-engineer-manual/scripts/seed-scenario.sh
maoberlehner: Changed

PR #649, opened 2026-06-18
feat(experiments): add @storyblok/experiments package (WDX-458)

maoberlehner:
New `@storyblok/experiments` package: framework-agnostic, server-first, dependency-free runtime A/B testing utilities built on Storyblok's experiments API (`GET /v2/cdn/experiments`).

Variants are whole separate stories (`original_slug` → `variant_slug`), not field-level content. Given the experiments payload (fetched by the user's own client), the library buckets a visitor, decides which story slug to render, and tracks exposure/conversion events.

## API

- `assignVariant({ experiment, visitorId })` — deterministic, sticky, storage-free bucketing (FNV-1a hash over `visitorId:experimentId`, weight-normalized).
- `resolveExperiment({ experiments, slug, assignment })` — maps the original slug to the control's original slug or a variant's `variant_slug`, with an exposure event. No-match passes the slug through unchanged.
- `trackEvent(event, { adapter })` and `createExperiments({ experiments, adapters })` factory — auto-fires exposure on resolve, attributes later conversions through adapters.
- `@storyblok/experiments/adapters` subpath — `fetchAdapter(url, opts?)` POSTs events as JSON; any `(event) => void | Promise` is a valid custom adapter.

## Backend dependency: `original_slug` must be a full slug

`resolveExperiment` matches the caller's slug against each mapping's `original_slug`. The CDN payload is only self-consistent for **root-level** stories today: `variant_slug` was fixed to return the full slug in `storyrails` [SHAPE-13989 (#8088)](https://github.com/storyblok/storyrails/pull/8088), but `original_slug` still serializes `story.slug` (the last path segment), not `story.full_slug`.

Consequence: for **folder-nested** stories the caller routes by full path (`pages/home`) while `original_slug` is just `home`, so no experiment matches and the slug passes through unchanged — control content is served and no exposure fires. Root-level experiments (where `slug === full_slug`) work correctly, which is what the e2e round-trip exercises.

The SDK can't work around this locally (it only has `original_slug` from the payload, and tail-matching is ambiguous across folders). Folder-nested experiments require a follow-up backend change mirroring #8088: `original_slug` → `full_slug` in both `cdn/experiment_variant_serializer.rb` and `experiment_story_mapping_serializer.rb`.

## Tests

- Unit tests co-located per module (hash, assign, resolve, track, adapters, create-experiments).
- **Manual end-to-end test** (`*.spec.e2e.ts`, never run in CI): drives the MAPI to set up a real running experiment, then round-trips the package against the live `GET /v2/cdn/experiments` and `cdn/stories` endpoints. The MAPI client has no experiments resource yet, so setup uses the generic HTTP escape hatches (`mapi.post/put/delete`). Covered by `test:types` and `lint`; excluded from the default `test` run by the `.spec.e2e.ts` suffix.

Fixes WDX-458


maoberlehner:
Superseded by #687. This PR closed automatically when its head branch was renamed `feat/WDX-458-experiments` → `feat/DX-458-experiments` (WDX→DX prefix rename). Continue review in #687.

PR #654, opened 2026-06-23
fix(richtext): require content on doc root, keep it optional on child nodes

### LINKED ISSUE #16: `StoryblokRichTextNode` type not 100% correct

---

## Expected Behavior

If I investigated it correctly, the type should require `content` to be defined at the first level, but all sub-nodes don't need `content` to be present.

## Current Behavior

The API can return `undefined` as `content` in sub-nodes of `StoryblokRichTextNode`, but the types right now say that in every level of the RichText structure content is present.

We use a custom type which raises the following error because sub-nodes always need content present, which is not the case at the API level.

image

## Steps to Reproduce

N/A

Lorti:
## Summary

Make `content` required on the root `doc` node and keep it optional on every other node in the generated `SbRichTextNode` type, so the types match what the Storyblok API actually returns. Also stop `normalizeNodes` from forcing `content: undefined` onto nodes that have none, and add a compile-time regression assertion checked by the existing `test:types` target.

## Root cause

The richtext node type is generated by `packages/richtext/src/static/generate/richtext-type.ts`, which emitted `content?: SbRichTextNode[]` for every node. As a result `content` was optional everywhere, including the document root. The Storyblok API always returns the root `doc` with a `content` array, while nested nodes (empty paragraphs, leaf nodes such as `image`) may omit `content` entirely.

## Note

Tightening `doc.content` from optional to required is a minor breaking change for code that builds `{ type: 'doc' }` literals without `content`. It matches both the issue's stated expectation and the API response, but I am happy to scope this down to only relax sub-nodes if you would prefer to avoid the breaking change.

Fixes #16

### DIFF
--- a/packages/richtext/src/static/generate/richtext-type.ts
+++ b/packages/richtext/src/static/generate/richtext-type.ts
@@ -36,7 +36,14 @@ function nodeShape(schema: Schema, name: string): string {
   const type = schema.nodes[name];
   const hasText = name === 'text' ? 'text: string;' : '';
   let children = '';
-  if (type.isBlock || name === 'doc' || type.spec.content || type.isInline) {
+  // The root `doc` node is the document body and always carries a `content`
+  // array, so it is required. Every other node may legitimately omit `content`,
+  // for example an empty paragraph or a leaf node like `image`, so for those it
+  // stays optional, matching what the Storyblok API actually returns. See #16.
+  if (name === 'doc') {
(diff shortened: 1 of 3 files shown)


maoberlehner:
Servus @Lorti! Looks good to me - @dipankarmaikap, please take a second look.

dipankarmaikap (approved):

PR #655, opened 2026-06-26
fix(cli): mark conditionally-returned StoryblokAsset fields as optional

### LINKED ISSUE #488: 🐛 [Bug]: Asset type does not reflect API response (is_private and Cloudinary fields missing / not optional)
### Package

storyblok (CLI)

### Bug Description

**Link to type definition**

https://github.com/storyblok/monoblok/blame/main/packages/cli/src/types/storyblok.ts#L15

**Description**

The generated TypeScript type for Storyblok image assets does not fully match the shape returned by the API.

In particular, some properties appear **conditionally in the API response**, but the type definition assumes a stricter structure.

For example, the field `is_private` sometimes appears in the asset response and sometimes does not.

Example responses from the same space:

Example 1:

```json
{
  "backgroundImage": {
    "id": 1,
    "alt": "",
    "name": "",
    "focus": "",
    "title": "",
    "source": "",
    "filename": "https://a.storyblok.com/f/...",
    "copyright": "",
    "fieldtype": "asset",
    "meta_data": {},
    "is_external_url": false
  }
}
```

Example 2:

```json
{
  "id": 2,
  "alt": "",
  "name": "",
  "focus": "",
  "title": "",
  "source": "",
  "filename": "https://a.storyblok.com/f/...",
  "copyright": "",
  "fieldtype": "asset",
  "meta_data": {},
  "is_private": false,
  "is_external_url": false
}
```

`is_private` is sometimes included and sometimes omitted.

Additionally, fields related to Cloudinary integrations can also appear in asset responses depending on configuration.

**Expected behavior**

The generated asset type should mark these fields as optional:

* `is_private`
* Cloudinary-related properties (when present)
* Any other conditionally returned asset properties

Example adjustment:

```ts
type StoryblokAsset = {
  id: number
  alt?: string
  name?: string
  focus?: string
  title?: string
  source?: string
  filename: string
  copyright?: string
  fieldtype: "asset"
  meta_data?: Record<string, unknown>

  is_external_url?: boolean
  is_private?: boolean

  // Cloudinary integration fields (optional)
  width?: number
  height?: number
...
}
```

**Why this matters**

The current type definitions can cause TypeScript errors or incorrect assumptions about the API response structure when consuming assets in generated types.

Making these fields optional would better reflect the actual API responses.

### Steps to Reproduce

Initialize a project using Storyblok and generate types using the monoblok CLI.

Use an asset field in a component schema.

Fetch content that contains that asset.

Inspect the API response.

Notice that some responses contain is_private while others do not.

### Expected Behavior

The generated asset type should reflect the actual API response structure.

Conditionally returned properties such as:

is_private

Cloudinary-related properties

should be optional in the generated type.

### Actual Behavior

The generated type assumes a fixed structure for assets, but the API returns slightly different shapes depending on configuration and asset state.

This can cause incorrect TypeScript assumptions when consuming the generated types.

### Code Sample

```javascript

```

### Environment

```shell
-
```

### Error Logs

```shell

```

### Additional Context

_No response_

MFA-G:
## What & why

The generated `StoryblokAsset` type (`packages/cli/src/types/storyblok.ts`) declares **every** property as required. However, several fields only appear in *some* asset responses depending on the asset, the space configuration, and integrations such as Cloudinary.

The two examples in #488 from the same space confirm this — `is_private` is present in one payload and absent in the other:

```jsonc
// no is_private
{ "id": 1, "alt": "", "name": "", "filename": "https://a.storyblok.com/f/...", "fieldtype": "asset", "meta_data": {}, "is_external_url": false }

// has is_private
{ "id": 2, "alt": "", "name": "", "filename": "https://a.storyblok.com/f/...", "fieldtype": "asset", "meta_data": {}, "is_private": false, "is_external_url": false }
```

Because this interface is read verbatim and emitted into the generated `storyblok.d.ts` (`generateStoryblokTypes` in `commands/types/generate/actions.ts`), consumers get a type that is stricter than the real API, forcing `// @ts-ignore` or unsafe casts when a field is missing.

This also already disagreed with the project's own JSON schema in `utils/storyblok-schemas.ts`, whose `required` list is only:

```
['id', 'fieldtype', 'filename', 'name', 'title', 'focus', 'alt']
```

## Change

Align the `StoryblokAsset` interface with that JSON schema by keeping the 7 core fields required and marking the remaining, conditionally-returned fields optional (`copyright`, `meta_data`, `source`, `is_external_url`, `is_private`, `src`, `updated_at`, and the Cloudinary keys `width`/`height`/`aspect_ratio`/`public_id`/`content_type`).

No runtime behavior changes — this is a type-only fix.

## Testing

From `packages/cli`:

- `pnpm nx run storyblok:test:types` (tsc --noEmit) — passes
- `pnpm nx run storyblok:lint` — passes
- `pnpm vitest run` — 717 passed, 1 todo, 1 skipped

Fixes #488

### DIFF
--- a/packages/cli/src/types/storyblok.ts
+++ b/packages/cli/src/types/storyblok.ts
@@ -2,25 +2,28 @@ export type StoryblokPropertyType = 'asset' | 'multiasset' | 'multilink' | 'tabl
 
 export interface StoryblokAsset {
   alt: string | null;
-  copyright: string | null;
   fieldtype: 'asset';
   id: number;
   filename: string | null;
   name: string;
   title: string | null;
   focus: string | null;
-  meta_data: Record<string, any>;
-  source: string | null;
-  is_external_url: boolean;
-  is_private: boolean;
-  src: string;
-  updated_at: string;
+  // The properties below are only present in some asset responses (they
+  // depend on the asset itself, the space configuration and integrations such
+  // as Cloudinary), so they are optional to reflect the actual API payloads.
+  copyright?: string | null;
+  meta_data?: Record<string, any>;
+  source?: string | null;
+  is_external_url?: boolean;
+  is_private?: boolean;
+  src?: string;
+  updated_at?: string;
   // Cloudinary integration keys
-  width: number | null;
-  height: number | null;
-  aspect_ratio: number | null;
-  public_id: string | null;
-  content_type: string;
+  width?: number | null;
+  height?: number | null;
+  aspect_ratio?: number | null;
+  public_id?: string | null;
+  content_type?: string;
 }
 
 export interface StoryblokMultiasset extends Array<StoryblokAsset> {}


maoberlehner (approved):
Looks good to me! Thank you for your contribution @MFA-G!

PR #661, opened 2026-06-30
fix(astro): circular dependency through StoryblokComponent

### LINKED ISSUE #547: [Bug]: TDZ ReferenceError when Storyblok components have circular dependency through `StoryblokComponent`

dipankarmaikap:
This PR fixes the TDZ (`ReferenceError`) caused by circular dependencies when Storyblok components are imported directly and also render `<StoryblokComponent />` internally.

It also adds both unit and Cypress tests to cover this scenario and help prevent future regressions.

Fixes #547.

### DIFF
--- a/packages/astro/src/vite-plugins/vite-plugin-import-storyblok-components.ts
+++ b/packages/astro/src/vite-plugins/vite-plugin-import-storyblok-components.ts
@@ -3,14 +3,12 @@ import { normalizePath } from '../utils/normalizePath';
 import { toCamelCase } from '../utils/toCamelCase';
 import type { Plugin } from 'vite';
 
-// Virtual module identifiers for Vite's module system
 const VIRTUAL_MODULE_ID = 'virtual:import-storyblok-components';
 const RESOLVED_VIRTUAL_MODULE_ID = `\0${VIRTUAL_MODULE_ID}`;
+
 /**
- * Vite plugin that automatically imports Storyblok components from a specified directory
- * and merges them with optional user-provided component mappings.
- *
- * @returns Vite plugin object
+ * Vite plugin that auto-imports Storyblok components from a directory
+ * and merges them with user-provided component mappings.
  */
 export function vitePluginImportStoryblokComponents(
   components: Record<string, string>,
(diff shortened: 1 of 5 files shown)


packages/astro/src/vite-plugins/vite-plugin-import-storyblok-components.ts
```diff
@@ -224,12 +267,27 @@ interface CreateComponentRegistrationCodeOptions {
...
   importPath,
 }: CreateComponentRegistrationCodeOptions): string {
+  // Use a unique variable name to avoid conflicts between components
+  const varName = `__${componentName}_component__`;
+  // Wrap in an object with a getter to defer access and avoid TDZ issues
+  const wrapperName = `__${componentName}_wrapper__`;
   return `
-  import ${componentName} from '${importPath}';
-  registerComponent('${componentName}', ${componentName});
+import ${varName} from '${importPath}';
```
maoberlehner: ~~nitpick: indentation issue here? Why not picked up by linter?~~
Edit: just saw this is a string output.

packages/astro/src/vite-plugins/vite-plugin-import-storyblok-components.ts
```diff
@@ -66,18 +66,26 @@ export function vitePluginImportStoryblokComponents(
     },
   };
 }
+export interface ResolvedComponent {
+  componentName: string;
+  importPath: string;
+}
```
maoberlehner, line 59: I think this is dead code.

packages/astro/src/vite-plugins/vite-plugin-import-storyblok-components.ts
```diff
@@ -224,12 +267,27 @@ interface CreateComponentRegistrationCodeOptions {
...
+  // Wrap in an object with a getter to defer access and avoid TDZ issues
+  const wrapperName = `__${componentName}_wrapper__`;
   return `
-  import ${componentName} from '${importPath}';
-  registerComponent('${componentName}', ${componentName});
+import ${varName} from '${importPath}';
+const ${wrapperName} = { get default() { return ${varName}; } };
+registerComponent('${componentName}', ${wrapperName});
 `.trim();
 }
```
maoberlehner, line 211: Could we return a structure like:

```ts
interface ComponentRegistrationParts {
  importStatement: string;      // import __X_component__ from '...';
  wrapperDefinition: string;    // const __X_wrapper__ = { get default() ... };
  registrationCall: string;     // registerComponent('X', __X_wrapper__);
}
```

instead of a string?

I think this would greatly help to simplify the code.

packages/astro/tests/vite-plugin-import-storyblok-components.test.ts
```diff
@@ -0,0 +1,102 @@
...
+      // Glob components registration loop should come before manual registration
+      const globLoopIndex = code.indexOf('for (const filePath in modules)');
+      const manualRegisterIndex = code.indexOf('registerComponent(\'card\'');
+
+      expect(globLoopIndex).toBeGreaterThan(-1);
+      expect(manualRegisterIndex).toBeGreaterThan(-1);
+      expect(globLoopIndex).toBeLessThan(manualRegisterIndex);
+    });
+  });
+});
```
maoberlehner, line 197: These tests hand-write the `manualImports` strings, but they don't match what `createComponentRegistrationCode` actually generates. They're missing the `const __x_wrapper__ = { get default() ... }` line and register the raw `__card_component__` instead of the wrapper. So the getter-wrapper path (the actual TDZ fix) is not tested. The tests would still pass if the fix were reverted.

If we move to the structured return above, we could feed real `createComponentRegistrationCode(...)` output into `generateModuleCode` here and assert the wrapper shows up and lands after `registerComponent` is defined.

maoberlehner (changes requested):

maoberlehner (approved):
LGTM!

PR #663, opened 2026-07-01
fix(cli): replace Angular env placeholders with actual values

dipankarmaikap:
## Summary

Updates Angular environment file handling in the `create` command to use simple string replacement for placeholders instead of regex-based value matching.

## Changes

- Add `updateAngularEnvironmentFiles` function for Angular-specific environment file handling
- Pass `template` parameter to `handleEnvFileCreation` to detect Angular projects
- Use `replaceAll` to replace `STORYBLOK_DELIVERY_API_TOKEN` and `STORYBLOK_REGION` placeholders with actual values

## Before

The Angular environment files kept the placeholder names as literal strings.

## After

```ts
// environment.ts
export const environment = {
  production: true,
  accessToken: 'actual-token-here', // Replace with your Storyblok Delivery API token
  region: 'eu', // Replace with your Storyblok region
};
```

## Related

Works with the Angular blueprint update: https://github.com/storyblok/blueprint-core-angular/pull/18

### DIFF
--- a/packages/cli/src/commands/create/actions.test.ts
+++ b/packages/cli/src/commands/create/actions.test.ts
@@ -1,22 +1,13 @@
 import { spawn } from 'node:child_process';
-import fs from 'node:fs/promises';
 import { vol } from 'memfs';
 import { beforeEach, describe, expect, it, type MockedFunction, vi } from 'vitest';
 import open from 'open';
 import { createEnvFile, extractPortFromTopics, fetchBlueprintRepositories, generateProject, handleEnvFileCreation, openSpaceInBrowser, repositoryToTemplate, updateAngularEnvironmentFiles } from './actions';
-import * as filesystem from '../../utils/filesystem';
 import type { RegionCode } from '../../constants';
 import { appDomains } from '../../constants';
 
 vi.mock('node:child_process');
-vi.mock('node:fs/promises', () => ({
-  default: {
-    access: vi.fn(),
-    readFile: vi.fn(),
-  },
-}));
 vi.mock('open');
-vi.mock('../../utils/filesystem');
 vi.mock('../../github', () => ({
   createOctokit: vi.fn(),
 }));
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/create/actions.test.ts
```diff
@@ -35,6 +36,7 @@ const mockedSpawn = vi.mocked(spawn);
 const mockedOpen = open as MockedFunction<typeof open>;
 const mockedSaveToFile = filesystem.saveToFile as MockedFunction<typeof filesystem.saveToFile>;
 const mockedFsAccess = vi.mocked(fs.access);
+const mockedFsReadFile = vi.mocked(fs.readFile);
```
maoberlehner: Better approach to test file system operations is to use `memfs` as in some of our other tests. I leave it up to you if you change it here or not, but consider it for future tests.

packages/cli/src/commands/create/actions.ts
```diff
@@ -115,8 +115,79 @@ ${Object.entries(storyblokVars).map(([key, value]) => `${key}=${value}`).join('\
...
+        content = content.replaceAll('STORYBLOK_REGION', region);
+      }
+
+      await saveToFile(filePath, content);
+    }
+    catch (error) {
+      const fsError = error as NodeJS.ErrnoException;
+      // If file doesn't exist, skip it silently
+      if (fsError.code === 'ENOENT') {
+        continue;
```
maoberlehner, line 155: Nitpick CLI still returns `ui.ok("Updated Angular environment files with: accessToken, region").` although nothing was updated. This makes it harder to find the problem in case the environment files do not exist (in the right place) for example.

maoberlehner (approved):
Looks good, left comments, your call if you want to act on them.

packages/cli/src/commands/create/actions.test.ts
dipankarmaikap: Thank you. I have updated it to use `memfs`

packages/cli/src/commands/create/actions.ts
dipankarmaikap: I have updated it.

PR #664, opened 2026-07-01
feat(cli): schema affected command (WDX-470)

maoberlehner:
Add `storyblok schema affected <entry-file>`: a read-only dry run of `schema push` that reports **which stories a pending schema change affects and which would break**, with a terminal summary and an optional non-zero exit for CI gating.

## How it works

1. `loadSchema` (local) then `fetchRemoteSchema` then `diffSchema` then `analyzeBreakingChanges` (reuses the push pipeline).
2. Compute impacted components (breaking updates, plus removed components when `--include-deleted` is set).
3. Fetch only the stories that use an impacted component, then run `validateStory` from `@storyblok/schema` for breakage. Use `--local` to analyze already-pulled story JSON instead of fetching from the space.
4. Walk each story's content for usage, and aggregate per-component and per-field totals.
5. Print the summary. The full per-story and per-field detail is written to the standard command report file when reporting is enabled (`--report-enabled`).

### Options

`-s/--space` (required), `-p/--path` (global), `--local`, `--include-deleted`, `--fail-on-break`.

See `packages/cli/src/commands/schema/affected/README.md` for the full option table and analysis scope.

## Notable correctness decisions

- **Breakage is diffed against old *and* new schema**, so only errors the change *introduces* count, and pre-existing invalid content is not misattributed.
- **Removed-component impact is gated behind `--include-deleted`**, mirroring `schema push --delete` (a plain push does not delete stale components).
- **Per-field `used` comes from the actual content walk**, so silent-but-real changes (for example `text` to `textarea`) still report usage.
- **Per-story fetch failures are logged and counted**, not silently dropped, and the progress bar reaches 100%.
- **Story fetching combines two MAPI filters.** `contain_component` matches nested-blok usage, and it has AND/superset semantics for a comma-separated list (confirmed via manual QA against a live space and storyrails source: `components @> [...]`), so we issue one request per impacted component and de-dupe refs by story id. A `filter_query[component][in]` request additionally matches stories that use an impacted component as their root content type with no nested bloks, which `contain_component` alone misses. Both are unioned by story id.
- **`--fail-on-break`** sets a non-zero exit code when any story would break, and misconfiguration or operational failures (empty entry file, missing `--space`, fetch errors) also exit non-zero so a broken run never reads as a green CI gate.
- **Shared pipeline helpers** (`load-schema`, `diff-schema`, `migrations/`) moved up to `schema/` so sibling subcommands no longer import from `push/`.

Fixes WDX-470


packages/cli/src/commands/schema/affected/index.ts
```diff
@@ -0,0 +1,159 @@
...
+  computeImpactedComponents,
+} from './actions';
+import { formatJson, formatStoryList, formatSummary } from './format';
+
+schemaCommand
+  .command('affected <entry-file>')
+  .alias('impact')
+  .description('Report which stories a schema change affects and which would break')
+  .option('-s, --space <space>', 'space ID')
+  .option('-p, --path <path>', 'path for file storage')
```
maoberlehner: Make this more descriptive. Also in the schema commands.

packages/cli/src/commands/schema/affected/index.ts
```diff
@@ -0,0 +1,159 @@
...
+  .option('--component <names>', 'restrict analysis to these components (comma-separated)')
```
maoberlehner: Remove this if not strictly needed for something

packages/cli/src/commands/schema/affected/index.ts
```diff
@@ -0,0 +1,159 @@
...
+  .option('--stories <path>', 'analyze locally pulled story JSON files instead of fetching remote')
```
maoberlehner: I don't like this flag. As it should work with the default path where the CLI saves stories. Let's brainstorm how to do this differently.

packages/cli/src/commands/schema/affected/index.ts
```diff
@@ -0,0 +1,159 @@
...
+  .option('--delete', 'treat remote-only components as removed (mirrors `schema push --delete`)', false)
```
maoberlehner: Weird flag. Sounds scary. Brainstorm this if we need it and alternative naming if so.

packages/cli/src/commands/schema/affected/index.ts
```diff
@@ -0,0 +1,159 @@
...
+  .option('--list', 'print the full list of affected stories', false)
```
maoberlehner: Do we need this?

packages/cli/src/commands/schema/affected/index.ts
```diff
@@ -0,0 +1,159 @@
...
+  .option('--output <file>', 'write the detailed impact report to a JSON file')
```
maoberlehner: Remove. Let's only rely on the report feature for json output.

PR #666, opened 2026-07-01
feat(schema): custom field plugins with Standard Schema validators

maoberlehner:
Lets a schema **register** its custom field plugins via a [Standard Schema](https://standardschema.dev) validator (Zod, Valibot, ArkType, or hand-written). A registered `field_type` then resolves to `validatorOutput & { plugin: string; _uid?: string }` for both **read and write**; unregistered custom fields keep today's `PluginFieldValue`. Non-breaking — the default plugin map is empty.

## How

- **New API in `@storyblok/schema`:** `defineFieldPlugin`, `defineSchema`, and `InferSchema` deriving a `fieldType → InferOutput<validator>` map, threaded end-to-end through a new `TFieldPlugins` generic — across `@storyblok/schema` **and** the capi/mapi clients' `withTypes` plumbing.
- **New tree-shakeable entrypoint** `@storyblok/schema/field-plugins`, shipping one official hand-written plugin, `storyblokColorField` (`field_type: 'storyblok-colorpicker'`, value `{ color: string }`), with no zod runtime dependency.
- **Shared codegen templates** (`tools/openapi-codegen/templates/{field,story,mapi-story}.ts`) carry the resolution logic; all four consumers regenerated (`schema`, `api-client`, `management-api-client`, `live-preview`).

Fixes WDX-454

### DIFF
--- a/packages/schema/src/field-plugins/index.ts
+++ b/packages/schema/src/field-plugins/index.ts
@@ -1,0 +1,8 @@
+/**
+ * Official Storyblok field plugins for `@storyblok/schema`. Each plugin is a
+ * {@link defineFieldPlugin} result backed by a hand-written Standard Schema
+ * validator; import individually so unused plugins tree-shake away.
+ */
+
+export { storyblokColorField } from './storyblok-color-field';
+export type { StoryblokColorFieldValue } from './storyblok-color-field';
--- a/packages/capi-client/playground/integration-tests/test/types/stories.test-d.ts
+++ b/packages/capi-client/playground/integration-tests/test/types/stories.test-d.ts
@@ -1,5 +1,6 @@
 import { describe, expectTypeOf, it } from 'vitest';
-import { defineBlock, defineField } from '@storyblok/schema';
+import { defineBlock, defineField, defineSchema, type Schema as InferSchema, type PluginFieldValue } from '@storyblok/schema';
+import { storyblokColorField } from '@storyblok/schema/field-plugins';
 
 import { createApiClient, type Story } from '@storyblok/api-client';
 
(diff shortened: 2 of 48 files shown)


packages/schema/src/field-plugins/index.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Official Storyblok field plugins for `@storyblok/schema`. Each plugin is a
+ * {@link defineFieldPlugin} result backed by a hand-written Standard Schema
+ * validator; import individually so unused plugins tree-shake away.
+ */
+
+export { storyblokColorField } from './storyblok-color-field';
```
maoberlehner, line 7: Should we provide "official" field plugins at all? Should we provide more right at the start or okay to extend in the future?

dipankarmaikap: I think we should keep it. It also serves as an example, and users can copy the `storyblok-color-field.ts` pattern for their own custom plugins.

dipankarmaikap (commented):
I like the API. It feels consistent with the existing `defineFieldPlugin` API.

One small thing to consider: should we warn when a `field_type` is registered but never used, or when a custom field references an unregistered `field_type`?

maoberlehner:
> I like the API. It feels consistent with the existing `defineFieldPlugin` API.
> 
> One small thing to consider: should we warn when a `field_type` is registered but never used, or when a custom field references an unregistered `field_type`?

@dipankarmaikap I've added a check for unregistered `field_type`s but I don't think a warning for never used is necessary as it does no harm.

dipankarmaikap (approved):
Looks good to merge from my side.

PR #667, opened 2026-07-01
feat(cli): add schema diff command with unified diff model

maoberlehner:
Adds `storyblok schema diff --from <space-id|path> --to <space-id|path>`, a read-only command that exposes the schema classification `schema push` computes internally. Each side auto-resolves a numeric space ID (fetched remotely) or a file path (loaded via jiti); authentication is required only when a side is a space.

The driving use case is translating schema changes between spaces: compute what changed, then run downstream operations from that diff.

## How

- **Unified diff model.** `diffSchema` is now symmetric over a new `NormalizedSchema` (identity-keyed maps) that both a space and a local file resolve to. Every `EntityDiff` carries field-level `FieldChange[]` (added/removed/modified with before/after) plus raw before/after snapshots, replacing the old text patch. This single model backs both `push` and the new command.
- **`push` renders from the structured model.** `formatDiffOutput` uses the shared `renderFieldChanges` helper instead of `createTwoFilesPatch`; push wording and behavior are preserved. The `create|update|unchanged|stale` action vocabulary is kept internally so `migrations/analyze.ts` and changeset code need no change.
- **Shared code relocated** per the no-sibling-imports rule: `diff-schema.ts` and `load-schema.ts` move from `push/` up to the parent `schema/` directory. `cleanComponent`/`cleanDatasource` are extracted in `serialize.ts` so diffing reuses the exact push normalization.
- **Output.** The diff document goes to stdout and the surrounding chrome to stderr, so `schema diff --from a --to b > changes.diff` captures the diff and nothing else; colors drop themselves when stdout is not a terminal. Entities are reported as added/changed/removed relative to `--to`, with per-entity field changes, sorted by name so two runs agree. Blocks whose group membership neither side can express are listed under the summary, since a schema file manages a block's group only when the block declares a `folder` and the folder itself still shows up as added or removed. No report file, and no output or format flags: a machine-readable mode belongs on the same `--format pretty|json` the validate commands already use, and that convergence is its own change.
- **Field-level granularity.** Nested records are compared property by property, so a change reads `schema.body.maximum` rather than a dump of the enclosing object. A list whose members are identified by `name` — datasource dimensions — is compared member by member (`dimensions.fr.entry_value`); when such a list differs only in order, the field is reported whole, so a change is never listed without saying what it was. Values over 200 characters are elided at render time.

## Identity

A diff is only meaningful if both sides are compared in the same identity space, and several values are per-space ids that a schema file cannot hold. So:

- Blocks and datasources are keyed by `name`, matching how `schema push` already identifies them. Renaming a datasource therefore reads as an add plus a remove rather than as one change. Keying datasources by `slug` would fix that, but it changes how `push` and `rollback` resolve an entity and how changeset entries are written, so it is filed separately rather than carried by a read-only command.
- Component-group membership is compared as slug paths, translated from the per-space group uuids, and block tags are compared by name, translated from the per-space tag ids. Two spaces that hold the same schema under different ids read as unchanged.
- The raw `component_group_uuid` escape hatch is compared only when the target is a schema file, since that is the one case where the value is authored rather than assigned.

Fixes WDX-469


packages/cli/src/commands/schema/diff/index.ts
```diff
@@ -0,0 +1,82 @@
...
+import { diffSchema } from "../diff-schema";
+import type { NormalizedSchema } from "../types";
+import { buildDiffReport, formatSchemaDiff, isSpaceRef, resolveSource } from "./actions";
+import type { SchemaDiffReport } from "./actions";
+
+schemaCommand
+  .command("diff")
+  .description("Diff two schemas (space IDs or local entry files) and report what changed")
+  .requiredOption(
+    "--from <source>",
```
alexjoverm: (minor, non-blocking) - not sure what's the standard syntax, but as can be both a `space` or a `path`, should we rather document it along the lines of `--from/--to <space id | path>`?

packages/cli/src/commands/schema/format-diff.ts
```diff
@@ -0,0 +1,117 @@
...
+  ["Datasources", "datasource"],
+];
+
+/**
+ * Renders a {@link DiffResult} for terminal display: a `+/~/-/=` icon and action
+ * tag per entity, field-level changes, and a colored summary line. Wording is
+ * supplied via {@link FormatDiffOptions} so both `schema push` and `schema diff`
+ * share one layout while keeping their own vocabulary.
+ */
+export function formatDiff(result: DiffResult, options: FormatDiffOptions): string {
```
alexjoverm, line 95: future note (no action required): we can adapt later this function (after merging `find`) to use pipes. In that way, if it's tty, will pretty print as it is now, otherwise will print it to stdout so users can do `schema diff --from x --to y > myfile.diff`

packages/cli/src/commands/schema/diff/index.ts
```diff
@@ -0,0 +1,82 @@
...
+    "Base schema to compare against: a space ID or a path to a schema entry file",
+  )
+  .requiredOption("--to <source>", "Target schema: a space ID or a path to a schema entry file")
```
alexjoverm: (blocking): @maoberlehner how does the usage of `space` id work in this context? When I run `storyblok schema init -s X` it gets created without a space id... 

I think that should be consistant accross the schema commands, or we keep just path?

Image

alexjoverm: is it possible that by using `space id` it checks from the live space? If that's the case, I feel unsure about it, mostly because:
- other commands take the space id to look into `.storyblok/<command>/<space-id>` so in a way we're introducing a different UX that can confuse users (just happened to me as well)
- as the intention of `schema` is to work on local files... wouldn't that go against it?

That said, it def is valuable for cases where they want to check differences between 2 spaces by just using a command. 

I would imagine that, with the introduction of schema, a number of users would organize their components in a git-based repo with branches and so. Maybe we can allow the command to accept a url, so in that way they can compare quickly 2 branches? like `--from https://github.com/<repo>/<branch>/<path-to->schema.ts` just to say an example


(I'm brainstorming here - take this as rough ideas but feel totally free to take what you consider the best)

packages/cli/src/commands/schema/diff/index.ts
```diff
@@ -0,0 +1,82 @@
...
+      ui.log(formatSchemaDiff(diffResult, from, to));
+
+      summary.total = diffResult.diffs.length;
+      summary.succeeded = summary.total;
+    } catch (maybeError) {
+      summary.failed += 1;
+      handleError(toError(maybeError), verbose);
+    } finally {
+      logger.info("Schema diff finished", { summary });
+      reporter.addSummary("schemaDiffResults", summary);
```
alexjoverm: (non-blocking): do you see value in this command generating reports? I see it more valuable for bulk operations

packages/cli/src/commands/schema/diff-schema.ts
```diff
@@ -0,0 +1,320 @@
...
+  return isRecord(value) ? value : {};
+}
+
+/**
+ * Like {@link diffKeyed}, but recurses into nested records so a change reads as
+ * the property that actually moved (`schema.body.maximum`) instead of two dumps
+ * of the whole enclosing object. Recursion stops at non-record values, which are
+ * reported whole.
+ */
+function diffKeyedDeep(
```
alexjoverm, line 149: claude finding:

diffKeyedDeep stops at non-record values, so any array is reported whole. Reproduced with a 60-entry datasource where exactly one entry's value changed:

Datasources
  ~ Big (changed)
    ~ entries
      - [{"name":"E0","value":"v0"},{"name":"E1",...   (~2.5 KB, one line)
      + [{"name":"E0","value":"v0"},{"name":"E1",...   (~2.5 KB, one line)

On main, createTwoFilesPatch over the multi-line formatValue output plus the +/- line filter showed only the changed entry. Real datasources run to thousands of entries, so this can emit megabyte-long terminal lines. It affects schema push too, not just the new command.

Suggestion: for arrays, either index into them in diffKeyedDeep (entries[30].value) or truncate in inlineValue with an elision marker and leave the full values to meta.diff.

alexjoverm (changes requested):

packages/cli/src/commands/schema/diff/index.ts
maoberlehner: Typically, the schema is independent of a space. E.g., a typical use case might be to have a dev space and a prod space. First, you iterate on your schema on dev, and later you push it to prod, or you might even have a separate space per feature branch.

Therefore, no space ID: the schema is not tied to a space.

It might look like an inconsistency, but consider that `.storyblok/schema` is just a default. Typically, you'd like to have your schema in `src` alongside your other code. In this context, it does not look like an inconsistency anymore, does it?

Maybe we should remove the default and force the user to specify a path. No strong opinion about it, but personally I try to default to a default value if we can provide one.

> is it possible that by using space id it checks from the live space?

Yes, that's kinda the whole point of the PR. If you want to diff two local schemas (which isn't a typical use case to begin with), you could just use `git diff` (the command can compare two local schemas too, but only because it was basically free to support it).

So the main use case is to show you a diff of your local changes to what is currently live on some space. E.g., before pushing your schema to that space (Imagine multiple people sharing a single staging environment. With the diff command, you can check what changes somebody else pushed that your push would overwrite). Also see: https://discord.com/channels/700316478792138842/1057984028088737802/1542517736603648062

alexjoverm: With that main use case in mind, I think it's great as you designed it as it covers it perfectly but allows some extra flexibility

packages/cli/src/commands/schema/diff/index.ts
maoberlehner: Done `--from <space-id|path>` / `--to <space-id|path>`.

packages/cli/src/commands/schema/format-diff.ts
maoberlehner: Done

packages/cli/src/commands/schema/diff/index.ts
maoberlehner: Agreed, not very valuable, dropped.

packages/cli/src/commands/schema/diff-schema.ts
maoberlehner: Improved.

alexjoverm (approved):

PR #672, opened 2026-07-03
fix(cli,mapi-client): preserve image dimensions in asset URLs on push

maoberlehner:
Assets migrated between spaces with `storyblok assets push` lost the `<width>x<height>` dimensions folder in their CDN URL (e.g. `.../f/SPACE/2048x1820/hash/image.jpg` became `.../f/SPACE/hash/image.jpg`), even when the source asset had it.

Root cause: Storyblok's backend only appends the dimensions folder to the S3 key when the sign request (`POST /v1/spaces/{id}/assets`) includes a `size` field — it does not derive it from the actual image. Two gaps combined to drop it:

- `@storyblok/management-api-client`'s `assets.create()` explicitly whitelisted only `short_filename`, `asset_folder_id`, `is_private` when calling `upload()`, silently discarding `size` (and `validate_upload`) even though both are part of its declared input type.
- The CLI's `assets push` never derived a `size` for pushed assets in the first place.

## Changes

- `packages/mapi-client/src/resources/assets.ts`: `create()` now forwards `size`/`validate_upload` to the sign request.
- `packages/cli/src/commands/assets/streams.ts`: when creating a pushed asset, reuse the dimensions already present in the source asset's filename URL (parsed via a new `extractAssetSizeFromFilename` helper) so pushed assets keep them.
- Added regression tests in both packages.

## Test plan

- [x] `pnpm nx run @storyblok/management-api-client:test` — 104 tests pass, including new `assets.create()` size-forwarding test
- [x] `pnpm nx run storyblok:test` — 726+ tests pass, including new push-command regression tests
- [x] `pnpm nx run storyblok:test:types` / `@storyblok/management-api-client:typecheck` — clean
- [x] Lint clean on both packages
- [x] Manually reproduced end-to-end against seeded QA spaces: created a UI-style asset with `size=2048x1820`, pulled and pushed it with the built CLI — confirmed the target asset's filename retains `2048x1820` (previously dropped before this fix)

Fixes WDX-480

### DIFF
--- a/packages/cli/src/commands/assets/streams.ts
+++ b/packages/cli/src/commands/assets/streams.ts
@@ -650,15 +650,8 @@ export const makeCleanupAssetFSTransport = (): CleanupAssetTransport =>
 const hasId = (a: unknown): a is { id: number } => {
   return !!a && typeof a === 'object' && 'id' in a && typeof (a as any).id === 'number';
 };
-const hasShortFilename = (a: unknown): a is { short_filename: string } => {
-  return !!a && typeof a === 'object' && 'short_filename' in a && typeof (a as any).short_filename === 'string';
-};
-const hasFilename = (a: unknown): a is { filename: string } => {
-  return !!a && typeof a === 'object' && 'filename' in a && typeof (a as any).filename === 'string';
-};
-const hasSize = (a: unknown): a is { size: string } => {
-  return !!a && typeof a === 'object' && 'size' in a && typeof (a as any).size === 'string';
-};
+const hasProp = <K extends string>(a: unknown, key: K): a is Record<K, string> =>
+  !!a && typeof a === 'object' && key in a && typeof (a as any)[key] === 'string';
 
 const processAsset = async ({
   localAsset,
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/assets/streams.ts
```diff
@@ -653,6 +653,12 @@ const hasId = (a: unknown): a is { id: number } => {
 const hasShortFilename = (a: unknown): a is { short_filename: string } => {
   return !!a && typeof a === 'object' && 'short_filename' in a && typeof (a as any).short_filename === 'string';
 };
+const hasFilename = (a: unknown): a is { filename: string } => {
+  return !!a && typeof a === 'object' && 'filename' in a && typeof (a as any).filename === 'string';
+};
+const hasSize = (a: unknown): a is { size: string } => {
+  return !!a && typeof a === 'object' && 'size' in a && typeof (a as any).size === 'string';
+};
```
alexjoverm: suggestion(non-blocking): 

```
const hasProp = <K extends string>(a: unknown, key: K): a is Record<K, string> =>
    !!a && typeof a === 'object' && key in a && typeof (a as any)[key] === 'string';

// then use it directly
  if (hasProp(a, 'short_filename')) { /* a.short_filename is string */ }
  if (hasProp(a, 'filename')) { /* a.filename is string */ }

// or re-write functions
const hasFilename = (a: unknown) => hasProp(a, 'short_filename')
```

alexjoverm (approved):

PR #675, opened 2026-07-07
feat(cli): selective components pull/push by filter, group, and tag (DX-273)

maoberlehner:
Adds selective sync to `components pull` and `components push` so teams sharing a single space can sync only their own components.

- `components pull` gains `--filter <glob>`, `--group <name|Parent/Child path>` (repeatable, includes descendant groups), and `--tag <name>` (repeatable, comma-separated).
- `components push` gains `--group` and `--tag` with the same semantics.
- Selectors combine with AND across types and OR within a repeatable selector. `--filter` is a `minimatch` glob applied on top.

Fixes DX-273


adr/0005-components-selective-pull-dependency-semantics.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Client-side filtering over server-side query parameters
+
+Selectors filter the full fetched or loaded component set in memory, rather than requesting a pre-filtered set from the API through `in_group` or `with_tags` query parameters. The dependency collector needs to see the complete component, group, and tag data to trace which groups, tags, and ancestors a matched component depends on. Filtering server-side would return only the matched components and leave the collector without the data it needs to resolve dependencies correctly.
+
+## Consequences
+
+- `pull` and `push` selectors are symmetric: the same `--filter`, `--group`, and `--tag` flags mean the same thing on both commands.
+- `push --filter` no longer expands to include sibling components referenced via `component_whitelist`; previously matched components pulled in their whitelisted siblings implicitly, and that behavior is now scoped to the option A dependency set.
```
maoberlehner, line 41: Are we okay with that change? IMO there was no good reason why we pushed whitelisted components when using filter, might even have been unexpected behavior by many.

adr/0005-components-selective-pull-dependency-semantics.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Context
+
+Teams that share a single Storyblok space often only want to sync the components they own, not the entire schema. `components pull` gained `--filter <glob>`, `--group <name|Parent/Child path>` (repeatable, and includes descendant groups), and `--tag <name>` (repeatable and comma-separated), and `components push` gained the matching `--group` and `--tag` selectors alongside its existing `--filter`. Selectors combine with AND across different selector types and OR within a repeatable selector's values.
+
+Selecting a subset of components raises a dependency question: when a component is selected, which of its related records (groups, tags, presets, and other components referenced from its schema) should be pulled or pushed alongside it? Components can reference groups and tags by ID (as their assigned group and tags) and can also reference groups and tags from inside field-level whitelists (`component_group_whitelist`, `component_tag_whitelist`), and they can reference other components by name through `component_whitelist`.
+
+## Decision
+
+### Dependency semantics (option A)
```
maoberlehner: @claude: Option A made sense during discussions but people reading this later have no idea what option B would've been.

packages/cli/src/commands/components/pull/index.ts
```diff
@@ -12,13 +12,17 @@ import { resolveCommandPath } from '../../../utils/filesystem';
...
+import { filterSpaceData, resolveGroupSelector, resolveTagSelector } from '../utils';
 
 const pullCmd = componentsCommand
   .command('pull [componentName]')
   .option('-f, --filename <filename>', 'custom name to be used in file(s) name instead of space id')
   .option('--sf, --separate-files', 'Argument to create a single file for each component')
   .option('--su, --suffix <suffix>', 'suffix to add to the file name (e.g. components.<suffix>.json)')
   .option('-s, --space <space>', 'space ID')
+  .option('--fi, --filter <filter>', 'glob pattern to select components by name')
+  .option('--gr, --group <group>', 'component group name or Parent/Child path (repeatable, includes descendants)', (value: string, previous: string[] = []) => [...previous, value])
```
maoberlehner: @claude: what does Parent/Child path mean?

packages/cli/src/commands/components/pull/README.md
```diff
@@ -48,6 +48,43 @@ This will download a single component and its dependencies to:
 | `--sf, --separate-files` | Create a separate file for each component | `false` |
 | `--su, --suffix <suffix>` | Suffix to add to the files names  | |
 | `-p, --path <path>` | Custom path to store the files | `.storyblok/components` |
+| `--fi, --filter <glob>` | Glob pattern to select components by name | - |
+| `--gr, --group <name>` | Component group name or `Parent/Child` path to select. Repeatable | - |
+| `--tg, --tag <name>` | Component tag name to select. Repeatable and comma-separated | - |
+
+## Selective pull
```
maoberlehner: @claude: remove this new section and add a new branch/pr to storyblok-docs-platform instead.

packages/cli/src/commands/components/push/README.md
```diff
@@ -54,10 +54,32 @@ This will upload a single component and its dependencies from:
...
 | `--fi, --filter <filter>` | Glob pattern to filter components by their name (e.g., "hero*" will match all components starting with "hero") | -                       |
+| `--gr, --group <name>`    | Component group name or `Parent/Child` path to select. Repeatable                                              | -                       |
+| `--tg, --tag <name>`      | Component tag name to select. Repeatable and comma-separated                                                   | -                       |
 | `--sf, --separate-files`  | Read from separate files instead of consolidated files                                                         | `false`                 |
 | `--su, --suffix <suffix>` | Suffix to add to the files names                                                                               | -                       |
 | `-p, --path <path>`       | Custom path to read the files from                                                                             | `.storyblok/components` |
 
+## Selective push
```
maoberlehner: @claude: same as above.

packages/cli/src/commands/components/README.md
```diff
@@ -9,3 +9,7 @@ The `components` module provides tools to manage Storyblok components and their
 - `delete` (coming soon): Remove components from your Storyblok space.
 
 > See each subcommand for detailed usage, options, and examples.
+
+## Selective sync
```
maoberlehner: @claude: same as above

adr/0005-components-selective-pull-dependency-semantics.md
maoberlehner: Good call. Added an "Options considered" section to the ADR in e1647be6d that lays out all three:

- **Option A (chosen):** matches plus their own direct dependencies, minus sibling components.
- **Option B:** matches plus their *full transitive* dependencies, following `component_whitelist` into sibling components recursively. This is what `push --filter` did before. Rejected because `component_whitelist` is name-based and self-heals, so siblings aren't needed for a valid push, and following it expands the selection outward until it reaches most of the space, which defeats a team syncing only what it owns.
- **Option C:** literal matches only, no dependencies. Rejected because group/tag references are ID-based and can't self-heal, so a cross-space push would leave them dangling (and a nested group without its ancestors 422s).

Also renamed the decision heading from "Dependency semantics (option A)" to just "Dependency semantics" so it reads in context.

packages/cli/src/commands/components/pull/index.ts
maoberlehner: A group path is a slash-separated chain of nested group names, e.g. `Checkout/Payment` means the `Payment` group nested under `Checkout`. You use it instead of a bare name when the same group name appears in more than one place in the tree and a bare name would be ambiguous. Reworded the `--group` help text on both pull and push in e1647be6d to spell this out: `component group to select by name (e.g. "Checkout"), or by a slash-separated path of nested group names to disambiguate (e.g. "Checkout/Payment")`.

packages/cli/src/commands/components/pull/README.md
maoberlehner: Done in e1647be6d: removed the "Selective pull" prose section (kept the terse flag rows in the options table, consistent with every other flag). Documented the selectors on the docs platform instead: storyblok/storyblok-docs-platform#570.

packages/cli/src/commands/components/push/README.md
maoberlehner: Done in e1647be6d: removed the "Selective push" section. Covered on the docs platform in storyblok/storyblok-docs-platform#570.

packages/cli/src/commands/components/README.md
maoberlehner: Done in e1647be6d: removed the "Selective sync" section. Covered on the docs platform in storyblok/storyblok-docs-platform#570.

packages/cli/src/commands/components/utils.ts
```diff
@@ -0,0 +1,242 @@
...
+          requiredTagIds.add(numericTagId);
+        }
+      });
+    }
+
+    if (component.schema) {
+      const schemaDeps = collectWhitelistDependencies(component.schema);
+      schemaDeps.groupUuids.forEach(groupUuid => requiredGroupUuids.add(groupUuid));
+      schemaDeps.tagIds.forEach(tagId => requiredTagIds.add(tagId));
+      // Option A: do NOT add schemaDeps.componentNames. Sibling components are not pulled in.
```
maoberlehner: @claude: not mention option A here, useless info without context.

packages/cli/src/commands/components/utils.ts
```diff
@@ -0,0 +1,242 @@
...
+        if (!Number.isNaN(numericTagId)) {
+          requiredTagIds.add(numericTagId);
+        }
+      });
+    }
+
+    if (component.schema) {
+      const schemaDeps = collectWhitelistDependencies(component.schema);
+      schemaDeps.groupUuids.forEach(groupUuid => requiredGroupUuids.add(groupUuid));
+      schemaDeps.tagIds.forEach(tagId => requiredTagIds.add(tagId));
```
maoberlehner, line 51: @claude: why do we need groupUuids and tagIds from whitelisted dependencies?

adr/0005-components-selective-pull-dependency-semantics.md
alexjoverm: I agree on the behaviour. Just a note - this might be perceived as a regression, or a fix, somewhere between that line (depending on how users were relying on it). But I see it low risk

adr/0005-components-selective-pull-dependency-semantics.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Consequences
+
+- `pull` and `push` selectors are symmetric: the same `--filter`, `--group`, and `--tag` flags mean the same thing on both commands.
+- `push --filter` no longer expands to include sibling components referenced via `component_whitelist`; previously matched components pulled in their whitelisted siblings implicitly, and that behavior is now scoped to the option A dependency set.
+- A cross-space push of a selected slice can leave harmless dangling `component_whitelist` names that reference components not yet present in the target space. These self-heal once the referenced components are pushed, since the whitelist match is name-based and requires no remapping.
+- Pull and push always fetch or load the full component, group, and tag data before filtering, so selective sync does not reduce the amount of data read from disk or from the API; it only reduces what gets written or uploaded.
+- The single-component `push <componentName>` path also no longer carries whitelisted sibling components through `component_whitelist`, for the same option A rationale: name-based references self-heal once the referenced component is pushed.
```
alexjoverm, line 44: Same here (maybe we should mention this behaviour change in the pkg reference?

alexjoverm (approved):

adr/0005-components-selective-pull-dependency-semantics.md
maoberlehner: Added a note to the docs PR.

PR #676, opened 2026-07-07
fix(cli): escape backslashes and newlines in schema init string values

### LINKED ISSUE #671: [Bug]: `schema init` doesn't escape string values — multiline strings break parsing, regex backslashes are silently lost
### Package

storyblok (CLI)

### Bug Description

`storyblok schema init` (introduced in `storyblok@4.19.0-alpha.0`) writes field option string values into single-quoted string literals without escaping anything except `'`. Two real-world consequences we hit on the very first bootstrap of an existing space:

1. **Multiline strings break the parse.** A field `description` containing a newline is emitted with a raw line break inside a single-quoted literal → the generated file is not valid TypeScript at all.
2. **Backslashes are silently swallowed — data corruption on push.** A `text` field with a validation `regex` of `^\s*\b\w+(?:\s+\w+){0,3}\s*$` is emitted as `regex: '^\s*\b\w+(?:\s+\w+){0,3}\s*$'`. That parses fine, but in a JS string `\s` is just `s` and `\b` is a backspace control character, so the value silently becomes `^s*[BS]w+(?:s+w+){0,3}s*$`. A subsequent `schema push` would overwrite the correct remote regex with the corrupted one. This one is nasty precisely because nothing errors — we only caught it because `--dry-run` showed an unexpected diff on a component we hadn't touched.

Root cause is the string branch of `formatValue()` in `packages/cli/src/commands/schema/utils.ts` (used by `init/generate-code.ts`), which only escapes single quotes. It should escape at least `\`, newlines, and control characters (or use `JSON.stringify`-style escaping).

### Steps to Reproduce

1. In a Storyblok space, create a component with:
   - a field whose `description` contains a line break, and
   - a `text` field with a validation regex containing `\s` / `\w` / `\b`.
2. Run `storyblok schema init --space <SPACE_ID> --out-dir storyblok/schema`.
3. Run `storyblok schema push storyblok/schema/schema.ts --space <SPACE_ID> --dry-run`.

### Reproduction URL

_Reproducible with any space matching the field setup in the steps above._

### Expected Behavior

- Generated files always parse.
- `schema init` → `schema push --dry-run` on an untouched space round-trips with no value-level diffs; string values (including regexes) survive byte-for-byte.

### Actual Behavior

Multiline description (component `vimeo-video`):

```ts
defineField('videoId', {
    description: 'The identifier can be found in the regular vimeo share link:
https://vimeo.com/<This is the Identifier>',
```

→ `ParseError: Unterminated string constant.`

Regex (component `fullscreen-cta-blok`), after fixing the parse error above, `--dry-run` shows the silent corruption:

```
~ fullscreen-cta-blok (update)
  -      regex: '^\s*\b\w+(?:\s+\w+){0,3}\s*$',
  +      regex: '^s*w+(?:s+w+){0,3}s*$',
```

### Environment

```shell
  System:
    OS: macOS 26.3.1
    CPU: (10) arm64 Apple M1 Pro
  Binaries:
    Node: 24.13.1
    Yarn: 4.1.1
    npm: 11.8.0
  npmPackages:
    @storyblok/nuxt: ^11.0.1 => 11.0.1
    @storyblok/schema: ^0.1.0-alpha.0 => 0.1.0-alpha.0
    nuxt: ^4.4.8 => 4.4.8
    storyblok: 4.19.0-alpha.0 => 4.19.0-alpha.0
    vue: ^3.5.39 => 3.5.39
```

### Error Logs

```shell
 Schema  Pushing schema...

✖ Failed to resolve schema
 Error

▲ error ParseError: Unterminated string constant.
 …/storyblok/schema/components/vimeo-video.ts:22:21
```

### Additional Context

Workaround: hand-escaped the generated values (`\n` for the newline, `\\s` etc. for the regex) — after that, the affected components round-trip as `unchanged` in `--dry-run`. Related (same command, different root cause): component group names containing `&` generate invalid identifiers — filed separately.

maoberlehner:
`storyblok schema init` (introduced in `storyblok@4.19.0-alpha.0`) serialized field option strings — and field names — into single-quoted TypeScript literals escaping **only** `'`. Backslashes and line terminators passed through raw, causing two real-world failures on the first bootstrap of an existing space:

1. **Multiline strings break the parse.** A `description` containing a newline emits a raw line break inside a single-quoted literal → `ParseError: Unterminated string constant`.
2. **Backslashes silently swallowed — data corruption on push.** A `regex` of `^\s*\b\w+(?:\s+\w+){0,3}\s*$` emits as `regex: '^\s*\b\w+...'`; that parses, but `\s`→`s` and `\b`→backspace, so the value silently becomes `^s*w+...`. A later `schema push` overwrites the correct remote regex with the corrupted one — nothing errors.

Root cause: the string branch of `formatValue()` in `packages/cli/src/commands/schema/utils.ts` (and the field-name escaping in `init/generate-code.ts`).

## Fix

Add a shared `quoteString()` helper that uses `JSON.stringify` for correct backslash/control-char/newline escaping while preserving the existing single-quote output style, and reuse it in both `formatValue` and the `defineField('name', …)` field name.

Because `serialize.ts` (push diffing) also uses `formatValue`, one fix makes `init` output and `push --dry-run` round-trip byte-for-byte — string values, including regexes, survive intact.

## Notes

- Users who already ran `init` and hand-fixed files are unaffected; only new bootstraps change.
- The separate `&`-in-group-name identifier bug (DX-478) is tracked independently and out of scope here.

Fixes #671

### DIFF
--- a/packages/cli/src/commands/schema/init/generate-code.test.ts
+++ b/packages/cli/src/commands/schema/init/generate-code.test.ts
@@ -44,6 +44,49 @@ describe('generateComponentFile', () => {
     expect(result).not.toContain('created_at');
   });
 
+  it('should escape backslashes and newlines in field values so the output round-trips', () => {
+    const component = {
+      id: 1,
+      name: 'vimeo-video',
+      created_at: '',
+      updated_at: '',
+      description: 'The identifier can be found here:\nhttps://vimeo.com/<id>',
+      schema: {
+        videoId: {
+          type: 'text',
+          pos: 0,
+          regex: String.raw`^\s*\b\w+(?:\s+\w+){0,3}\s*$`,
+        },
+      },
+    };
+
+    const result = generateComponentFile(component as any);
+
+    // Backslashes are escaped so the regex survives byte-for-byte (was silently
+    // corrupted to `^s*w+...` when only single quotes were escaped).
+    expect(result).toContain(String.raw`regex: '^\\s*\\b\\w+(?:\\s+\\w+){0,3}\\s*$',`);
+    // The multiline description is escaped instead of breaking the parse.
+    expect(result).toContain(String.raw`description: 'The identifier can be found here:\nhttps://vimeo.com/<id>',`);
+    // No raw line break leaks into a string literal.
+    expect(result).not.toContain('found here:\n');
+  });
+
+  it('should escape single quotes in field names', () => {
+    const component = {
+      id: 1,
+      name: 'page',
+      created_at: '',
+      updated_at: '',
+      schema: {
+        'it\'s': { type: 'text', pos: 0 },
+      },
+    };
+
+    const result = generateComponentFile(component as any);
+
+    expect(result).toContain('defineField(\'it\\\'s\', {');
+  });
+
   it('should reverse-map wire reference keys to the DSL form', () => {
     const component = {
       id: 1,
--- a/packages/cli/src/commands/schema/init/generate-code.ts
+++ b/packages/cli/src/commands/schema/init/generate-code.ts
@@ -4,6 +4,7 @@ import {
   DATASOURCE_STRIP_KEYS,
   formatValue,
   INDENT,
+  quoteString,
   stripKeys,
 } from '../utils';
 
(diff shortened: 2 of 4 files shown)


dipankarmaikap (approved):
LGTM!

PR #678, opened 2026-07-07
fix(cli): drive single-option empty type by required, not UI setting

### LINKED ISSUE #668: [Bug]: Wrong single-option field types
### Package

storyblok (CLI)

### Bug Description

We had a runtime issue, because a single-option field returned `""`, although the generated types indicated that this is not a possible state.

That lead me to test how the config for the single-option field affects the UI and the types. The crucial options were "Required" and "Hide empty option". Especially the behavior of the later one was not clear to me.

### Steps to Reproduce

1. Create a component with eight fields:
    ```
    name: first
    options: { One: 'one', Two: 'two' }
    required: true
    hide empty option: false
    
    name: second
    options: { One: 'one', Two: 'two' }
    required: true
    hide empty option: true
    
    name: third
    options: { One: 'one', Two: 'two' }
    required: false
    hide empty option: false
    
    name: fourth
    options: { One: 'one', Two: 'two' }
    required: false
    hide empty option: true
    
    name: fifth
    options: { One: 'one', Two: 'two', Empty: '' }
    required: true
    hide empty option: false
    
    name: sixth
    options: { One: 'one', Two: 'two', Empty: '' }
    required: true
    hide empty option: true
    
    name: seventh
    options: { One: 'one', Two: 'two', Empty: '' }
    required: false
    hide empty option: false
    
    name: eighth
    options: { One: 'one', Two: 'two', Empty: '' }
    required: false
    hide empty option: true
    ```
2. Generate types with the storyblok CLI

### Reproduction URL

_No response_

### Expected Behavior

Generate types, which reflect the possible states of the JSON payload. I would expect types like this:

```ts
export interface MyComponent {
  first: "one" | "two";
  second: "one" | "two";
  third: "one" | "two" | "";
  fourth: "one" | "two" | "";
  fifth: "one" | "two";
  sixth: "one" | "two";
  seventh: "one" | "two" | "";
  eighth: "one" | "two" | "";
  component: "HnnngTest";
  _uid: string;
}
```

### Actual Behavior

The generated types look like this:

```ts
export interface MyComponent {
  first: "" | "one" | "two";
  second: "one" | "two";
  third?: "" | "one" | "two";
  fourth?: "one" | "two";
  fifth: "" | "one" | "two";
  sixth: "one" | "two" | "";
  seventh?: "" | "one" | "two";
  eighth?: "one" | "two" | "";
  component: "HnnngTest";
  _uid: string;
}
```

This can lead to runtime issues, if you rely on the types being correct, especially with `??` for fallbacks.

```ts
// Type is ('one' | 'two'), but the value is ''
const fieldValue = third.third ?? 'one';
```

The key problems are:
- "Hide empty option" only affects the UI, so it should not have any impact on the types
- "Required" does not allow empty strings as valid field values, so you would never be able to save `""` as a value. It does not matter if there is a manually created empty-option, because the option is selected it behaves like clearing the field.
- Fields are never omitted from the json, when a field is empty while saving the story, it will always be set to `""`. My guess is that only conditional fields might be truly optional, but I was not able to check that, because I've used a fresh space on the free-plan for testing this, which does not support conditional fields.

### Environment

```shell
System:
    OS: macOS 26.5.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 64.48 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 26.4.0 - /Users/me/.local/share/mise/installs/node/latest/bin/node
    npm: 11.17.0 - /Users/me/.local/share/mise/installs/node/latest/bin/npm
    pnpm: 11.9.0 - /Users/me/.local/share/mise/installs/pnpm/11.9.0/pnpm
    bun: 1.3.14 - /Users/me/.local/share/mise/installs/bun/latest/bin/bun
  Browsers:
    Brave Browser: 141.1.83.112
    Chrome: 149.0.7827.201
    Firefox: 151.0.4
    Safari: 26.5
  npmPackages:
    storyblok: ^4.18.2 => 4.18.2
```

### Error Logs

```shell

```

### Additional Context

With this config, I only see one workaround for generating types that would let you avoid runtime errors: You would need to never add an empty option manually and misuse "Hide empty option" to control if `""` should be added to the types or not by keeping it in sync with the "Required" config.

maoberlehner:
Single/multi-option field type generation added the `""` union member based on `exclude_empty_option` ("Hide empty option") — a UI-only setting that never affects the stored payload. Whether `""` is a reachable value is actually determined by `required`.

This drives the empty-string member by `required` instead:

- `required: true` → `"one" | "two"` (never `""`)
- `required: false` → `"one" | "two" | ""` (clearable to `""`)

Any manually-added empty option is normalized out first, then `""` is re-added only when the field is not required. `exclude_empty_option` no longer influences the generated types.

## Reproduction (from the issue)

The 8-field matrix in DX-475 now matches the reporter's expected output for the value union in every case.

| field | config | before | after |
|---|---|---|---|
| `first` | req | `"" \| "one" \| "two"` | `"one" \| "two"` |
| `second` | req, hide | `"one" \| "two"` | `"one" \| "two"` |
| `fourth` | opt, hide | `"one" \| "two"` | `"" \| "one" \| "two"` |
| `fifth` | req, +manual "" | `"" \| "one" \| "two"` | `"one" \| "two"` |

## Out of scope (follow-up)

The reporter also flagged that `required: false` emits an **optional** property (`third?:`), but option fields are never omitted from the payload. That `required → optional (?)` mapping is a *global* mechanism affecting every field type, with its own design questions (conditional fields). Left untouched here; needs a separate ticket. After this fix `third` is `third?: "" | "one" | "two"` — the `""` is now correct, the `?` is the follow-up.

Fixes DX-475
Fixes #668

### DIFF
--- a/packages/cli/src/commands/types/generate/actions.test.ts
+++ b/packages/cli/src/commands/types/generate/actions.test.ts
@@ -1046,6 +1046,84 @@ describe('component property type annotations', () => {
   });
 });
 
+describe('single-option empty value', () => {
+  const opt = (values: string[], required: boolean, excludeEmpty: boolean) => ({
+    type: 'option' as const,
+    pos: 0,
+    required,
+    exclude_empty_option: excludeEmpty,
+    options: values.map(value => ({ name: value, value })),
+  });
+
+  const generate = async (schema: Record<string, unknown>) => {
+    const component: Component = {
+      name: 'test_component',
+      display_name: 'Test Component',
+      created_at: '2023-01-01T00:00:00Z',
+      updated_at: '2023-01-01T00:00:00Z',
+      id: 1,
+      schema,
+      internal_tags_list: [],
+      internal_tag_ids: [],
+    } as unknown as Component;
+    const spaceData: SpaceComponentsData = {
+      components: [component],
+      datasources: [],
+    } as unknown as SpaceComponentsData;
+    return (await generateTypes(spaceData, { strict: false })) as string;
+  };
+
+  // "" inclusion must be driven by `required`, never by `exclude_empty_option`
+  // (UI-only "Hide empty option"). Assert on the field's own union, not the
+  // `[k: string]: unknown` index signature (unrelated non-strict-mode behavior).
+
+  it('required field excludes "" regardless of exclude_empty_option', async () => {
+    const result = await generate({
+      first: opt(['one', 'two'], true, false),
+      second: opt(['one', 'two'], true, true),
+    });
+    expect(result).toContain('first: "one" | "two";');
+    expect(result).toContain('second: "one" | "two";');
+    expect(result).not.toContain('first: "" | "one" | "two"');
+  });
+
+  it('non-required field includes "" regardless of exclude_empty_option', async () => {
+    const result = await generate({
+      third: opt(['one', 'two'], false, false),
+      fourth: opt(['one', 'two'], false, true),
+    });
+    expect(result).toContain('third?: "" | "one" | "two";');
+    expect(result).toContain('fourth?: "" | "one" | "two";');
+  });
+
+  it('required field with a manual empty option still excludes ""', async () => {
+    const result = await generate({
+      fifth: opt(['one', 'two', ''], true, false),
+      sixth: opt(['one', 'two', ''], true, true),
+    });
+    expect(result).toContain('fifth: "one" | "two";');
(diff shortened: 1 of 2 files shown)


alexjoverm (approved):
LGTM, thanks Markus!

PR #679, opened 2026-07-07
fix(cli): sanitize schema init generated identifiers and file names

### LINKED ISSUE #670: [Bug]: `schema init` generates invalid TS identifiers for component group names containing "&"
### Package

storyblok (CLI)

### Bug Description

`storyblok schema init` (introduced in `storyblok@4.19.0-alpha.0`) generates syntactically invalid TypeScript when a component group (folder) name contains characters that are not valid in JS identifiers, e.g. `&`.

Our space has component groups named `Interaktion & CTAs`, `Medien & Visuals`, and `Navigation & Struktur`. The generated code contains identifiers like `interaktion&CtasFolder`, which is not a valid identifier, so the whole generated schema fails to parse — both in `tsc` and in a subsequent `storyblok schema push`.

Looking at `packages/cli/src/commands/schema/init/generate-code.ts`, `toCamelCase()` only collapses `[\s_-]+` before uppercasing the next character; any other non-identifier characters (`&`, `+`, `/`, `(`, `)`, digits at the start, emoji, …) pass through into the variable name unchanged. The kebab-cased file names also keep the `&` (e.g. `interaktion-&-ctas.ts`), which is unpleasant for shells/CI, though not strictly invalid.

### Steps to Reproduce

1. In a Storyblok space, create a component group named `Interaktion & CTAs` with at least one component in it.
2. Run `storyblok schema init --space <SPACE_ID> --out-dir storyblok/schema`.
3. Inspect the generated files, or run `storyblok schema push storyblok/schema/schema.ts --space <SPACE_ID> --dry-run`.

### Reproduction URL

_Reproducible with any space containing a component group with `&` in its name (see steps above)._

### Expected Behavior

Generated variable names are sanitized to valid JS identifiers (e.g. `interaktionCtasFolder`), and file names contain only filesystem/shell-safe characters (e.g. `interaktion-ctas.ts`). `schema push` on a freshly generated `schema init` output should always parse.

### Actual Behavior

Generated folder file (`components/folders/interaktion-&-ctas.ts`):

```ts
export const interaktion&CtasFolder = defineBlockFolder({
  name: 'Interaktion & CTAs',
  ...
});
```

and consuming files:

```ts
import { interaktion&CtasFolder } from './folders/interaktion-&-ctas';

export const formBlokBlock = defineBlock({
  component_group_uuid: interaktion&CtasFolder.uuid,
  ...
```

`storyblok schema push … --dry-run` then fails with:

```
▲ error ParseError: Unexpected token, expected ","
 …/storyblok/schema/schema.ts:46:20
```

### Environment

```shell
  System:
    OS: macOS 26.3.1
    CPU: (10) arm64 Apple M1 Pro
  Binaries:
    Node: 24.13.1
    Yarn: 4.1.1
    npm: 11.8.0
  npmPackages:
    @storyblok/nuxt: ^11.0.1 => 11.0.1
    @storyblok/schema: ^0.1.0-alpha.0 => 0.1.0-alpha.0
    nuxt: ^4.4.8 => 4.4.8
    storyblok: 4.19.0-alpha.0 => 4.19.0-alpha.0
    vue: ^3.5.39 => 3.5.39
```

### Error Logs

```shell
 Schema  Pushing schema...

✖ Failed to resolve schema
 Error

▲ error ParseError: Unexpected token, expected ","
 …/storyblok/schema/schema.ts:46:20
```

### Additional Context

We worked around it by manually renaming the identifiers/files after `schema init`. Related (same command, different root cause): string values are not escaped either — filed separately.

maoberlehner:
`storyblok schema init` generated syntactically invalid TypeScript when a component-group or datasource name contained characters that aren't valid in JS identifiers (e.g. `&`). The generated `export const … = defineDatasource(...)`, its `import`, and its schema-object key all derived from a camelCaser that only treated `[\s_-]+` as word boundaries, so a datasource named `Colors & Sizes` produced `export const colors&SizesDatasource = …` — the whole `schema.ts` then failed to parse in `tsc` and in a subsequent `schema push`.

The originally reported trigger (component **group** names) was already resolved by the directory/`slugify` layout; this PR fixes the underlying weakness, which is still live via **datasource** variable names (derived from the arbitrary `datasource.name`).

## How

- **`toCamelCaseIdentifier`** — built on the existing `slugify`, always yields a valid identifier: symbols stripped, leading digit `_`-prefixed, empty result → `_`. Well-formed names are unchanged (`teaser_list → teaserListBlock`, `Categories → categoriesDatasource`).
- **Hardened `toKebabCase`** so generated file names contain only shell-safe `[a-z0-9-]`.
- **`resolveVarNames`** collision dedup (append `2`, `3`, …), threaded index-aligned through `generateComponentFile` / `generateDatasourceFile` / `generateSchemaFile` and `init/actions.ts` so each `export const`, its `import`, and its schema-object key always agree.
- Documented that file names rely on source uniqueness (datasource `slug` / component `name` are unique), so file names are sanitized but not deduped.

Fixes DX-478
Fixes #670

### DIFF
--- a/packages/cli/src/commands/schema/init/generate-code.ts
+++ b/packages/cli/src/commands/schema/init/generate-code.ts
@@ -26,14 +26,18 @@ function toCamelCaseIdentifier(str: string): string {
 }
 
 /**
- * Converts a string to kebab-case.
- * Handles snake_case, camelCase, PascalCase, and space-separated words.
+ * Converts a string to kebab-case, keeping only filesystem/shell-safe
+ * characters. Handles snake_case, camelCase, PascalCase, and space-separated
+ * words; any remaining non-`[a-z0-9-]` characters collapse to a single `-`.
  */
 function toKebabCase(str: string): string {
   return str
     .replace(/[\s_]+/g, '-')
     .replace(/([a-z])([A-Z])/g, '$1-$2')
-    .toLowerCase();
+    .toLowerCase()
+    .replace(/[^a-z0-9-]+/g, '-')
+    .replace(/-{2,}/g, '-')
+    .replace(/^-+|-+$/g, '');
 }
 
 /** Returns the variable name for a component. e.g. `'teaser_list'` -> `'teaserListBlock'` */
--- a/packages/cli/src/commands/schema/init/generate-code.ts
+++ b/packages/cli/src/commands/schema/init/generate-code.ts
@@ -50,6 +50,24 @@ export function datasourceVarName(name: string): string {
   return `${toCamelCaseIdentifier(name)}Datasource`;
 }
 
+/**
+ * Resolves an ordered list of raw names to unique variable names. Names that
+ * sanitize to the same identifier get a numeric suffix (`…2`, `…3`), so the
+ * generated `export const`s and schema-object keys never collide. Index-aligned
+ * to `rawNames`.
+ */
+export function resolveVarNames(rawNames: string[], baseVarName: (name: string) => string): string[] {
+  const used = new Set<string>();
+  return rawNames.map((raw) => {
+    const base = baseVarName(raw);
+    let candidate = base;
+    let n = 2;
+    while (used.has(candidate)) { candidate = `${base}${n++}`; }
+    used.add(candidate);
+    return candidate;
+  });
+}
+
 /** Returns the file name (without extension) for a component. e.g. `'teaser_list'` -> `'teaser-list'` */
 export function componentFileName(name: string): string {
   return toKebabCase(name);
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/schema/init/generate-code.ts
```diff
@@ -11,36 +12,70 @@ import {
...
+    const base = baseVarName(raw);
+    let candidate = base;
+    let n = 2;
+    while (used.has(candidate)) { candidate = `${base}${n++}`; }
+    used.add(candidate);
+    return candidate;
+  });
+}
+
+/**
```
dipankarmaikap, line 71: This is an issue. Unique component names such as `hero_cta` and `hero-cta` still collide after `toKebabCase()`. Since `componentFileName()` generates file names from the kebab-case value, both end up as `hero-cta.ts`.

dipankarmaikap (changes requested):

packages/cli/src/commands/schema/init/generate-code.ts
maoberlehner: Fixed!

dipankarmaikap (approved):
Nice work! LGTM.

PR #683, opened 2026-07-08
POC: OAuth Authorization Code Grant login in the CLI (DX-484)

maoberlehner:
> [!WARNING]
> **POC, do not merge.** This branch exists to answer the DX-484 questions against the real backend. The code is throwaway by design (no unit tests, `konsola` output, full-catalog scopes). Productization happens in follow-up tickets.

Proof of concept for OAuth Authorization Code Grant login in the CLI (Linear [DX-484](https://linear.app/storyblok/issue/DX-484)). Adds `storyblok oauth setup|login|call|refresh` POC commands and runs the full flow end to end against production EU with the sandbox org. Full evidence lives in [`DX-484-oauth-poc-findings.md`](https://github.com/storyblok/monoblok/blob/feat/DX-484-oauth-login-poc/DX-484-oauth-poc-findings.md); spec and plan are under `docs/superpowers/`.

### 1. Does `offline_access` return a refresh token?

**Yes.** `POST /oauth/token` returns `refresh_token` on the initial code exchange and on every refresh. Access tokens live 15 minutes (`expires_in: 900`), refresh tokens 1 month (rolling).

### 2. Does refresh work as documented (new access token, refresh token unchanged)?

**No: the refresh token ROTATES on every refresh.** Each refresh revokes the old grant and returns a new access token AND a new refresh token; reusing the pre-rotation refresh token yields `400 {"error": "invalid_grant", …}`. Consequence: clients must atomically persist the new refresh token after every refresh. A crash between refresh and persist permanently kills the session, and two processes sharing one credentials file will invalidate each other. The MAPI-client interceptor needs single-flight refresh with persist-before-use. The backend docs saying "refresh token unchanged" are wrong; the backend team should be told.

### 3. Are expired-token vs missing-scope errors distinguishable (401 vs 403)?

**Yes, cleanly, by status code.**

| Case | Status | Body |
| --- | --- | --- |
| Expired access token (16 min) | `401` | `{"error": "Unauthorized"}` |
| Missing scope (`stories:read`-only grant calling components) | `403` | `{"error": "Insufficient scope: components:read is required"}` |

Neither sends a `WWW-Authenticate` header on MAPI endpoints, so discriminate on status code only. Refresh after expiry recovers the session (verified: 401 → refresh → 200), so a refresh-on-401 interceptor is safe; never refresh on 403.

### 4. Does find-or-create behave sanely on re-runs?

**Yes.** Re-running `oauth setup` finds the existing "Storyblok CLI" app by name and reads the secret via `GET /v1/oauth_clients/:id`; no duplicate apps. Two create-time API facts the issue's sketch missed: `POST /v1/oauth_clients` **422s without a unique `slug`** (format validation without `allow_blank` plus global uniqueness), and the scope catalog returns `{resource, actions}` groups where valid scope strings are `resource:action` (32 scopes + `offline_access`).

### 5. Does `/oauth/authorize` accept `dev_oauth_redirect_uri`?

**Yes.** With `oauth_redirect_uri` on port 4900 and `dev_oauth_redirect_uri` on port 4901, a full authorize + exchange against 4901 succeeded. Both URIs are valid concurrently, so dev and prod ports can coexist on one app.

### 6. Consent UX surprises with a CLI as the app?

None blocking. App shows as "Storyblok CLI" with a placeholder icon. The consent screen includes a space picker and correctly shows only the requested scope subset when the CLI narrows the request. Entry via `GET /oauth/init` on the MAPI host works as documented.

## MAPI client integration assessment

OAuth in `@storyblok/management-api-client` would work like this: login stays outside the client (host does the consent dance and hands over tokens plus a persist callback); requests send `Authorization: Bearer <access_token>`; refresh proactively or on 401 only (never 403, which is a scope problem); refresh must be single-flight with persist-before-use because the refresh token rotates and the old one is single-use, then retry once.

**Verdict: practical for interactive, single-process use; risky beyond that.** The mechanics are fine, but strict rotation with no grace window means: two processes sharing one credentials file invalidate each other on refresh (in-process single-flight does not fix cross-process races; needs file locking or per-process grants), a crash between refresh and persist permanently kills the session, and 15-minute access tokens make refresh constant. Recommendation: ship OAuth for the CLI, and ask the backend team for a reuse grace period on rotated refresh tokens or a machine-to-machine grant before targeting long-running or multi-process consumers (MCP server, CI).

## Additional findings

- Env-var client credentials (`STORYBLOK_OAUTH_CLIENT_ID`/`SECRET`) work with zero setup, so the CI path is viable.
- `POST /oauth/token` is form-encoded (`application/x-www-form-urlencoded`), not JSON.
- PKCE S256 is mandatory for scoped grants and verified end to end.
- `/v1/oauth_clients` requires `can_manage_org?` plus the org `allow_oauth_grants` flag (distinct 403s for each).
- Auth header formats: PAT is `Authorization: <token>`, OAuth is `Authorization: Bearer <token>`; `token_type` comes back lowercase (`"bearer"`).

## Try it yourself

Prerequisites: a PAT with an org manager role on an org that has OAuth grants enabled (`allow_oauth_grants`), and a space in that org. Everything runs against production EU.

```bash
pnpm install && pnpm nx build storyblok

# Provision the "Storyblok CLI" app and store client credentials (PAT is used once, never stored)
node packages/cli/dist/index.mjs oauth setup --token <PAT>

# Full authorization code + PKCE flow: opens the browser for consent, callback on localhost:4900
node packages/cli/dist/index.mjs oauth login

# Probe MAPI with the access token (defaults to GET /v1/spaces/:id/stories)
node packages/cli/dist/index.mjs oauth call --space <SPACE_ID>

# Refresh: prints whether access and refresh tokens rotated
node packages/cli/dist/index.mjs oauth refresh
```

Tokens land in `~/.storyblok/credentials.json` under `oauth.eu`. To reproduce the 401 vs 403 findings: wait 16 minutes and run `oauth call` again for the 401, or log in with `oauth login --scope stories:read offline_access` and call `oauth call /v1/spaces/<SPACE_ID>/components --space <SPACE_ID>` for the 403.

### Manual path (without the PAT setup)

If you already have a client id and secret for an existing OAuth app (for example from the UI, or handed over by an org admin), skip the PAT path entirely. The app's redirect URI must be `http://localhost:4900/oauth/callback`. Either store the credentials:

```bash
node packages/cli/dist/index.mjs oauth setup --client-id <id> --client-secret <secret>
```

or skip storage entirely and pass them via env vars (the zero-setup CI path):

```bash
STORYBLOK_OAUTH_CLIENT_ID=<id> STORYBLOK_OAUTH_CLIENT_SECRET=<secret> \
  node packages/cli/dist/index.mjs oauth login --scope stories:read stories:write offline_access
```

Note for the manual path: no scope catalog is stored, so `oauth login` without `--scope` falls back to `stories:read stories:write offline_access`. Whatever you request must be a subset of the app's allowed scopes.

### DIFF
--- a/DX-484-oauth-poc-findings.md
+++ b/DX-484-oauth-poc-findings.md
@@ -1,56 +1,87 @@
-# DX-484 OAuth POC — Findings
+# DX-484 OAuth POC Findings
 
-**Environment:** production EU (`app.storyblok.com` / `mapi.storyblok.com`), sandbox org.
-**Date:** <!-- fill in when the runbook runs -->
-**CLI branch:** `feat/DX-484-oauth-login-poc`
+**Environment:** production EU (`app.storyblok.com` / `mapi.storyblok.com`), sandbox org, space `monoblok-cli-qa` (290020686946333).
+**Date:** 2026-07-07 (setup) / 2026-07-08 (full runbook).
+**CLI branch:** `feat/DX-484-oauth-login-poc` (`storyblok oauth setup|login|call|refresh` POC commands).
 
 ## Must-answer questions
 
 ### 1. Does `offline_access` actually return a refresh token?
-- **Answer:**
-- **Evidence:** (raw `/oauth/token` response with secrets redacted)
-- Share with the MCP team — they are blocked on this question.
+
+- **Answer: YES.** Requesting `offline_access` in the authorize scope returns a `refresh_token` from `POST /oauth/token`, both on the initial code exchange and on every refresh.
+- **Evidence:** raw token response (redacted):
+
+  ```json
+  {
+    "access_token": "sb_o…(masked)",
+    "refresh_token": "sb_o…(masked)",
+    "token_type": "bearer",
+    "expires_in": 900,
+    "scope": "asset_folders:read … workflows:write offline_access"
+  }
+  ```
+
+  Notes: `expires_in` is 900 (15 min, matches `OauthGrant::ACCESS_TOKEN_LIFETIME`). Both token kinds use an `sb_o…` prefix (`sb_oat_` access / `sb_ort_` refresh). The response contains **no `space_ids` field**, although the grant is space-scoped by the consent selection; clients cannot learn the granted space(s) from the token response. Share this section with the MCP team.
 
 ### 2. Does refresh work as documented (new access token, refresh token unchanged)?
-- **Answer:**
-- **Note:** storyrails `token_endpoint.rb` revokes the old grant and issues a NEW refresh token on every
-  refresh (rotation) — the "refresh token unchanged" expectation appears wrong at the code level.
-  Record what actually happens.
-- **Evidence:** (`oauth refresh` output)
+
+- **Answer: NO. The refresh token ROTATES on every refresh.** Each `grant_type=refresh_token` call returns a new access token AND a new refresh token; the old refresh token is revoked and single-use.
+- **Evidence:**
+  - `oauth refresh` (twice in a row): `access token changed: true`, `Refresh token rotated: true` both times; response keys `access_token, refresh_token, token_type, expires_in, scope`.
+  - Reusing the pre-rotation refresh token: `400 Bad Request` `{"error": "invalid_grant", "error_description": "The provided access grant is invalid, expired, or revoked (…)"}`.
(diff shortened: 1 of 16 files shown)


maoberlehner:
Superseded by #703 (feat(cli): add oauth login). Closing the POC.

PR #684, opened 2026-07-08
feat(richtext): type richtext field value from OpenAPI spec

dipankarmaikap:
## Summary

Types the richtext field value in OpenAPI format and wires `@storyblok/richtext` and `storyblok` (CLI) into the codegen pipeline so the node/mark types are generated from the spec rather than maintained by hand.

### `tools/openapi-codegen`
- **`richtext-field-value.yaml`** — expanded from a loose `type: object` into a fully discriminated schema with `$defs` for all 17 node types and 12 mark types, mirroring `richtext-attrs.ts`
- **`overlay.openapi.yaml`** — added `RichTextNode` and `RichTextMark` as top-level schema refs so they are emitted as named types
- **`aliases.ts`** — registered `RichtextDoc`, `RichTextNode`, `RichTextMark`
- **`index.ts`** — fixed `copyWrapperTemplates` to skip writing the empty `types/_sources.ts` when no wrapper templates are requested

### `packages/richtext`
- **`scripts/generate.ts`** — new OpenAPI codegen entry point (mirrors `live-preview`), generates `RichtextDoc`, `RichTextNode`, `RichTextMark` into `src/generated/`
- **`src/generated/overlay/types.gen.ts`** — committed generated output
- **`package.json`** — added `generate:openapi` script, `@storyblok/openapi-codegen` devDep, and nx target
- **`src/static/generate/richtext-type.ts`** — stop generating `SbRichTextNode`/`SbRichTextMark` as hand-rolled unions; delegate to OpenAPI-generated types with `_key`, `context`, and common optional fields intersected in
- **`src/static/types.ts`** — `SbRichTextDoc` is now `RichtextDoc`; `SbRichTextInput` explicitly includes `SbRichTextDoc`
- **test-utils** — updated to use `SbRichTextMark[]` directly and `NonNullable<SbRichTextInput>` for fixture types

### `packages/cli` (storyblok)
- **`scripts/generate.ts`** — new OpenAPI codegen entry point, generates `RichtextDoc`, `RichTextNode`, `RichTextMark` into `src/generated/`
- **`src/generated/overlay/types.gen.ts`** — committed generated output
- **`package.json`** — added `generate:openapi` script, `@storyblok/openapi-codegen` devDep, and nx target
- **`src/types/storyblok.ts`** — replaced the hand-written `StoryblokRichtext` interface (loose `type: string` with `any` attrs) with a backward-compatible alias to `RichtextDoc`; also exports `RichTextNode` and `RichTextMark`

## Testing

- `pnpm nx run @storyblok/richtext:test:types` ✅
- `pnpm nx run @storyblok/richtext:test` ✅ (174 tests)

Fixes DX-487


packages/richtext/scripts/generate.ts
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env -S node --experimental-strip-types --no-warnings=ExperimentalWarning
```
alexjoverm, line 1: (non-blocking): are the flags in this line needed? Shouldn't it be `#!/usr/bin/env tsx` like in the clients/generate.ts?

.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
```
alexjoverm, line 1: @dipankarmaikap I would keep this out of the scope of this PR - we have a ticket to revisit formatting using OXC

packages/cli/src/types/storyblok.ts
```diff
@@ -125,10 +125,7 @@ export interface StoryblokTable {
...
-  type: string;
-  content?: StoryblokRichtext[];
-  marks?: StoryblokRichtext[];
-  attrs?: Record<string, any>;
-  text?: string;
-}
+export type { RichtextDoc, RichTextMark, RichTextNode } from '../generated/overlay/types.gen';
+
+// Backward-compatible alias
+export type { RichtextDoc as StoryblokRichtext } from '../generated/overlay/types.gen';
```
alexjoverm, line 131: While it's exported as an alias, the type is stricter than before - it's fine if we want to publish it like this, as it's a type fix, but let's keep in mind some users might take it as a breaking change as technically it might be.

packages/richtext/src/static/types.ts
```diff
@@ -25,9 +28,9 @@ export interface RenderSpec {
 }
 
 /** Canonical type for a Storyblok RichText JSON root */
-export type SbRichTextDoc = SbRichTextNode & { type: 'doc' };
+export type SbRichTextDoc = RichtextDoc;
 export type SbRichTextTextNode = SbRichTextNode & { type: 'text' };
-export type SbRichTextInput = SbRichTextNode | SbRichTextNode[] | null | undefined;
+export type SbRichTextInput = SbRichTextDoc | SbRichTextNode | SbRichTextNode[] | null | undefined;
```
alexjoverm, line 33: The type is different than before - it's fine if we want to publish it like this, as it's a type fix, but let's keep in mind some users might take it as a breaking change as technically it might be.

package.json
```diff
@@ -30,11 +30,6 @@
   "resolutions": {
     "cypress": "^14.3.3"
   },
-  "pnpm": {
```
alexjoverm, line 33: (non-blocking): same here - it's not related to the PR. No issue keeping it as it's low risk, but better to tackle it in its own PR moving forward. Will give better visibility, for example, in cases where a regression is introduced

pnpm-workspace.yaml
```diff
@@ -6,3 +6,5 @@ packages:
 onlyBuiltDependencies:
```
alexjoverm, line 6: see https://github.com/storyblok/monoblok/pull/684/changes#r3553314515

alexjoverm (commented):
@dipankarmaikap added a few minor notes, but will leave to @maoberlehner the final review.

Important note: if we squash & merge, this PR will trigger a new version on Angular, CLI, JS, Richtext and Vue SDKs. If that's intended, fine - otherwise, we need to rebase the commits and not squash when merging

tools/openapi-codegen/specs/overlay.openapi.yaml
```diff
@@ -23,6 +23,10 @@ components:
       $ref: ./shared/stories/field-types/plugin-field-value.yaml
     RichtextFieldValue:
       $ref: ./shared/stories/field-types/richtext-field-value.yaml
+    RichTextNode:
```
maoberlehner, line 26: Naming of the Richtext / Rich Text field already is a mess, but at least we should not make it worse. The overlay currently uses `Richtext`. Let's keep it that way. And/or talk to docs and product about what the right name is (in the Storyblok UI, the field is named `Richtext`).

tools/openapi-codegen/specs/shared/stories/field-types/richtext-field-value.yaml
```diff
@@ -1,18 +1,597 @@
...
+        type: string
+        enum: [blok]
+      attrs:
+        type: object
+        required: [id, body]
+        properties:
+          id:
+            type: [string, 'null']
+            description: Blok instance ID
+          body:
```
maoberlehner, line 577: Should use:

```yaml
body:
  oneOf:
    - type: array
      items:
        $ref: ../block-content.yaml   # from field-types/ dir
    - type: 'null'
```

tools/openapi-codegen/specs/shared/stories/field-types/richtext-field-value.yaml
```diff
@@ -1,18 +1,597 @@
...
+      type:
+        type: string
+        enum: [blok]
+      attrs:
+        type: object
+        required: [id, body]
+        properties:
+          id:
+            type: [string, 'null']
+            description: Blok instance ID
```
maoberlehner, line 576: According to docs team, we should use `Block` not `Blok`.

PR #685, opened 2026-07-09
fix: change storyblokEditable to have narrow type

dipankarmaikap:
Closes #468.

Thanks to @intellix for the original contribution!

This picks up the fix from #468 and takes it a step further:
- Narrows the `storyblokEditable` parameter type to `{ _editable?: string }` — the function only needs this one property, so there's no reason to require callers to extend a broad type like `SbBlokData`
- Makes the `blok` parameter optional
- Tightens the internal guard to `!== 'string'` for correct type narrowing inside the function body

### DIFF
--- a/packages/js/src/editable.ts
+++ b/packages/js/src/editable.ts
@@ -1,6 +1,4 @@
-import type { SbBlokData } from './types';
-
-export default (blok: SbBlokData) => {
+export default (blok: { _editable?: string }) => {
   if (typeof blok !== 'object' || typeof blok._editable === 'undefined') {
     return {};
   }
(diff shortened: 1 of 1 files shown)


alexjoverm (changes requested):
@dipankarmaikap looks good, but the CI is failing... Maybe the type needs to be translated to the rest of SDKs? 

Also note - as type is changed and strengthen, in a way it's a breaking change. It's just a type, and we better consider it a type fix rather than a breaking change, but it's a thing to be aware in case anyone raises it.

dipankarmaikap:
CI is passing now — yesterday there were widespread GitHub Actions failures unrelated to this PR.

On the breaking change: agreed, it's a type fix. Narrowing to `{ _editable?: string }` is actually more permissive for callers than `SbBlokData` — any valid call before still works.

alexjoverm (approved):

PR #686, opened 2026-07-09
fix(cli): match uppercase X in asset dimension segment on push (MNT-1380)

LucasKuhn:
## Why

Porsche's migration (MNT-1370) hit inconsistent dimension-preservation after upgrading to
#672: some pushed assets kept their `<width>x<height>` URL segment, others silently
dropped it — even though the source assets looked identical.

## Root cause

Casing. `extractAssetSizeFromFilename`'s regex only matched lowercase `x`:

| Source segment | Result before fix |
|---|---|
| `2042x1331` (lowercase) | kept |
| `860X645` (uppercase) | **dropped** |

storyrails' own `S3_FILENAME_FORMAT` is case-insensitive and legitimately produces
uppercase-`X` segments, so the CLI's stricter regex was silently losing dimensions the
backend considers valid. Fix: add the `i` flag.

## Reproduce

1. Push an asset whose *source* CDN URL has an uppercase-X segment, e.g.
   `.../f/SPACE/860X645/hash/image.jpg`, with `storyblok assets push`.
2. Before this fix: target asset's URL drops the dimensions folder entirely.
3. A lowercase-x source keeps it either way — which is what made this easy to miss.

Fixes MNT-1380

### DIFF
--- a/packages/cli/src/commands/assets/utils.test.ts
+++ b/packages/cli/src/commands/assets/utils.test.ts
@@ -11,6 +11,10 @@ describe('extractAssetSizeFromFilename', () => {
     expect(extractAssetSizeFromFilename('https://a.storyblok.com/f/293255674717942/8ae82d3a12/image.jpg')).toBeUndefined();
   });
 
+  it('extracts the dimensions segment when it uses an uppercase X', () => {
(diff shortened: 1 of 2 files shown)


maoberlehner (approved):
LGTM, thanks for fixing this directly.

PR #688, opened 2026-07-10
feat(cli): add --all flag for assets transfer command

maoberlehner:
Adds an `--all` flag to `storyblok assets transfer` that transfers every asset in a space into the organization's shared asset library, so enterprise customers no longer need to pass individual asset IDs.

```bash
storyblok assets transfer --all --folder-id 789 --space YOUR_SPACE_ID
storyblok assets transfer --all --folder-id 789 --space YOUR_SPACE_ID --dry-run
```

## How

Enumerate-then-transfer. A new `fetchAllSpaceAssetIds` action paginates the space's asset list (via the existing `fetchAllPages` helper) to collect all asset IDs, then hands that array to the existing, unchanged `transferAssets`. The current per-ID reporting, backpressure lock, summary, and exit-code logic are reused as-is.

Behavior:

- `transfer --all` transfers every asset in the space.
- The positional `[asset-id...]` argument is now optional; it is required only when `--all` is not used.
- `--all` combined with explicit asset IDs is rejected (exit 2).
- Neither `--all` nor asset IDs provided is rejected (exit 2).
- `--all` on an empty space exits 0 with a "No assets found" message and transfers nothing.
- `--all --dry-run` enumerates the assets (read-only list calls, zero `convert` calls) and prints an accurate plan, then exits 0.
- An API failure while enumerating assets for `--all` exits cleanly through `handleError` (exit 2), matching the sibling `pull`/`push` commands, rather than surfacing as an unhandled rejection.

Safety for this bulk, one-way operation relies on `--dry-run` (no confirmation prompt), consistent with the existing command.

Fixes DX-457

### DIFF
--- a/packages/cli/src/commands/assets/actions.ts
+++ b/packages/cli/src/commands/assets/actions.ts
@@ -42,6 +42,34 @@ export const fetchAssets = async ({ spaceId, params }: {
 };
 
 /**
+ * Fetches the IDs of every asset in a space by paginating the Management API
+ * asset list. Used by `assets transfer --all` to resolve the full working set
+ * before transferring. Mirrors the pagination in `fetchAssetInternalTagsByName`.
+ */
+export const fetchAllSpaceAssetIds = async (spaceId: string): Promise<number[]> => {
+  try {
+    const client = getMapiClient();
+    const assets = await fetchAllPages(
+      (page: number) => client.assets.list({
+        path: { space_id: Number(spaceId) },
+        query: { page, per_page: 100 },
+        throwOnError: true,
+      }),
+      data => data?.assets ?? [],
+    );
+    return assets.reduce<number[]>((ids, asset) => {
+      if (typeof asset?.id === 'number') {
+        ids.push(asset.id);
+      }
+      return ids;
+    }, []);
+  }
+  catch (maybeError) {
+    handleAPIError('pull_assets', toError(maybeError));
+  }
+};
+
+/**
  * Fetches the space's internal tags of type `asset` keyed by name.
  *
  * Used by `assets push` to translate source-space tag names carried in pulled
--- a/packages/cli/src/commands/assets/actions.ts
+++ b/packages/cli/src/commands/assets/actions.ts
@@ -66,7 +66,7 @@ export const fetchAllSpaceAssetIds = async (spaceId: string, params?: AssetListQ
     }, []);
   }
   catch (maybeError) {
-    handleAPIError('pull_assets', toError(maybeError));
+    handleAPIError('transfer_enumerate_assets', toError(maybeError));
   }
 };
 
--- a/packages/cli/src/commands/assets/actions.ts
+++ b/packages/cli/src/commands/assets/actions.ts
@@ -359,11 +359,20 @@ export interface TransferResult {
 }
 
 /**
+ * Upper bound on promises allocated at once by `transferAssets`, so very large
+ * `--all` sets don't build one suspended promise per asset up front. Concurrency
+ * is still governed by the backpressure lock; this only caps promise allocation.
+ */
+const TRANSFER_CHUNK_SIZE = 500;
+
+/**
  * Transfers multiple assets into the shared asset library, bounding in-flight
  * requests with the shared pipeline backpressure lock (2× the configured rate
  * limit), matching the throttle headroom used by the asset and story streams.
  * Per-asset errors are captured as failed results rather than aborting the
- * whole batch.
+ * whole batch. Fan-out is chunked (`TRANSFER_CHUNK_SIZE`) so very large asset
+ * sets don't allocate one promise per asset up front, and `onProgress` reports
+ * completion count as each asset finishes.
  */
 export const transferAssets = async (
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/assets/transfer/index.ts
```diff
@@ -50,13 +52,46 @@ transferCmd
...
+      catch (maybeError) {
+        handleError(toError(maybeError), verbose);
+        process.exitCode = 2;
+        return;
+      }
+      if (ids.length === 0) {
+        ui.info(options.query
+          ? `No assets in space ${space} match the query. Nothing to transfer.`
+          : `No assets found in space ${space}. Nothing to transfer.`);
+        logger.info('Transferring assets finished (no assets found)');
```
alexjoverm, line 89: @maoberlehner I think we're missing a few logger entries on critical points along the process, including the assets enumeration (similar to assets pull)

packages/cli/src/commands/assets/actions.ts
```diff
@@ -41,6 +41,35 @@ export const fetchAssets = async ({ spaceId, params }: {
...
 
+/**
+ * Fetches the IDs of every asset in a space (optionally filtered) by
+ * paginating the Management API asset list. Used by `assets transfer --all`
+ * to resolve the full working set before transferring. Mirrors the
+ * pagination in `fetchAssetInternalTagsByName`.
+ */
+export const fetchAllSpaceAssetIds = async (spaceId: string, params?: AssetListQuery): Promise<number[]> => {
+  try {
+    const client = getMapiClient();
```
alexjoverm, line 52: (minor):  `fetchAllSpaceAssetIds` calls `handleAPIError` in its catch block, which **always throws** (the return type is `never`). The command's catch block in `transfer/index.ts:74-78` then catches that rethrown `APIError` and passes it to `handleError`. This works, but `handleAPIError` already enriches the error with action context (`pull_assets` label), and then `handleError` logs it again. This is fine for CLI output (the user sees a clear message), but the action name `'pull_assets'` is misleading — this is a transfer enumeration, not a pull.

**Suggestion:** Either:
- (a) Use a distinct action key (add `'enumerate_assets'` to `API_ACTIONS`), or  
- (b) Accept the `pull_assets` label since it is the same underlying endpoint.

packages/cli/src/commands/assets/transfer/index.ts
```diff
@@ -50,13 +52,46 @@ transferCmd
       return;
     }
 
-    const ids = assetIds.map(id => Number(id)).filter(id => !Number.isNaN(id));
-    if (ids.length === 0) {
-      handleError(new CommandError(`Please provide at least one valid asset ID.`), verbose);
+    // `--query` selects a filtered subset of the space, so it implies the bulk
+    // path just like `--all`; either flag means "transfer what's in the space",
+    // which is mutually exclusive with naming explicit asset IDs.
+    const bulk = Boolean(options.all || options.query);
```
alexjoverm, line 64: question(non-blocking): I'm a bit unsure about the naming and semantics here, as I think it can be confusing for the user.

The comment it says both options are mutually exclusive, but seems they're not? (which is perfect, maybe the comment is misleading) so user can run `--all --query "search=foo"`, which semantically can be read as "Transfer all assets, but filter them by X". But --all, behaviourally, refers to "where to target it" (subject) rather than "which assets" (object).

Have you considered to name the flag along the lines of `--to all|space|shared`?

alexjoverm (changes requested):

packages/cli/src/commands/assets/transfer/index.ts
maoberlehner: > The comment it says both options are mutually exclusive, but seems they're not?

I agree the comment is misleading. In case of `--all --query "search=foo"` `--all` would be a no-op as `--query` implies `--all`. Although I find `--all --query "search=foo"` confusing too because spelled out it says `Transfer all assets but not really all but filtered by query.`. While `--query "search=foo"` implies `Transfer all that match query.`.

> Have you considered to name the flag along the lines of `--to all|space|shared`?

Hmm, not really, as `transfer` is our name for `transfer to shared`, so it implicitly contains `shared` target.
Maybe `transfer` was the wrong direction and we should've gone the `move` route. Then `--to` would make sense.

alexjoverm: Agree, I think `--query` implicitly implies will be a filtered bulk operation :) I would suggest we make `--all` and `--query` fully mutually exclusive

(and btw, disregard the comment about `--to` as I just realized I misunderstood it 😅 )

packages/cli/src/commands/assets/transfer/index.ts
maoberlehner: Added!

packages/cli/src/commands/assets/actions.ts
maoberlehner: I did go with `a`.

alexjoverm (approved):
Neat! Thanks @maoberlehner

PR #689, opened 2026-07-10
fix(live-preview): narrow storyblokEditable parameter type and tighten guard

dipankarmaikap:
Mirrors the type fix from #685 for the `@storyblok/live-preview` package.

- Narrows the `storyblokEditable` parameter type to `{ _editable?: string }` — removes the dependency on the generated `BlockContent` type for this function
- Tightens the internal guard to `typeof editable !== 'string' || !editable` to catch non-string values at runtime

Fixes DX-489

### DIFF
--- a/packages/live-preview/src/editable.ts
+++ b/packages/live-preview/src/editable.ts
@@ -1,13 +1,11 @@
-import type { BlockContent } from './generated/types/field';
-
 interface EditableOptions {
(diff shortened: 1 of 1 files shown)


maoberlehner (approved):
LGTM

alexjoverm (approved):

PR #695, opened 2026-07-14
chore(ci): build, lint and test all packages

maoberlehner:
Run CI (`build,lint,test,test:types`) across **all** projects instead of only published packages — so tooling, `@storyblok/openapi`, `monoblok-cli`, and the example playgrounds are all covered.

## Changes

**CI (`ci.yaml`)**
- Matrix selector → `-p="*"` (all projects).
- **Static job name** `build (${{ matrix.name }})` → checks are `build (linux)` / `build (windows)`, decoupled from the selector. GitHub derives a matrix job's name from every matrix value, so a changing `projects` list used to rename the required check and leave it *"Waiting for status"* forever (the #694 problem).

**Playground repairs** (surfaced by building everything)
- **astro-ssr**: added `packages/astro/playground/shared/package.json` so `vue`/`react`/`@storyblok/astro` resolve from the shared components dir under pnpm's strict layout.
- **client-nextjs**: `storyblok-js-client` was `file:..` (wrong path) → `workspace:*`; removed an illegal extra export from `app/page.tsx`.
- **client-vanilla**: cast the MAPI component body the client `post` types don't model.
- **richtext-vanilla**: declared the `?url&raw` combined-query module.
- **next-latest** / **next-latest-static**: typed `blok` props; dropped the `next lint` script (removed in Next 16).

**Structural opt-outs** (can't build reliably in CI; expose no `build` target via `build:demo`)
- **astro-ssg** / **astro-test**: pre-render by fetching live Storyblok content at build time.
- **next-13-pages-router**: `@types/react` version skew vs. the newer `@storyblok/react` (intentionally-legacy demo).

## ⚠️ Coordinated rollout (required)

Renames the required check `build (ubuntu-latest, tag:npm:public)` → `build (linux)`. The shared `protect-release-branches` ruleset must flip in lockstep:

1. Merge to `main` via **admin bypass** (own checks pass; the old required context won't report).
2. **Immediately** update the ruleset required context → `build (linux)`.
3. Propagate to `alpha` (**merge** `main` → `alpha`, don't rebase the published branch).

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -3,6 +3,13 @@ packages:
   - packages/*/playground
   - packages/*/playground/*
   - tools/*
+catalog:
+  # Pin a single React 19 types version so the React playgrounds and the
+  # workspace-linked @storyblok/react resolve one @types/react copy. Must track
+  # what Next.js expects (its generated route types reference the global React
+  # namespace), otherwise a duplicate ReactNode causes type errors.
+  "@types/react": 19.2.3
+  "@types/react-dom": 19.2.3
 onlyBuiltDependencies:
   - cypress
   - esbuild
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,14 +33,17 @@ permissions:
 
 jobs:
   build:
+    name: build (${{ matrix.name }})
     environment: test
     strategy:
       fail-fast: false
       matrix:
         include:
           - os: ubuntu-latest
-            projects: tag:npm:public
+            name: linux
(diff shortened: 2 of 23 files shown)


pnpm-workspace.yaml
```diff
@@ -3,6 +3,13 @@ packages:
   - packages/*/playground
   - packages/*/playground/*
   - tools/*
+catalog:
```
alexjoverm, line 6: Nice improvement!

alexjoverm (approved):
Great one, thanks Markus!

PR #703, opened 2026-07-20
feat(cli): add oauth login

maoberlehner:
Adds OAuth login (Authorization Code grant with PKCE) to the Storyblok CLI, alongside the existing PAT and email/password flows.

- `storyblok login --oauth` (also selectable in interactive `storyblok login`) opens the browser for consent, receives the callback on `http://localhost:4900/oauth/callback`, exchanges the code for tokens, and introspects the grant for scopes and authorized spaces. Email stays the default strategy in the interactive prompt.
- Nothing to configure: the CLI ships a baked-in first-party OAuth client, the model `gh` and `gcloud` use. It is a public client, so the secret is not a security boundary; PKCE protects the code exchange. One integration app covers every region. `STORYBLOK_OAUTH_CLIENT_ID`/`STORYBLOK_OAUTH_CLIENT_SECRET` still override it for development or self-hosted instances.
- Login requests the full scope catalog plus `offline_access` in one consent, so every command is covered. This must stay a subset of the registered app's allowed scopes.
- OAuth sessions are stored per region under an `oauth` key in `~/.storyblok/credentials.json` (mode `0600`), kept separate from PAT entries. An `activeRegion` pointer records the region most recently logged into, so the right session loads when several regions are authenticated. A PAT still wins over an OAuth session, and now says so instead of silently shadowing a fresh `login --oauth`.
- Grants can be space-scoped: commands targeting a space outside the grant fail early (exit code 2) with clear guidance to re-run login and select that space.
- Port 4900 must be free while authorizing (it is the app's only registered redirect URI). If it is taken, the error names the process and PID holding it.
- `storyblok logout` revokes the grant server-side, then clears the local OAuth session for the region, including the `activeRegion` pointer.
- The OAuth code lives in `src/lib/oauth/`, not under `src/commands/`: it registers no command, it is auth support the login, logout, session and program modules depend on.

## Keeping a long command authenticated

Access tokens live 15 minutes, which is shorter than a real `stories push` or `assets pull`. Rather than resolving a token once per command, the CLI hands the management API client a credential provider that is resolved before every request:

- Under a minute of lifetime left, the request waits for a refresh.
- Between one and three minutes, the cached token is still served while a refresh runs alongside, so a long-running command never stalls on it.
- Above that, the cached token is used as is.

The refresh token rotates and is single-use, so the whole read-exchange-persist cycle runs under a lock on the credentials file: two processes refreshing at once would otherwise each present the same token, and the loser's grant is already revoked server-side, ending the session for good. After acquiring the lock a process re-reads, so a token another process just rotated is reused instead of exchanged again.

There is deliberately no 401-triggered retry. With rotating refresh tokens, two racing recoveries replay the same token, which an authorization server is entitled to read as a breach and answer by revoking the whole grant — worse than the 401 being recovered from.

Commands that need no authentication never resolve the credential, so they never trigger a refresh. A failed refresh is reported once by the provider itself, because commands differ in how much of a failed request they surface.

This needs one addition to `@storyblok/management-api-client`: `personalAccessToken` and `oauthToken` now also accept a function, resolved per request via a request interceptor. Passing a string behaves exactly as before — the header is resolved once at construction and no interceptor is registered.

## Credentials file durability

`credentials.json` holds the PAT entries and every region's OAuth session, and each mutation rewrites it whole. Writes now go through an atomic replace: the temp file is created with mode `0600` from the start, so a secret never exists on disk with default permissions, and it is flushed before the rename. A file that was already too permissive is tightened on the next write, which the previous write could not do. Windows gets a short retry on the rename, where scanners and indexers hold the target briefly.

One module owns the file for both the PAT and OAuth halves, since they have to share the lock to serialize at all. The lock is a directory whose mtime is refreshed while held: collectable after a crash, never mistaken for stale during a slow refresh. It is hand-rolled rather than `proper-lockfile` because that pulls in `graceful-fs`, which requires the real `fs` module — the test suite's in-memory filesystem would not cover it, and the tests would write lock directories into the real home directory.

## Verification

The "Storyblok CLI" integration app is registered and its credentials are baked in, so the flow has been exercised end to end against production, including a real space-scoped consent:

- Authorize, introspect, and run commands against the granted space; a space outside the grant is refused before any request.
- A real refresh inside the blocking window rotates both tokens and persists them, and the command succeeds.
- The background window returns the cached token without blocking, and the rotation still lands before the process exits.
- Three concurrent processes racing one single-use refresh token all succeed, the token pair rotates exactly once, and the session stays usable. Without the lock, the same shape loses 7 of 8 concurrent writes.
- A reader polling through 40 concurrent writes never observes a truncated or empty file (113k reads).
- A stale lock left by a crashed process is reclaimed; a lock held by a live process makes the next writer wait rather than fail.
- `logout` revocation is real: a restored pre-logout session is dead server-side and surfaces re-login guidance.
- Commands that build their own client (`user`, `create`) refresh identically to those using the shared client.

Further details in the QA notes on #751, which stacks on this branch.

Fixes DX-490


maoberlehner:
Wait for the product feature to go live before merging!

packages/cli/src/commands/oauth/actions.ts
```diff
@@ -0,0 +1,103 @@
...
+interface OauthClientDetail extends OauthClientSummary {
+  oauth_identifier: string;
+  oauth_secret?: string;
+}
+
+interface OauthClientsListResponse {
+  oauth_clients: OauthClientSummary[];
+}
+
+interface OauthClientResponse {
```
alexjoverm: minor (non-blocking): I'd suggest to name interaces as `OAuth` (first 2 letter capitals) as seems to be the way named in the spec https://oauth.net/2/

packages/cli/src/commands/user/actions.ts
alexjoverm: The 401 error message still says "The token provided ... is invalid" even for OAuth access tokens, which is misleading. The error path should differentiate between PAT and OAuth credentials.

packages/cli/src/session.ts
```diff
@@ -26,29 +30,65 @@ function createSession() {
...
+        state.login = undefined;
+        state.password = undefined;
+        state.region = undefined;
+        state.authType = undefined;
+      }
     }
     state.envLogin = false;
   }
 
+  async function loadOauthSession(): Promise<boolean> {
```
alexjoverm: Seems to silently picks the first region with stored tokens (eu wins over us if both have tokens). No last-login timestamp exists to disambiguate, which could confuse users who log into multiple regions.

alexjoverm (commented):
@maoberlehner added some comments. Adding also here some minor Claude findings:

    Minor (4):
     - test/setup.ts session mock does not include clearOauthSession or authType: 'pat', which is fragile for future OAuth test paths.
     - performInteractiveLogin returns { token: '', region } for OAuth -- the empty string token is a latent footgun.
     - grant.ts uses raw fetch while actions.ts uses customFetch -- inconsistent but defensible; needs a clarifying comment.
     - DEFAULT_LOGIN_SCOPES hard-codes a restrictive fallback for BYO-client path without warning the user.

     Nit (1): colorPalette.OAUTH ?? colorPalette.PRIMARY fallback is dead code.

packages/cli/src/commands/oauth/server.ts
```diff
@@ -0,0 +1,52 @@
...
+        return;
+      }
+
+      res.writeHead(200, { 'Content-Type': 'text/html' });
+      res.end(SUCCESS_PAGE);
+      server.close();
+      clearTimeout(timer);
+
+      const error = url.searchParams.get('error');
+      if (error) {
```
dipankarmaikap, line 45: **Question:** Shouldn't we perform this check first? If there's an error, I don't think we should be returning a `200` response.

packages/cli/src/commands/oauth/token-endpoint.ts
```diff
@@ -0,0 +1,45 @@
...
+export interface TokenResponse {
+  access_token: string;
+  refresh_token?: string;
+  expires_in: number;
+  scope?: string;
+  raw: Record<string, unknown>;
+}
+
+export const exchangeToken = async (region: RegionCode, params: Record<string, string>): Promise<TokenResponse> => {
+  const response = await fetch(`https://${managementApiRegions[region]}/oauth/token`, {
```
dipankarmaikap: Should be using the `customFetch` like all the other places.

packages/cli/src/commands/oauth/setup/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (!isRegion(region)) {
+      handleError(new CommandError(`The provided region "${region}" is not valid.`), verbose);
+      return;
+    }
+
+    try {
+      // Manual path: explicit client id/secret (or interactive when only one is supplied).
+      if (options.clientId || options.clientSecret) {
+        const clientId = options.clientId ?? await input({ message: 'OAuth client id:', required: true });
+        const clientSecret = options.clientSecret ?? await password({ message: 'OAuth client secret:' });
```
dipankarmaikap: **Question:** What happens if the user passes an empty string? Currently, there's no validation for that case. Maybe we could add a simple check to ensure the input isn't empty, display an appropriate message, and prompt the user to try again.

packages/cli/src/commands/oauth/actions.ts
maoberlehner: Done

packages/cli/src/commands/oauth/grant.ts
```diff
@@ -0,0 +1,39 @@
+import type { RegionCode } from '../../constants';
+import { CommandError } from '../../utils';
+import { customFetch, FetchError } from '../../utils/fetch';
+import { getStoryblokUrl } from '../../utils/api-routes';
+import type { OAuthGrantSpace } from './store';
+
+export interface GrantIntrospection {
+  scopes: string[];
+  expires_at: string;
```
dipankarmaikap: Should this be `expires_at?: string` — we do not have to add a fallback on line 35. code already handles undefined.

packages/cli/src/commands/logout/index.ts
dipankarmaikap: We can remove one `konsola.br`.

packages/cli/src/commands/oauth/login-flow.ts
```diff
@@ -0,0 +1,98 @@
...
+export const performOAuthLogin = async (options: {
+  region: RegionCode;
+  openBrowser?: (url: string) => Promise<unknown>;
+}): Promise<OAuthLoginResult> => {
+  const { region } = options;
+  const openBrowser = options.openBrowser ?? (url => open(url));
+  const ui = getUI();
+
+  const client = await resolveOAuthClient(region);
+  const resolvedScopes = (await getOAuthEntry(region)).client?.scopes ?? client.scopes;
```
dipankarmaikap: Might be worth a double-check here. If `STORYBLOK_OAUTH_CLIENT_ID` / `STORYBLOK_OAUTH_CLIENT_SECRET` are set, but a stored client also exists for the same region, the `client` on line `49` comes from the environment variables, while `resolvedScopes` reads the scopes from the stored client.

This could potentially lead to a mismatch between the client credentials and the scopes being used.

*P.S. AI review flagged this as a potential issue.*

dipankarmaikap (changes requested):

packages/cli/src/session.ts
maoberlehner: Improved

packages/cli/src/commands/user/actions.ts
maoberlehner: 401 now points OAuth sessions to storyblok login --oauth.

packages/cli/src/commands/oauth/server.ts
maoberlehner: Yes, good catch.

packages/cli/src/commands/oauth/setup/index.ts
maoberlehner: Validation added.

packages/cli/src/commands/oauth/login-flow.ts
maoberlehner: Confirmed. env-var clients carry no scopes, so line 50 requested the stored client's scopes with the env client's id. Fixed.

alexjoverm (changes requested):
BUG 1 — oauth setup --token hangs after API errors
  The spinner at setup/index.ts:64 is never stopped when findOrCreateCliClient() throws. the process hangs indefinitely instead of exiting. Fix: add spinner.failed() in the catch block.

  BUG 2 — PAT logout destroys OAuth client credentials
  logout/index.ts:30 calls removeAllCredentials() which writes {} to the entire credentials file. This wipes out OAuth client credentials stored via oauth setup. Fix: only remove the PAT-specific entry (api.storyblok.com), not the whole file.

dipankarmaikap (approved):
Looks good to me! We just need to get the CI passing so everything is green, but otherwise this looks good. Nice job! 🎉

alexjoverm (approved):

packages/cli/src/program.ts
```diff
@@ -70,16 +73,45 @@ export function getProgram(): Command {
       applyConfigToCommander(ancestry, resolvedConfig);
       setActiveConfig(resolvedConfig);
 
-      // Initialize mapiClient
+      // Initialize mapiClient with the active credential (PAT or OAuth access token).
       const { state, initializeSession } = session();
       await initializeSession();
-      if (state.password) {
+      if (state.authType === "oauth" && state.region) {
```
alexjoverm, line 78: (blocker): No mid-run token refresh

Access tokens live 15 minutes. refreshOAuthTokens has exactly one caller — the preAction hook — and isExpiringSoon uses a 120-second skew, so a run starts with 13–15 minutes of token life and nothing renews it. stories push and assets pull on real spaces routinely exceed that; every in-flight request starts 401ing mid-run. PAT users never hit this, so OAuth is strictly worse for the CLI's core workloads.

packages/cli/src/commands/oauth/store.ts
```diff
@@ -0,0 +1,96 @@
...
+
+// The `oauth` credentials section holds one entry per region, plus an
+// `activeRegion` pointer marking the region the user most recently logged into.
+// The pointer disambiguates which session to load when several regions are
+// authenticated at once. Its key never collides with a region code.
+type OAuthStore = Partial<Record<RegionCode, OAuthRegionEntry>> & { activeRegion?: RegionCode };
+
+const credentialsPath = (): string => join(getStoryblokGlobalPath(), "credentials.json");
+
+const readAll = async (): Promise<Record<string, unknown>> => {
```
alexjoverm: minor (non-blocking): Credentials are rewritten whole - no lock, no atomic replace

Every mutation is read-whole → mutate → write-whole with no compare-and-swap, so concurrent processes lose updates last-writer-wins. Losing a rotated refresh token is unrecoverable — the previous one is already revoked server-side. writeFile also truncates, so a crash mid-write leaves a truncated file and takes the PAT entry down with it. performOAuthLogin does two back-to-back read-modify-write cycles for no benefit.

alexjoverm (changes requested):

packages/cli/src/lib/oauth/client.test.ts
```diff
@@ -0,0 +1,31 @@
+import { afterEach, describe, expect, it } from "vitest";
```
alexjoverm, line 1: suggestion (blocking): @maoberlehner actually `oauth` is not a command per se, while now is structured in a command folder. Shall we rather place it within `lib/` in its own module?

alexjoverm:
@maoberlehner a few extra findings:

**1. If you close the opened tab, the CLI hangs**

Run storyblok login, select Oauth, the region, and when the webpage opens, close it and return the CLI. It hangs, expected behaviour would be that it handles the signal correctly and gracefully shuts down the command with an error

**2. It allows me to select spaces from other regions**

Not sure if it's ours or backend, but worth posting - After selecting `oauth`, and region `eu`, in the consent screen it allow me to select a `ap` (Australia), and signaled a sucessful login. 

image


Of course, when later trying to operate with the space, it errs:

```
▲ error Failed to pull stories
▲ error The requested resource was not found
```

**3. [UX incovenience] No search bar to select spaces**

Again, might be for backend - but I find quite inconvenient not having the posibility to search as you need to scroll through all spaces to find the ones you need (also, a select all would be convenient)

image

alexjoverm:
@maoberlehner a _potential_ finding. Seems after successful login, then the CLI neither allows me to logout nor to login again.

**Note:** this happened when I logged in via OAuth, and then on a different version of the CLI (the one we have in `main`) I tried to logout/login

image

maoberlehner:
> @maoberlehner a _potential_ finding. Seems after successful login, then the CLI neither allows me to logout nor to login again.
> 
> **Note:** this happened when I logged in via OAuth, and then on a different version of the CLI (the one we have in `main`) I tried to logout/login

Changed it so we save the oauth state to a separate file. Although it's somewhat of an edge case, it definitely happen for people having multiple projects with different versions or a local install here and using npx there.

PR #706, opened 2026-07-23
fix: workers-safe per-second request throttle across the API clients

### LINKED ISSUE #319: 🐛 [Bug]: When using shopfiy hydrogen the request count is limited to the rateLimit
### Package

storyblok-js-client (Client SDK)

### Bug Description

When using the [shopify hydrogen](https://shopify.dev/docs/storefronts/headless/hydrogen/getting-started) framework with the storyblok js api like mentioned here: https://www.storyblok.com/tp/headless-commerce-with-shopify-hydrogen-and-storyblok

The hydrogen server will fail after 5 successfull (of even failed) storyblok api requests.  

It looks like an isse with the throttled Promise handling / rateLimiting. 

### Steps to Reproduce

I used the guide here for basic setup https://shopify.dev/docs/storefronts/headless/hydrogen/getting-started

I follwed the guide here to implement storyblok api into shopify hydrogen: https://www.storyblok.com/tp/headless-commerce-with-shopify-hydrogen-and-storyblok 

The default rateLimit is 5 and the timeout is 1000ms. So you can reproduce it with the mentioned documents, start the server and refreshing the site 5 times in one second or even in 1 Minute, after that you will get an error and the hydrogen server will fail and will never get healthy again, you must restart the server

### Expected Behavior

- It is fine to get an error messeage when rate limit is reaced - but i didnt get one. 
- the api allows new requests after some time.
- the api allows more than 5 requests in the server lifetime

### Actual Behavior

Not more the 5 api requests are possible (even when there was a silent period)
you can increase the count when increasing the rateLimit -> 100 -> you have 100 requests and the server will fail.

### Code Sample

```javascript

```

### Environment

```shell
System:
    OS: Linux 6.8 Linux Mint 22.2 (Zara)
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700H
    Memory: 3.08 GB / 15.33 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 20.19.3 - ~/.nvm/versions/node/v20.19.3/bin/node
    npm: 11.5.1 - ~/.nvm/versions/node/v20.19.3/bin/npm
    pnpm: 10.6.4 - ~/.local/share/pnpm/pnpm
    Watchman: 4.9.0 - /usr/bin/watchman
  Browsers:
    Chrome: 140.0.7339.185
  npmPackages:
    @storyblok/react: ^5.4.12 => 5.4.12 
    react: ^18.2.0 => 18.3.1
```

### Error Logs

```shell
Error: The Workers runtime canceled this request because it detected that your Worker's code had hung and would never generate a response. Refer to: https://developers.cloudflare.com/workers/observability/errors/
    at async Object.fetch (file:///.../project/node_modules/miniflare/dist/src/workers/core/entry.worker.js:1032:22)
```

### Additional Context

I know this is a special case with shopify and hydrogen, but the promise issue in the error message **indicates some implementation issues**. 

I had a deeper look and i think the issue can (maybe) found here:

https://github.com/storyblok/monoblok/blob/56b5861920583c24131951215de0affd68e0f6aa/packages/js-client/src/throttlePromise.ts#L66-L76 

**So the when `activeCount >= limit` the queue is filled but there is no reject, and no next()**

I also miss the possibility to handle such things on my own like rateLimit = 0. 

At the end the queue does not handle promises the right way and that leads to trouble in server env for shopify hydrogen.

### LINKED ISSUE #353: 🐛 [Bug]: Storyblok @storyblok/react package causes Worker hangs in Hydrogen/Oxygen
### Package

### Bug Description

Shopify Hydrogen is based on React Router, and I would expect the react integration to work reasonably well out-of-the-box, but it does not

Following the guide here: https://www.storyblok.com/mp/announcing-react-sdk-v4-with-full-support-for-react-server-components and when using `storyblokInit`, the following error happens almost immediately:

```
A hanging Promise was canceled. This happens when the worker runtime is waiting for a Promise from JavaScript to resolve, but has detected that the Promise cannot possibly ever resolve because all code and events related to the Promise's I/O context have already finished.
[Error: The script will never generate a response.]
```

The issue does not occur if you manually use the api with `new Storyblok(...)`

### Steps to Reproduce

1. Initialize a new project following these instructions: https://shopify.dev/docs/storefronts/headless/hydrogen/getting-started
2. Follow this guide here: https://www.storyblok.com/mp/announcing-react-sdk-v4-with-full-support-for-react-server-components 
3. Attempt to a render a story more than once

### Expected Behavior

At least some more helpful error

### Actual Behavior

The following error is generated:

```
A hanging Promise was canceled. This happens when the worker runtime is waiting for a Promise from JavaScript to resolve, but has detected that the Promise cannot possibly ever resolve because all code and events related to the Promise's I/O context have already finished.
[Error: The script will never generate a response.]
```

### Code Sample

```javascript

```

### Environment

```shell
System:
    OS: Linux 6.14 Ubuntu 25.04 25.04 (Plucky Puffin)
    CPU: (20) x64 13th Gen Intel(R) Core(TM) i9-13900H
    Memory: 17.44 GB / 62.44 GB
    Container: Yes
    Shell: 5.2.37 - /bin/bash
  Binaries:
    Node: 22.18.0 - /home/john/.volta/tools/image/node/22.18.0/bin/node
    npm: 11.5.2 - /home/john/.volta/tools/image/npm/11.5.2/bin/npm
  Browsers:
    Firefox: 143.0.4
    Firefox Developer Edition: 143.0.4
  npmPackages:
    @storyblok/react: ^5.4.14 => 5.4.14 
    react: 18.3.1 => 18.3.1
```

### Error Logs

```shell

```

### Additional Context

The reason I'm reporting it with storyblok and not hydrogen/shopify is because I think that in this particular context, some resource is  getting reused. I think there's a retry or throttle mechanism which is global and leaks.

### LINKED ISSUE #533: 🐛 [Bug]: useStoryblokApi() in Astro Cloudflare Workers SSR environment causes request hangs/crashes
### Package

@storyblok/astro (Astro Integration)

### Bug Description

When `useStoryblokApi()` is used in Astro SSR request paths running on Cloudflare Workers, navigation can eventually cause requests to hang and the Worker to fail. In testing, replacing those server-side `useStoryblokApi()` calls with plain `fetch()` calls to the Storyblok CDN endpoints stopped the crashes. The issue reproduces locally with `astro dev` + `@astrojs/cloudflare` and on a deployed Cloudflare Worker. It has been confirmed and replicated in both Chrome and Safari.

### Steps to Reproduce

1. Clone the repro project: https://github.com/nonsponsored/bug-astro-ssr-cloudflare-workers-storyblok
2. Add a `.env` file with:
- `STORYBLOK_API_KEY=<preview token>`
- `STORYBLOK_API_VERSION=draft`
- `STORYBLOK_REGION=<us or eu>`
3. Install dependencies and run the project:
`npm install`
`npm run dev`
4. Open the repo in a browser.
5. Repeat clicking back and forth between:
- `/sdk-space`
- `/sdk-links`
6. Continue navigating until the Worker fails.
7. Optionally repeat the same test on a deployed Cloudflare Worker build.

### Expected Behavior

SSR navigation using useStoryblokApi() should continue to work without hanging, suspending the Worker, or throwing a runtime exception.

### Actual Behavior

After repeated navigation, the Worker eventually fails. In the full application this could happen after roughly 5 navigations on content-heavy pages; in the minimal repro it can take around 20–30 clicks.
Observed failures include:

- Cloudflare Worker error: `The Workers runtime canceled this request because it detected that your Worker's code had hung and would never generate a response`
- Cloudflare error page / Worker exception (`1101`)
- In local dev, repeated `WebSocket is closed due to suspension` messages and eventual request failure

### Code Sample

```javascript

```

### Environment

```shell
System:
  OS: macOS 15.7.4
  CPU: (8) arm64 Apple M3
  Memory: 94.33 MB / 24.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 22.22.1
  npm: 10.9.4
  pnpm: 10.28.1
Browsers:
  Brave Browser: 146.1.88.136
  Chrome: 146.0.7680.155
  Firefox: 148.0.2
  Safari: 26.3
  Safari Technology Preview: 26.0
npmPackages:
  @storyblok/astro: ^9.0.0 => 9.0.0 
  astro: ^6.0.8 => 6.0.8
```

### Error Logs

```shell

```

### Additional Context

_No response_

maoberlehner:
On Cloudflare Workers (and comparable edge runtimes such as miniflare and Shopify Oxygen), a backlog of Storyblok requests builds up until the Worker returns `1101` errors. Clearing the backlog restores service briefly, then it rebuilds. Reported in storyblok/monoblok#533 (Astro), storyblok/monoblok#319 and storyblok/monoblok#353 (Hydrogen/Oxygen), and previously in the archived js-client issue storyblok/monoblok#682.

## Root cause

Two of the clients released a throttle slot from a fire-and-forget `setTimeout`: `storyblok-js-client` decremented its in-flight count inside a `setTimeout(interval)` callback, and `@storyblok/management-api-client` used `async-sema`'s `RateLimit`, which releases its token from a `setTimeout(delay)`. Cloudflare Workers suspends the isolate once a response is sent and drops any timer that has not yet fired. Because the client is a cross-request singleton (`@storyblok/astro` stores it on `globalThis`, and the React/Hydrogen integrations follow the same pattern), the dropped release leaked the in-flight count across requests until the throttle wedged, SSR handlers hung, and Cloudflare returned `1101`. (`@storyblok/api-client` released its slot on request completion rather than from a timer, so it did not hang; its only defect was the modeling bug below.)

There was also a modeling bug. Verified against the backend (storyrails), the limits the clients pace against are enforced as **requests per second** in a fixed one-second window: CDN listing tiers `per_page <= 25 -> 50`, `26-50 -> 15`, `51-75 -> 10`, `76-100 -> 6`; published/cached `1000`; Management API `3` (free) / `6` (paid). `@storyblok/api-client` modeled these numbers as a **concurrency cap** instead, which does not bound the per-second rate (many fast requests can complete within one second and still trip the limit). The genuine concurrency cap the backend enforces is a separate limit (default 30 per space), which is the only thing the `X-RateLimit-Policy: "space-concurrent-requests";q=N` header describes.

## Fix

Model the throttle as a **per-second sliding-window rate limiter** and make its waiting Workers-safe. At most `limit` calls may start within any rolling one-second window; once full, further calls wait until the oldest ages out. The window is a list of recent start timestamps pruned against `Date.now()` on every attempt, and every wait resolves the promise the caller already awaits. It therefore keeps no counter that a scheduled callback must decrement, and schedules no timer that outlives the awaited call, so it neither leaks nor deadlocks when the isolate suspends between requests. A non-positive or non-finite `limit` (or, in `storyblok-js-client`, `interval`) disables throttling. Server-side limits remain covered by the existing 429 retry path.

The limiter lives in `src/utils/throttle.ts`, duplicated **verbatim** in `@storyblok/api-client` and `@storyblok/management-api-client` (a deliberate copy in each package rather than a shared runtime dependency, so the independently published clients stay decoupled). `storyblok-js-client` carries the same algorithm in its own `throttlePromise.ts`, which keeps that package's existing callable + `abort()` interface.

Per package:

* **storyblok-js-client** (`throttlePromise.ts`): replaces the concurrency limiter with the per-second window; `abort()` now rejects waiting calls with an `AbortError`.
* **@storyblok/api-client** (capi): replaces the concurrency `createThrottle` with the shared per-second limiter; the tier manager and server-header adaptation are unchanged.
* **@storyblok/management-api-client** (mapi): replaces `async-sema`'s `RateLimit` (which released its token from a fire-and-forget timer, the same Workers hazard) with the shared limiter, and drops the `async-sema` dependency.

## Semantics change

The clients no longer cap simultaneous in-flight requests; they pace requests to the server's per-second budget, which is what the tier numbers always meant. Bursts beyond the window wait rather than being sent and rejected. Recorded in `adr/0011-workers-safe-per-second-request-throttle.md`.

## Notes

* Follow-up option (deferred): also honor the server's `"space-concurrent-requests"` quota as a client-side concurrency cap; the per-second window plus the 429 retry path covers typical usage today.

Fixes storyblok/monoblok#319<br>Fixes storyblok/monoblok#533<br>Fixes storyblok/monoblok#319<br>Fixes storyblok/monoblok#353


packages/js-client/src/throttlePromise.ts
```diff
@@ -7,51 +7,81 @@ class AbortError extends Error {
...
+        if (starts.length < limit) {
+          starts.push(now);
+          resolve();
+          return;
+        }
+
+        // Window is full: retry once the oldest call ages out. The timer only
+        // resolves the promise the caller awaits, so it is never orphaned.
+        const wait = starts[0] + interval - now;
+        let reject_: (reason: unknown) => void;
```
dipankarmaikap, line 71: **Non-blocking:** Is `reject_` the right function name here, or would something like `onReject` be more idiomatic?

packages/capi-client/src/utils/throttle.ts
```diff
@@ -0,0 +1,72 @@
...
+      };
+
+      attempt();
+    });
+  };
+
+  return {
+    execute: <T>(fn: () => Promise<T>): Promise<T> => acquire().then(fn),
+    setLimit: (n: number) => {
+      limit = n;
```
dipankarmaikap, line 77: **Question:** Should we clear `starts` when `setLimit(0)` or `setLimit(Infinity)` is called?

packages/capi-client/src/utils/rate-limit.ts
```diff
@@ -27,9 +29,10 @@ const MAX_RATE_LIMIT = 1_000;
...
   /**
-   * Fixed maximum number of concurrent requests per second.
-   * When set, disables automatic per_page tier detection and all requests
-   * share a single queue at this limit. Capped at 1000.
+   * Fixed number of requests per second. When set, disables automatic
+   * per_page tier detection and all requests share a single per-second
+   * window at this limit. Capped at 1000. Despite the name, this does not
+   * cap simultaneous in-flight requests.
    */
   maxConcurrency?: number;
```
dipankarmaikap, line 42: Should we add a `@deprecated` warning or something?

dipankarmaikap (commented):
Overall, this looks good to me—just a few minor comments.

I'm waiting for the **Publish Any Commit** CI to pass so I can test it on a Cloudflare Workers deployment.

packages/js-client/src/throttlePromise.ts
maoberlehner: reject_ carries the trailing underscore to avoid shadowing the Promise executor's own reject param in the enclosing scope. It's purely internal so I think it is okay to keep as is.

packages/capi-client/src/utils/throttle.ts
maoberlehner: Not needed. Stale timestamps in starts age out within one interval on the next prune, so they never affect a later positive limit. The actual risk is now fixed by re-checking the disabled guard at the top of the attempt, which resolves them.

packages/capi-client/src/utils/rate-limit.ts
maoberlehner: Good catch! Added and made equal to mapi config.

dipankarmaikap (approved):
Great work! 🎉

PR #707, opened 2026-07-23
chore(schema): dogfood Block<TName> component-typing helper in astro playground

maoberlehner:
Playground-only change: dogfoods a userland `Block<TName>` helper for typing framework components against a `@storyblok/schema` schema across the Astro playground. No published package code is affected.

```ts
// playground/astro/src/schema/schema.ts
export type Block<TName extends Blocks['name']> = BlockContent<
  Extract<Blocks, { name: TName }>,
  Blocks,
  FieldPlugins
>;
```

Components now type props by block name (`Block<'hero'>`) instead of the verbose `BlockContent<typeof heroBlock, Blocks, FieldPlugins>` + block-def import. Also renames the `blok` prop/variable to `block` throughout the playground (intentional divergence from the SDK `blok` convention, for clarity in this example app).

## Why

The verbose per-component pattern needed four imports and a block-def reference. `Block<TName>` collapses that to one name-keyed lookup, with `FieldPlugins` baked in so custom-field types resolve automatically. This proves the pattern documented in the companion docs PR.

## Changes

- Add `Block<TName>` and `AnyBlock` exports to the playground `schema.ts`.
- Retype all 6 Astro components (`hero`, `intro`, `media`, `teaser`, `teaser-list`, `storyblok-component`) via `Block<'...'>` / `AnyBlock`.
- Rename `blok` -> `block` prop/variable (dispatcher's inner `.map` var renamed to `child` to avoid shadowing); update both `<StoryblokComponent>` call sites.
- Untouched: `storyblok` import paths, the `bloks` field type, `blocks` field names.

## Companion PR

Docs: storyblok/storyblok-docs-platform#591 - adds a "Typing components" section documenting this helper.

Fixes DX-518

### DIFF
--- a/packages/schema/playground/astro/src/components/storyblok/hero.astro
+++ b/packages/schema/playground/astro/src/components/storyblok/hero.astro
@@ -1,37 +1,31 @@
 ---
-import type { BlockContent } from '@storyblok/schema';
-import type { Blocks, FieldPlugins } from '../../schema/schema';
-import type { heroBlock } from '../../schema/blocks/hero';
-
-// Passing `FieldPlugins` as the third argument resolves the registered
-// `storyblok-colorpicker` custom field to `{ color: string; plugin: string }`.
-type HeroContent = BlockContent<typeof heroBlock, Blocks, FieldPlugins>;
+import type { Block } from '../../schema/schema';
 
 interface Props {
-  blok: HeroContent;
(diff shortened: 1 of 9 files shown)


alexjoverm (approved):
Nice DX nicety!

SirPepe (approved):
LGTM!

PR #709, opened 2026-07-24
fix(richtext): render emoji as unicode span instead of image

dipankarmaikap:
## Summary

Replaces `` emoji rendering with `<span data-emoji="...">🚀</span>` across all renderers (HTML string, Vue, React, Angular).

## Motivation

The previous ``-based rendering had two bugs:
- `alt` was always `undefined` — broken accessibility
- missing `fallbackImage` produced `src="null"` — spurious network request

Modern browsers render Unicode emoji natively. Users who want image-based rendering can override via a custom component/renderer.

## Changes

- **`nodes.ts`** — `renderHTML` now emits `<span data-emoji="🚀" data-name="rocket">🚀</span>`; `addAttributes` unchanged so `` still round-trips via `parseHTML`
- **`render-map.generated.ts`** — regenerated; emoji entry is now `{ tag: 'span' }`
- **`attribute.ts`** — `fallbackImage` added to `EXCLUDED_ATTRS` and removed from `DEFAULT_ATTR_MAP` (prevents `src="..."` leaking onto the span)
- **`richtext-attrs.ts`** — added `NODES_WITH_REQUIRED_ATTRS` so that `emoji`, `blok`, and `image` nodes have required `attrs` in the discriminated union
- **`richtext-type.ts`** (generator) — emits `attrs:` (required) for nodes in `NODES_WITH_REQUIRED_ATTRS`, so `node.type === 'emoji'` alone narrows `node.attrs.emoji` to `string`
- **`types.generated.ts`** — regenerated; `emoji`, `blok`, and `image` union members now have required `attrs`
- **`util.ts`** — `getEmojiText` simplified to use discriminated union narrowing directly; no separate type guard needed
- **`render-richtext.ts`** — XSS fix: emoji text is passed through `escapeHtml` before interpolation into the HTML string
- **All four renderers** — use `node.type === 'emoji'` narrowing to access `node.attrs.emoji` directly
- **`util.test.ts`** — new unit tests for `getEmojiText`

## Behaviour fix note

Rendered HTML changes from `` to `<span data-emoji="...">🚀</span>`.

**Migration (one selector change):**
```css
/* before */
img[data-emoji] { ... }

/* after — also works on old output */
[data-emoji] { ... }
```

This is a behaviour fix, not a breaking change: the old rendering was actively buggy (`alt="undefined"`, `src="null"`). JSON schema and HTML round-trip parsing are unaffected. Shipping as a minor bump.

Fixes DX-516

### DIFF
--- a/packages/richtext/src/render-richtext.ts
+++ b/packages/richtext/src/render-richtext.ts
@@ -1,6 +1,6 @@
 import { escapeHtml } from './utils';
 import { optimizeImage } from './images-optimization';
-import { areLinkMarksEqual, attrsToHtmlString, getStaticChildren, getTextNodeLinkMark, isSelfClosing, isTableHeaderRow, normalizeNodes, processAttrs, resolveTag, styleToString } from './static';
+import { areLinkMarksEqual, attrsToHtmlString, getEmojiText, getStaticChildren, getTextNodeLinkMark, isSelfClosing, isTableHeaderRow, normalizeNodes, processAttrs, resolveTag, styleToString } from './static';
 import type { RenderSpec, SbRichTextElement, SbRichTextInput, SbRichTextMark, SbRichTextNode, SbRichTextRenderContext, SbRichTextTextNode } from './static';
 /**
  * Renders a Storyblok RichText JSON document to an HTML string.
@@ -81,7 +81,8 @@ function renderNode(node: SbRichTextNode, context?: SbRichTextRenderContext): st
     return `<${tag}>${inner}</${tag}>`;
   }
 
-  return `<${tag}${htmlAttrs}>${content}</${tag}>`;
+  const textContent = getEmojiText(node);
+  return `<${tag}${htmlAttrs}>${textContent ?? content}</${tag}>`;
 }
 
 /** Renders an image node with optimization applied. */
--- a/packages/richtext/src/render-richtext.ts
+++ b/packages/richtext/src/render-richtext.ts
@@ -1,6 +1,6 @@
 import { escapeHtml } from './utils';
 import { optimizeImage } from './images-optimization';
-import { areLinkMarksEqual, attrsToHtmlString, getEmojiText, getStaticChildren, getTextNodeLinkMark, isSelfClosing, isTableHeaderRow, normalizeNodes, processAttrs, resolveTag, styleToString } from './static';
+import { areLinkMarksEqual, attrsToHtmlString, getStaticChildren, getTextNodeLinkMark, isSelfClosing, isTableHeaderRow, normalizeNodes, processAttrs, resolveTag, styleToString } from './static';
 import type { RenderSpec, SbRichTextElement, SbRichTextInput, SbRichTextMark, SbRichTextNode, SbRichTextRenderContext, SbRichTextTextNode } from './static';
 /**
  * Renders a Storyblok RichText JSON document to an HTML string.
@@ -81,8 +81,10 @@ function renderNode(node: SbRichTextNode, context?: SbRichTextRenderContext): st
     return `<${tag}>${inner}</${tag}>`;
   }
 
-  const textContent = getEmojiText(node);
-  return `<${tag}${htmlAttrs}>${textContent ?? content}</${tag}>`;
+  if (node.type === 'emoji') {
+    return `<${tag}${htmlAttrs}>${node.attrs.emoji}</${tag}>`;
+  }
+  return `<${tag}${htmlAttrs}>${content}</${tag}>`;
 }
 
 /** Renders an image node with optimization applied. */
--- a/packages/richtext/src/render-richtext.ts
+++ b/packages/richtext/src/render-richtext.ts
@@ -82,7 +82,7 @@ function renderNode(node: SbRichTextNode, context?: SbRichTextRenderContext): st
   }
 
   if (node.type === 'emoji') {
-    return `<${tag}${htmlAttrs}>${node.attrs.emoji}</${tag}>`;
+    return `<${tag}${htmlAttrs}>${escapeHtml(node.attrs.emoji)}</${tag}>`;
   }
   return `<${tag}${htmlAttrs}>${content}</${tag}>`;
 }
--- a/packages/richtext/src/static/util.ts
+++ b/packages/richtext/src/static/util.ts
@@ -75,3 +75,22 @@ export function attrsToHtmlString(attrs: Record<string, unknown>): string {
 
   return result;
 }
+
+/**
+ * Returns the static text content for leaf nodes whose text comes from attrs
+ * rather than child nodes (e.g. emoji).
+ *
+ * @param node - The RichText node
+ * @returns The text string, or null if the node has no attr-based text content
+ *
+ * @example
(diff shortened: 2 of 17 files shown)


packages/richtext/src/render-richtext.ts
```diff
@@ -81,7 +81,8 @@ function renderNode(node: SbRichTextNode, context?: SbRichTextRenderContext): st
     return `<${tag}>${inner}</${tag}>`;
   }
 
-  return `<${tag}${htmlAttrs}>${content}</${tag}>`;
+  const textContent = getEmojiText(node);
+  return `<${tag}${htmlAttrs}>${textContent ?? content}</${tag}>`;
```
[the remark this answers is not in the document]
alexjoverm: @dipankarmaikap this one looks relevant

packages/richtext/src/static/util.ts
```diff
@@ -75,3 +75,22 @@ export function attrsToHtmlString(attrs: Record<string, unknown>): string {
...
+ * rather than child nodes (e.g. emoji).
+ *
+ * @param node - The RichText node
+ * @returns The text string, or null if the node has no attr-based text content
+ *
+ * @example
+ * getEmojiText({ type: 'emoji', attrs: { emoji: '🚀' } }) // '🚀'
+ * getEmojiText({ type: 'paragraph', content: [...] })      // null
+ */
+export function getEmojiText(node: SbRichTextNode): string | null {
```
alexjoverm: suggestion(non-blocking): worth to have a small unit test case for this

packages/richtext/src/render-richtext.ts
```diff
@@ -81,7 +81,8 @@ function renderNode(node: SbRichTextNode, context?: SbRichTextRenderContext): st
     return `<${tag}>${inner}</${tag}>`;
   }
 
-  return `<${tag}${htmlAttrs}>${content}</${tag}>`;
+  const textContent = getEmojiText(node);
```
alexjoverm: @dipankarmaikap a Claude finding, feel free to decide if you see it relevant.

--

In the HTML string renderer, `getEmojiText(node)` is checked *after* `getStaticChildren`, and uses `textContent ?? content`. Since emoji no longer has static children (the render map entry changed from `{ tag: 'img', attrs: {...} }` to `{ tag: 'span' }`), the `getStaticChildren` path will return `null` and execution correctly falls through to the final `return` where `textContent` is set. The flow is correct.

However, if a user registers a custom renderer for `emoji` (`context.renderers.emoji`), it fires on line 37-43 with `content = ''` (empty string, since emoji has no child nodes). The `getEmojiText` path is never reached in that case, which means the custom renderer receives `children: ''` instead of `children: '🚀'`. This is consistent with the old behavior (where `content` was also `''` for emoji), but it means custom emoji renderers must extract the emoji from `node.attrs.emoji` themselves rather than relying on `children`. This is worth documenting.

alexjoverm (changes requested):
@dipankarmaikap overall looks great! Just a few small fixes needed.

In the PR description, there is a `Breaking change` section. As this is a bugfix, I would rename it as otherwise can be confusing for the reader. If this would be a breaking change, we'd need to publish it as a major, but we're rather considering it a behaviour-fix as the expected behavior is not the current one.

Aside from that, couple of comments to revisit!

packages/richtext/src/static/util.ts
```diff
@@ -75,3 +75,22 @@ export function attrsToHtmlString(attrs: Record<string, unknown>): string {
...
+  if (node.type === 'emoji') {
```
maoberlehner: nitpick: I always find `if` statements like these a little odd. If `node` isn't an emoji, why even call `getEmojiText` on it in the first place? It's mixing two concerns: checking the type of `node` and getting the emoji.

You might consider splitting it into two functions, one for determining the type and the other for getting the emoji text if it is an emoji:

```ts
// 1. Determine the type (Type Guard)
export function isEmojiNode(
  node: SbRichTextNode
): node is SbRichTextNode & { type: 'emoji'; attrs: { emoji: string } } {
  if (node.type !== 'emoji') {
    return false;
  }
  
  const attrs = (node as Record<string, unknown>).attrs as Record<string, unknown> | undefined;
  return typeof attrs?.emoji === 'string';
}

// 2. Extract the data (Precondition: node is already known to be an emoji)
export function getEmojiText(
  node: SbRichTextNode & { type: 'emoji'; attrs: { emoji: string } }
): string {
  return node.attrs.emoji;
}

// Or maybe just skip `getEmojiText`:
```

```astro
  <RenderTable node={node} options={options} />
) : isEmojiNode(node) ? (
  <ElementTag node={node}>{node.attrs.emoji}</ElementTag>
) : getStaticChildren(node) ? (
```

Also see: https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html

maoberlehner: The even better solution would be if we had more strongly typed nodes to begin with, so that each `node.type === 'emoji'` always has `node.attrs.emoji`.

alexjoverm (changes requested):
@dipankarmaikap one more finding from Claude that can be worth exploring. Not fully sure if the case it's often reproducible, but good to be guarded regardless:

----

### Critical: Runtime crash on malformed emoji node (missing `attrs`)

**Files:**
- `packages/richtext/src/render-richtext.ts:84`
- `packages/vue/src/rich-text-renderer.ts:169`
- `packages/react/src/core/rich-text-renderer.tsx:149`
- `packages/angular/src/lib/richtext/rich-text.component.ts:336`
- `packages/astro/src/components/richtext/RenderNode.astro:47`
- `packages/svelte/src/lib/richtext/RenderNode.svelte:49`

**Problem:** All six renderers access `node.attrs.emoji` without any runtime guard. The safety relies entirely on the generated type making `attrs` required for emoji nodes. However, the runtime data comes from the Storyblok API. If the API ever returns an emoji node without `attrs` (malformed data, partial content, or forward-compatibility edge cases), every renderer will throw `TypeError: Cannot read properties of undefined (reading 'emoji')`.

**Suggestion:** Add a lightweight runtime guard. For example, in the HTML string renderer:

```typescript
if (node.type === 'emoji') {
  const emoji = node.attrs?.emoji;
  if (emoji) {
    return `<${tag}${htmlAttrs}>${escapeHtml(emoji)}</${tag}>`;
  }
  return `<${tag}${htmlAttrs}></${tag}>`;
}
```

Apply the equivalent pattern in all six renderers. This preserves the type narrowing benefit while being defensively safe at runtime.

dipankarmaikap:
@alexjoverm I don't think it's necessary. This would apply to all other nodes and marks as well. If the product changes this behaviour in the future, it would be a breaking change, and we'd need to release a fix anyway.

alexjoverm:
And the attrs are always fullfilled, yes? even if it's an empty object.<br><br>Then approving as that was the only thing

alexjoverm (approved):

packages/richtext/src/static/util.ts
dipankarmaikap: Fixed.

packages/richtext/src/static/util.ts
dipankarmaikap: This is removed, so no need to add a test

PR #710, opened 2026-07-24
feat(openapi): type rich text field value — spec, aliases, naming

dipankarmaikap:
## Summary

Extracts the `tools/openapi-codegen` changes from #684 into a standalone PR so the spec work can be reviewed and merged independently.

### `tools/openapi-codegen`

**Spec (`specs/`)**
- `richtext-field-value.yaml` — expanded from a loose `type: object` into a fully discriminated schema with `$defs` for all 17 node types and 12 mark types
  - `BlockNode.attrs.body` items reference `../block-content.yaml` (canonical definition) instead of an inline duplicate
- `overlay.openapi.yaml` — added `RichTextNode` and `RichTextMark` as top-level schema refs so they are emitted as named types; renamed `RichtextFieldValue` → `RichTextFieldValue`
- `mapi/components/field-types/richtext-field.yaml` — description updated to follow naming conventions

**Source (`src/`)**
- `aliases.ts` — registered `RichTextDoc`, `RichTextNode`, `RichTextMark`; updated `RichtextFieldValue` → `RichTextFieldValue`
- `index.ts` — fixed `copyWrapperTemplates` to skip writing the empty `types/_sources.ts` when no wrapper templates are requested
- `known-types.ts`, `templates.ts` — updated `RichtextFieldValue` → `RichTextFieldValue`; registered `RichtextFieldValue` in `provides` for backward compat

**Templates (`templates/`)**
- `field.ts` — updated import/export/map entry for `RichTextFieldValue`; added deprecated `RichtextFieldValue` alias

## Backward compatibility

`RichtextFieldValue` has been renamed to `RichTextFieldValue`. To avoid breaking consumer packages all at once, the generated `field.ts` now also exports:

```ts
/** @deprecated Use RichTextFieldValue instead. Will be removed in a future major version. */
export type RichtextFieldValue = RichTextFieldValue;
```

This means packages that currently import `RichtextFieldValue` (`mapi-client`, `schema`, `capi-client`, `migrations`, etc.) will continue to compile after regeneration — TypeScript/editors will surface the `@deprecated` warning so each package can migrate at its own pace.

## Naming conventions applied

| Context | Convention | Example |
|---|---|---|
| Type names | `RichText` (PascalCase) | `RichTextFieldValue`, `RichTextDoc` |
| Sentence start | `Rich text` | `Rich text field type …` |
| Mid-sentence | `rich text` | `… rich text document …` |
| Wire values | unchanged | `enum: [richtext]`, `toggle-richtext` |
| Node type names | `Block` not `Blok` | `BlockNode` |

Fixes DX-487


tools/openapi-codegen/specs/shared/stories/field-types/rich-text-field-value.yaml
```diff
@@ -0,0 +1,564 @@
...
+                required: [alt, title, source, copyright]
+                properties:
+                  alt:
+                    type: [string, 'null']
+                  title:
+                    type: [string, 'null']
+                  source:
+                    type: [string, 'null']
+                  copyright:
+                    type: [string, 'null']
```
maoberlehner, line 477: Those are all optional, and custom ones are allowed (`Custom metadata fields` in the UI: https://app.storyblok.com/#/me/org/settings?tab=shared_assets_settings)

tools/openapi-codegen/specs/shared/stories/field-types/rich-text-field-value.yaml
```diff
@@ -1,18 +1,580 @@
...
+              - type: 'null'
+      content:
+        type: array
+        items:
+          $ref: '#/$defs/RichTextNode'
+      marks:
+        type: array
+        items:
+          $ref: '#/$defs/RichTextMark'
+
```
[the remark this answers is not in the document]
maoberlehner: Correct, but let's ignore that. Yes, storyrails does not guarantee that but storyfront does. And we rely on correct data that is not guaranteed for other fields already too. But maybe we should document this decisions as a comment in here?

tools/openapi-codegen/specs/shared/stories/field-types/rich-text-field-value.yaml
```diff
@@ -0,0 +1,564 @@
...
+    properties:
+      type:
+        type: string
+        enum: [highlight]
+      attrs:
+        type: object
+        required: [color]
+        properties:
+          color:
+            type: string
```
maoberlehner: should be `[string, 'null']`

tools/openapi-codegen/specs/shared/stories/field-types/rich-text-field-value.yaml
```diff
@@ -0,0 +1,564 @@
...
+        enum: [image]
+      attrs:
+        type: object
+        required: [src, id, alt, title, source, copyright, meta_data]
+        properties:
+          id:
+            type: [integer, 'null']
+            description: Storyblok asset ID
+          src:
+            type: string
```
maoberlehner: Must be nullable

tools/openapi-codegen/specs/shared/stories/field-types/rich-text-field-value.yaml
```diff
@@ -0,0 +1,564 @@
...
+    type: object
+    required: [type]
+    properties:
+      type:
+        type: string
+        enum: [paragraph]
+      attrs:
+        type: object
+        required: [textAlign]
+        properties:
```
maoberlehner, line 231: `class` is missing

tools/openapi-codegen/specs/shared/stories/field-types/rich-text-field-value.yaml
```diff
@@ -0,0 +1,564 @@
...
+    type: object
+    required: [type, attrs]
+    properties:
+      type:
+        type: string
+        enum: [heading]
+      attrs:
+        type: object
+        required: [textAlign, level]
+        properties:
```
maoberlehner, line 278: `class` is missing

tools/openapi-codegen/specs/shared/stories/field-types/rich-text-field-value.yaml
```diff
@@ -0,0 +1,564 @@
...
+              - type: string
+                enum: [_self, _blank, _parent, _top]
+              - type: 'null'
+          linktype:
+            description: Type of link
+            oneOf:
+              - type: string
+                enum: [story, url, email, asset]
+              - type: 'null'
+          custom:
```
maoberlehner, line 73: should be nullable

maoberlehner (changes requested):

maoberlehner:
- `pnpm nx run-many -t generate` leads to uncommitted changes!
- `pnpm nx run @storyblok/schema:build --skip-nx-cache` shows `[MISSING_EXPORT] Warning` the alias must also be added to Zod output (internal-schemas.ts + validate-story.ts).
- `'heading', 'paragraph', 'list_item', 'blockquote', 'code_block'` need `dir: 'ltr' | 'rtl' | null`

tools/openapi-codegen/specs/shared/stories/field-types/rich-text-field-value.yaml
dipankarmaikap: I don't think that's true. If you add any class to a heading, they're applied as marks on the heading rather than as direct attributes of the heading itself.

tools/openapi-codegen/specs/shared/stories/field-types/rich-text-field-value.yaml
dipankarmaikap: Same as heading, dont think its valid.

tools/openapi-codegen/specs/shared/stories/field-types/rich-text-field-value.yaml
dipankarmaikap: @maoberlehner Can you please tell me which attribute you are talking about here?

dipankarmaikap:
> * `pnpm nx run-many -t generate` leads to uncommitted changes!

Should we do this in this PR?

packages/migrations/src/types.ts
```diff
@@ -5,6 +5,7 @@ export type {
   Component,
   Datasource,
   MultilinkFieldValue,
+  RichTextFieldValue,
   RichtextFieldValue,
```
maoberlehner, line 9: Can we mark this (and maybe others in other places) as deprecated?

tools/openapi-codegen/specs/shared/stories/field-types/rich-text-field-value.yaml
maoberlehner: This `custom` one `type: [object, 'null']`

tools/openapi-codegen/specs/shared/stories/field-types/rich-text-field-value.yaml
maoberlehner: You're right!

tools/openapi-codegen/specs/shared/stories/field-types/rich-text-field-value.yaml
```diff
@@ -0,0 +1,601 @@
...
+        properties:
+          id:
+            type: [string, 'null']
+            description: Block instance ID
+          body:
+            description: Array of embedded component instances
+            oneOf:
+              - type: array
+                items:
+                  $ref: ../block-content.yaml
```
maoberlehner: `block-content.yaml` is the read variant with mandatory `_uid` on blocks. But for write operations, this should be `block-content-input.yaml` (`_uid` optional).

Not sure how to best solve this:

a) Leave it like it is and accept that people need to write `_uid: ""` or `_uid: "anything"`
b) Change it to `block-content-input.yaml` and accept that when reading, people need to check `block?._uid` if they need it (although in 99% of cases people probably don't care).
c) Introduce `rich-text-field-value-input.yaml` which would lead to a lot of duplication in our code even after extracting what we can extract.

I leave it to you to decide which one is the best trade-off.

maoberlehner:
> > * `pnpm nx run-many -t generate` leads to uncommitted changes!
> 
> Should we do this in this PR?

Yes, absolutely!

maoberlehner (approved):

packages/migrations/src/types.ts
dipankarmaikap: Yes, we already added the `@deprecated` JSDoc annotation to `RichtextFieldValue`. Since it's re-exported from `mapi-client`, the deprecation warning will show up here as well.

tools/openapi-codegen/specs/shared/stories/field-types/rich-text-field-value.yaml
dipankarmaikap: Going with Option `b` here.

PR #713, opened 2026-07-27
test: demonstrate bugs by adding a kitchen sink schema + component

jbasoo:
## Objective
Add a kitchen_sink block and component to the @storyblok/schema Astro playground using one field of every Storyblok field type as a test fixture for the schema package. This PR is purely to demonstrate bugs, not actually be merged.

Field-type coverage was previously scattered across 14 blocks, with several types never rendered by any component — so nothing forced the resolved types to be checked against a consumer. The component here consumes each field exactly one way, with no workarounds: where that doesn't compile, the type error is left in place as the finding.

## Findings
Three defects surfaced, all documented inline:

1. `richtext` cannot be rendered — @storyblok/schema types don't compose with @storyblok/richtext. This is the one intentional type error on the branch.
2. `number` is typed number but the API requires and returns a string — silently fails tests but it meant the committed showcase and kitchen_sink stories were not pushable.

## Verification
Schema pushed and seeded against a live space; story renders all fields.

### DIFF
--- a/packages/schema/playground/astro/src/components/storyblok/kitchen-sink.astro
+++ b/packages/schema/playground/astro/src/components/storyblok/kitchen-sink.astro
@@ -27,41 +27,36 @@ import Teaser from './teaser.astro';
  * | ------------ | ------------------------------------------------------------- |
  * | `text`       | `string`                                                      |
  * | `textarea`   | `string`                                                      |
- * | `richtext`   | `{ type: 'doc'; content?: { [key: string]: unknown }[] }`      |
+ * | `richtext`   | `{ type: 'doc'; content?: { [key: string]: unknown }[] }`     |
  * | `markdown`   | `string`                                                      |
  * | `number`     | `number`                                                      |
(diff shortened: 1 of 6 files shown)


packages/schema/playground/astro/src/components/storyblok/kitchen-sink.astro
```diff
@@ -0,0 +1,293 @@
...
+type KitchenSinkContent = BlockContent<typeof kitchenSinkBlock, Blocks, FieldPlugins>;
+
+interface Props {
+  blok: KitchenSinkContent;
+}
+
+const { blok } = Astro.props;
+
+// richtext type incompatible with renderRichText — see gap 1 above.
+const richtextHtml = blok.richtext ? renderRichText(blok.richtext) : null;
```
jbasoo, line 91: This is the main blocker I've found so far trying to implement the schema package on another project. The richtext type is a shallow doc, not recursive so not walkable and incompatible with existing code as well as the storyblok/richtext package.

maoberlehner:
Thank you for this PR! It proved highly valuable to us for improving the schema package!

PR #714, opened 2026-07-28
fix: standarize stderr for human output, consolidate printing methods

alexjoverm:
## Summary

- **All CLI UI output now goes to stderr**, reserving stdout for data
- **Single output module** — every command uses `getUI()`, no alternatives remain
- **Standardized exit codes** — `1` for runtime/API errors, `2` for usage/validation errors

## Why stderr for everything, not just errors?

`stderr` is misnamed. [POSIX defines it](https://pubs.opengroup.org/onlinepubs/9699919799/functions/stdin.html) as the stream for "diagnostic output" — not "error output". The [CLI Guidelines (clig.dev)](https://clig.dev/) put it clearly: *"Log messages, errors, and so on should all be sent to stderr, so that when commands are piped together, these messages are displayed to the user and not fed into the next command."*

Without this, piping breaks:

```bash
# Before: JSON file contains spinner text, titles, success messages
storyblok components pull -s 12345 > components.json

# After: only data hits the file, UI still shows in terminal
storyblok components pull -s 12345 > components.json
```

In Node.js, `console.error()` writes to stderr — that's why the UI module delegates everything through it, not because the output is an error.

## Why delete konsola instead of deprecating?

`konsola` was a thin wrapper around `console.*` with formatting. `getUI()` already does the same thing with more features (spinners, progress bars, disabled mode for tests). Keeping both means contributors have to choose between two APIs that do the same thing. Deleting it removes the ambiguity — there's one way to print, documented in `CLAUDE.md`.

Same reasoning for `ProgressDisplay` — it reimplemented progress bars that `getUI().createProgressBar()` already provides.

## Why does handleError not use getUI()?

`handleError` runs before `preAction` initializes the UI singleton, so it can't depend on `getUI()`. It uses its own `stderr` wrapper (a two-line object delegating to `console.error`) for the same effect.

## If you're working on a parallel branch

- Replace `import { konsola } from '../../utils/konsola'` with `import { getUI } from '../../lib/ui'`
- `konsola.warn(msg)` → `ui.warn(msg)`, `konsola.error(msg)` → `handleError(new CommandError(msg))`
- `console.log` / `console.info` for user messages → `ui.info()`, `ui.ok()`, `ui.log()`
- The UI module moved from `src/utils/ui.ts` to `src/lib/ui/` — update import paths

Fixes DX-521


packages/cli/src/commands/datasources/delete/index.test.ts
```diff
@@ -1,10 +1,7 @@
-import { konsola } from '../../../utils';
 import { deleteDatasource } from './actions';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import '../index';
 import { datasourcesCommand } from '../command';
-import chalk from 'chalk';
-import { colorPalette } from '../../../constants';
 import { fetchDatasource } from '../pull/actions';
 import { confirm } from '@inquirer/prompts';
```
maoberlehner, line 6: Interactive prompts with inquirer still write to stdout. Idk how relevant this is, as interactive prompts are not really usable when piping anyway, but it can be easily fixed with `{ output: process.stderr }`.

packages/cli/src/program.ts
maoberlehner: This still exists with `1`, although, as I understand it, this category of errors should exit `2`.

maoberlehner (changes requested):

maoberlehner:
`components push --space 290020686946333` on main produces a different output than on this branch, and it does not end, but I need to kill it with `ctrl + c`:

```bash
# main
  Storyblok CLI


 Components  Pushing components...


ℹ Attempting to push components from space 290020686946333 to 290020686946333


ℹ Using graph-based dependency resolution

📊 Built dependency graph with 4 nodes
✅ Graph validation passed

Processing 4 resources...
✓ Component→ cta - Updated (66ms)
✓ Component→ blog - Updated (66ms)
✓ Component→ hero - Updated (893ms)
✓ Component→ page - Updated (76ms)

 Success

✔ Completed: 4 updated, 0 unchanged, 0 failed in 970ms

8 requests made

ℹ Components reference datasources: categories

ℹ To manage datasources, use: storyblok datasources push
```

```bash
# this branch
  Storyblok CLI


 Components  Pushing components...


ℹ Attempting to push components from space 290020686946333 to 290020686946333


ℹ Using graph-based dependency resolution

 Pushing  [■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■          ] 75% | 1s | 3/4 processed

 Success

✔ Completed: 4 updated, 0 failed in 1.0s

ℹ Components reference datasources: categories

ℹ To manage datasources, use: storyblok datasources push

 Pushing  [■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■] 100% | 0s | 4/4 processed
```

maoberlehner:
Claude findings:

Help text goes to stdout on error paths. writeOut is not configured, and program.on('command:*') calls program.outputHelp() → stdout. storyblok bogus-command > out.json writes 3112 bytes of help into the file. Use outputHelp({ error: true }) there. (--help on stdout is correct and should stay.)

The PR description's headline example doesn't work. storyblok components pull -s 12345 > components.json yields a 0-byte file — pull writes to .storyblok/components/<space>/components.json and never touches stdout. No command emits data to stdout today. Reserving stdout is the right foundation, but the description claims behavior that doesn't exist; I'd reword it or add a genuine stdout-emitting path.

--no-ui-enabled has no effect — output is identical with and without it. Relevant here because the PR adds ui.test.ts coverage for disabled mode while the flag that would trigger it is dead.

types generate without --space reports No local components found for space undefined as a FileSystemError → exit 1; a missing required option should be a CommandError → 2.

--version isn't recognized (only -v, --vers) and exits 1.

assets/push/index.ts:182 and assets/transfer/index.ts:81 force process.exitCode = 2 after handleError(toError(maybeError)) — a generic runtime error getting the usage code, contradicting the new scheme. Cheap to align while standardizing.

packages/cli/src/utils/auth.ts
```diff
@@ -21,6 +21,7 @@ export function requireAuthentication(state: SessionState, verbose = false): sta
       new CommandError(`You are currently not logged in. Please run ${chalk.hex(colorPalette.PRIMARY)('storyblok login')} to authenticate, or ${chalk.hex(colorPalette.PRIMARY)('storyblok signup')} to sign up.`),
       verbose,
     );
+    process.exitCode = 2;
```
maoberlehner: `handleError` above already sets the exit code. Is this on purpose?

packages/cli/src/utils/error/error.ts
```diff
@@ -50,20 +56,20 @@ function handleVerboseError(error: unknown): void {
   if (error instanceof CommandError || error instanceof APIError || error instanceof FileSystemError) {
     const errorDetails = 'getInfo' in error ? error.getInfo() : {};
     if (error instanceof CommandError) {
-      konsola.error(`Command Error: ${error.getInfo().message}`, errorDetails);
+      stderr.error(`${chalk.red.bold('▲ error')} Command Error: ${error.getInfo().message}`, errorDetails);
```
maoberlehner: This (`▲`, `ℹ`) error formatting is also in `ui.ts`. Maybe we can move it to a single place?

packages/cli/src/commands/datasources/delete/index.test.ts
alexjoverm: Great point, they def should go through stderr!

packages/cli/src/program.ts
alexjoverm: `handleError` would check the kind of error, and exit with `2` or `1` dependending on the case. Check https://github.com/storyblok/monoblok/pull/714/changes#diff-52fdd9fc80336f4d1d9a3b30b245fd73662b479b1b82f82f9bc7eebf33b3936bR108-R110

packages/cli/src/utils/auth.ts
alexjoverm: Good point. Likely a leftover as now handleError also runs the exitCode

packages/cli/src/utils/error/error.ts
alexjoverm: Probably we can reuse the UI module here as well. Problem is that error.ts can run before ui one gets initialized, but will look into a way to avoid that

packages/cli/src/commands/assets/push/index.test.ts
```diff
@@ -934,7 +932,7 @@ describe('assets push command', () => {
     expect(actions.updateAsset).not.toHaveBeenCalled();
     const logFile = getLogFileContents(LOG_PREFIX);
     expect(logFile).toContain('Unexpected token');
-    expect(process.exitCode).toBe(2);
+    expect(process.exitCode).toBe(1);
```
maoberlehner, line 935: I've noticed that we never reset the exit code, maybe we should reset it in `packages/cli/test/setup.ts`?

packages/cli/src/index.ts
```diff
@@ -20,21 +21,24 @@ import './commands/stories';
...
+console.error('');
 
 // Handle invalid commands
 program.on('command:*', () => {
   console.error(`Invalid command: ${program.args.join(' ')}`);
-  konsola.br();
-  program.outputHelp();
-  process.exit(1);
+  console.error('');
+  program.outputHelp({ error: true });
```
maoberlehner: This now renders `▲ error Usage: …` plus a `logger.error` record containing the whole help text.

packages/cli/src/lib/logger/logger-transport-console.ts
```diff
@@ -26,10 +26,10 @@ export class ConsoleTransport implements LogTransport {
         (console.warn ?? console.log).call(console, line);
```
maoberlehner: Should we use `console.error` here and above always too? If not, why not?

maoberlehner (changes requested):

maoberlehner:
Claude findings:

0. PR body is stale: `handleError` *does* use `getUI()` now (`packages/cli/src/utils/error/error.ts:72`), so the "own stderr wrapper" paragraph no longer applies.

**1. Commander-detected usage errors exit `1`, not `2` — the PR's headline exit-code contract is not met.**

`program.ts:174` wraps commander's `writeErr` output in a `CommandError`, and `handleError` sets `process.exitCode = 2`. But commander's `error()` calls `writeErr` and then immediately `process.exit(1)`, which discards the `exitCode` that was just set. Measured:

| Invocation                                                      | Exit  | Expected |
| --------------------------------------------------------------- | ----- | -------- |
| `storyblok --vers` (unknown option)                             | **1** | 2        |
| `storyblok components pull --nope`                              | **1** | 2        |
| `storyblok assets pull -s <id> --target bogus` (invalid choice) | **1** | 2        |
| `storyblok components pull --space` (missing option arg)        | **1** | 2        |
| `storyblok boguscmd` (own `command:*` handler)                  | 2     | 2 ✓      |
| `storyblok components` (missing subcommand)                     | 2     | 2 ✓      |
| `storyblok components pull` (missing `--space`, app-thrown)     | 2     | 2 ✓      |

**2. `--no-ui-enabled` silences error messages entirely — failures become invisible.**

```
$ storyblok components pull --no-ui-enabled          # missing --space
exit=2, stderr = 22 bytes (just the banner). No message.
$ storyblok components pull --space 1 --no-ui-enabled # API 404
exit=1, stderr = 22 bytes. No message.
```

`handleError` routes through `getUI()`, so disabling UI drops diagnostics too. `--no-ui-enabled` is the natural CI flag, and CI is exactly where you need the error text. Errors should bypass the `enabled` gate (or the flag should only suppress spinners/bars/decoration, not messages).

**3. The startup banner bypasses the UI module — the one thing the PR set out to eliminate.**

`index.ts:29-33` prints the banner with five raw `console.error` + `chalk` calls, hand-reimplementing `ui.title()`. Consequences: it ignores `--no-ui-enabled` (the 22 bytes above), and it prints on `--version` and `--help`. Given the PR's stated goal — "single output module, no alternatives remain" — this should be `getUI().title(...)`, deferred until after `preAction` resolves config.

**4. `--no-ui-enabled` is honored inconsistently.** Post-`preAction` errors respect it (finding 2); commander parse errors, which fire before the UI singleton is reconfigured, print the full formatted error. Same flag, opposite behavior depending on when the error occurs.

**5. Behavior change: aborting an interactive prompt now exits non-zero, inconsistently.** Cancelling `login` exits **1** (was `0` on `main`) and surfaces the raw inquirer string as `▲ error User force closed the prompt with 0 null`. Cancelling `datasources delete` exits **0** ("Deletion aborted by user"). Pick one — user-initiated cancellation is conventionally `130` (SIGINT) or `0`, and the raw inquirer message shouldn't reach users.nfigured, print the full formatted error. Same flag, opposite behavior depending on when the error occurs.

packages/cli/src/commands/assets/push/index.test.ts
alexjoverm: Good one, would prevent side effects from happening.

packages/cli/src/lib/logger/logger-transport-console.ts
alexjoverm: Yeah, this part now can be pretty much refactored!

packages/cli/src/commands/components/pull/index.ts
```diff
@@ -59,41 +59,63 @@ pullCmd
...
+      logger.info('Fetched groups', { count: groups?.length ?? 0 });
+      barGroups.increment();
+
+      let presets = presetsResult;
+      logger.info('Fetched presets', { count: presets?.length ?? 0 });
+      barPresets.increment();
+
+      let internalTags = tagsResult;
+      logger.info('Fetched tags', { count: internalTags?.length ?? 0 });
+      barTags.increment();
```
maoberlehner, line 98: nitpick: not quite sure about this pattern as I find progress bars that jump from 0 to 100 in one increment a little counterintuitive. Could we do something different, like the classic progress bar trick of progressing from `0` to `rand(70, 85)` in ~2secs in 1% increments, wait for all requests to finish, and then increment from `rand(70, 85)` to 100 in ~0.5sec with a single %-based loading bar for all?

Or do one progress bar for all and increment once for each request.

packages/cli/src/utils/error/error.ts
```diff
@@ -47,54 +47,78 @@ export function getResponseStatus(maybeError: unknown): number | undefined {
...
   }
   else {
-    konsola.br();
-    konsola.info('For more information about the error, run the command with the `--verbose` flag');
+    // Always print the verbose hint to stderr, even with --no-ui-enabled (CI)
+    console.error('');
+    console.error(`For more information about the error, run the command with the \`--verbose\` flag`);
   }
 
   if (!process.env.VITEST) {
```
maoberlehner: nitpick: as it was already here before, feel free to ignore my comment, but why do we even do that?

packages/cli/src/commands/components/push/graph-operations/resource-processor.ts
```diff
@@ -1,57 +1,85 @@
...
-import chalk from 'chalk';
 import { getActiveConfig } from '../../../../lib/config';
+import { getUI } from '../../../../lib/ui';
+import { getLogger } from '../../../../lib/logger/logger';
+
+interface ProgressBar {
+  increment: (count?: number) => void;
+  setTotal: (total: number) => void;
+  stop: () => void;
+}
```
maoberlehner: nitpick: can we get this from `lib/ui`?

packages/cli/src/commands/create/index.ts
```diff
@@ -127,21 +127,14 @@ export const createCommand = program
...
     try {
-      spinnerBlueprints.start('Fetching starter templates...');
+      activeSpinner = ui.createSpinner('Fetching starter templates...');
       const templates = await fetchBlueprintRepositories();
-      spinnerBlueprints.succeed('Starter templates fetched successfully');
+      activeSpinner.succeed('Starter templates fetched successfully');
 
       if (templates.length === 0) {
-        spinnerBlueprints.failed();
+        activeSpinner.failed();
```
maoberlehner, line 136: nitpick: pre-existing feel free to ignore, we call `activeSpinner.succeed` and then `activeSpinner.failed` might be cleaner to only call `activeSpinner.succeed` if `templates.length >== 0`

maoberlehner (approved):

packages/cli/src/utils/error/error.ts
alexjoverm: the "why" I don't know, but this is IMO a clear anti-pattern. Before I already removed others from the code-base, will remove this as well. Well spotted

packages/cli/src/commands/components/push/graph-operations/resource-processor.ts
alexjoverm: actually it is, just wasn't exported 🤦 great catch!

packages/cli/src/commands/components/pull/index.ts
alexjoverm: I think it's a great idea, but I'd rather tackle it as a separate ticket. As you well spotted, there are some commands that are quite fast/small/non-bulk, and for those a progress bar might not be the best UI component fit. In that ticket, we can evaluate how's the UI/UX of all commands, spot inconsistencies and improve them 😄

PR #716, opened 2026-07-28
fix(js): accept any blok type in storyblokEditable

### LINKED ISSUE #711: [Bug]: storyblokEditable rejects generated component types without explicit _editable
### Package

`@storyblok/js`

### Bug Description

`storyblokEditable` rejects valid Storyblok component types that do not explicitly declare an optional `_editable` property. This affects component interfaces generated from Storyblok schemas, which declare `_uid` and `component` but may not declare `_editable`.

### Steps to Reproduce

Compile the following code with TypeScript strict mode enabled:

```ts
import { storyblokEditable } from "@storyblok/js";

interface FeatureStoryblok {
  headline?: string;
  component: "feature";
  _uid: string;
}

declare const blok: FeatureStoryblok;

storyblokEditable(blok);
```

### Expected Behavior

`storyblokEditable` accepts a valid Storyblok component type. The `_editable` property is optional at runtime, so its absence from the declared component interface should not prevent the call from compiling.

### Actual Behavior

TypeScript rejects the call because the generated component type has no explicitly declared property in common with the current `{ _editable?: string }` parameter type.

### Environment

```shell
@storyblok/js: 6.3.0
typescript: 5.8.3
strict: true
```

### Error Logs

```shell
Argument of type FeatureStoryblok is not assignable to parameter of type { _editable?: string | undefined; }.
  Type FeatureStoryblok has no properties in common with type { _editable?: string | undefined; }.
```

dipankarmaikap:
## Problem

`storyblokEditable` was typed as:

```ts
(blok?: { _editable?: string }) => ...
```

`{ _editable?: string }` is a **weak type** — all properties are optional. TypeScript's weak-type check ([ts(2559)](https://www.typescriptlang.org/tsconfig#strict)) rejects any argument whose declared type shares **zero property names** with the parameter type.

User-defined blok interfaces like:

```ts
interface FeatureBlok {
  headline?: string;
  component: 'feature';
  _uid: string;
}

declare const blok: FeatureBlok;
storyblokEditable(blok); // ❌ ts(2559): Type 'FeatureBlok' has no properties in common with type '{ _editable?: string | undefined; }'
```

have no property named `_editable`, so TypeScript refused the call even though the function handles the absent field gracefully at runtime (returning `{}`).

## Why not `Record<string, unknown>` or `{ [key: string]: unknown }`?

Both require the argument type to carry an index signature. Interfaces without one (the common case for user blok types) produce the same error in a different form: ts(2345) `Index signature for type 'string' is missing in type 'FeatureBlok'`.

## Fix

Widen the parameter to `unknown`:

```ts
(blok?: unknown) => ...
```

The function already defensively narrows the value at runtime (`typeof blok !== 'object'`, `typeof _editable !== 'string'`), so no structural constraint should be imposed on the caller. `unknown` is the honest type: accept anything, prove safety internally.

## Testing

- Existing test preserved and passing
- 3 new tests added: type-level assertion (no `@ts-expect-error`), runtime `{}` when `_editable` absent, runtime data attributes when `_editable` present on a typed blok
- Added `tsconfig.test.json` — the main tsconfig explicitly excludes `*.test.ts`, so type-level assertions in tests were never verified. The new config extends the base but re-includes test files, enabling `tsc --noEmit --project tsconfig.test.json`

Fixes DX-519
Fixes #711

### DIFF
--- a/packages/js/src/editable.ts
+++ b/packages/js/src/editable.ts
@@ -1,11 +1,17 @@
-export default (blok?: { _editable?: string }) => {
-  if (typeof blok !== 'object' || typeof blok._editable !== 'string') {
+export default (blok?: unknown) => {
+  if (typeof blok !== 'object' || blok === null) {
+    return {};
+  }
+
+  const _editable = (blok as Record<string, unknown>)._editable;
+
+  if (typeof _editable !== 'string') {
     return {};
   }
 
   try {
     const options = JSON.parse(
-      blok._editable.replace(/^<!--#storyblok#/, '').replace(/-->$/, ''),
+      _editable.replace(/^<!--#storyblok#/, '').replace(/-->$/, ''),
     );
 
     if (options) {
--- a/packages/js/src/index.test.ts
+++ b/packages/js/src/index.test.ts
@@ -279,6 +279,50 @@ describe('@storyblok/js', () => {
       expect(editableResult['data-blok-uid']).toBeDefined();
       expect(fetchSpy).toHaveBeenCalled();
     });
+
+    it('accepts a blok type that does not declare _editable', () => {
+      // FeatureStoryblok has no _editable and shares no property names with
+      // { _editable?: string }, which would trigger ts(2559) (weak type check).
+      // The parameter type `unknown` accepts any value, resolving this entirely.
+      interface FeatureStoryblok {
+        headline?: string;
+        component: 'feature';
+        _uid: string;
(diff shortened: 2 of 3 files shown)


dipankarmaikap:
Closing this PR since this should be fixed in the CLI instead. The CLI should generate types that include the `_editable` field, aligning with the schema package and the way it handles types.

PR #719, opened 2026-07-28
feat(cli): generate schema-derived types with types generate --future-schema

maoberlehner:
Adds `storyblok types generate --future-schema`, which generates accurate TypeScript types for a UI-managed Storyblok space, and deprecates the legacy generator.

## Why

The legacy generator builds types with `json-schema-to-typescript` from pulled component JSON, and gets them wrong in ways that matter: field optionality does not follow `required`, `bloks` fields are not narrowed by `component_whitelist`, the nestable and root distinction is ignored, and custom field types fall back to a loose parser hook. It also requires a prior `components pull` with matching flags, which is a persistent footgun.

`@storyblok/schema` already models all of this correctly at the type level, but only for people who define their schema in code. This flag gives UI-managed spaces the same types without adopting a code-driven schema workflow.

## Approach

The command fetches the space's components and emits **block definition type literals** plus the public surface of a hand-written `schema.ts`. TypeScript then resolves content shapes in the user's own project through `BlockContent`, so the CLI duplicates none of the field-to-value rules and cannot drift from them.

```ts
import type { Block, Schema } from './.storyblok/types/295018/storyblok-schema';

interface Props { block: Block<'hero'> }

const client = createApiClient({ accessToken }).withTypes<Schema>();
```

`Block<'hero'>` is the user-facing surface, matching the helper dogfooded in the Astro playground in 2852ad1cc. The definition types are plumbing for `withTypes<Schema>()`, `BlockContent`, and the story aliases.

Emitted exports: `Block<TName>`, `AnyBlock`, `Schema`, `Blocks`, `FieldPlugins`, `Story`, `StoryMapi`, and one `<Name>BlockDefinition` per block. `--type-prefix` and `--type-suffix` apply to every one of them.

`adr/0012-schema-derived-type-generation.md` records the decision and, importantly, the rejected alternative: resolving flattened content interfaces with the TypeScript compiler API. That was prototyped and rejected because self-referencing blocks collapse to `any` under every `NodeBuilderFlags` combination, `Prettify` destroys `aliasSymbol` so named types inline as raw structure, the property walk would duplicate the field-to-value rules in JS, and it would add `typescript` as a CLI runtime dependency with a platform-specific native binary. The ADR exists so nobody retries it.

## Deprecation

Running `types generate` without `--future-schema` now prints a deprecation warning naming the replacement and the concrete defects. Nothing is removed and no timeline is committed. The legacy path is otherwise byte-identical, and by default the new output goes to a separate file (`storyblok-schema.d.ts`), so the two generators do not collide unless `--filename` forces them to, which warns.

`--strict`, `--custom-fields-parser`, `--compiler-options`, and `--suffix` are legacy-only and error when combined with `--future-schema`, rather than being silently ignored. A value that came from a config file rather than the command line is reported as ignored instead, so a project that configures `strict` for the legacy generator can still use this flag.

## Custom field types

Custom fields need their `field_type` bound to a validator with `defineFieldPlugin`. The CLI looks for `schema/schema.ts` under the CLI base path (`--path`, default `.storyblok`) by convention, or an explicit `--field-plugins <path>`, and accepts either a `schema` export (a `defineSchema` result) or a bare `fieldPlugins` record. Unregistered `field_type`s fall back to an untyped value and are reported as a warning.

## Documentation

The user-facing documentation lives in a docs PR, [storyblok/storyblok-docs-platform#597](https://github.com/storyblok/storyblok-docs-platform/pull/597), not in this repo. The package README only gains the deprecation callout and the new flags in its existing options table.

## Notes for the reviewer

1. **One unrelated lockfile line.** Adding `@storyblok/schema` as a `packages/cli` devDependency required `pnpm install`, which also dropped `async-sema 3.1.1` from the `@storyblok/management-api-client` snapshot. That is stale drift from ff3642d72: mapi-client no longer declares that dependency, only `packages/cli` does. It is a correction, not an accidental removal, and splitting it out would just mean the next `pnpm install` regenerates it.

2. **Land #718 first, and resolve one overlap by hand.** The branches overlap in `packages/cli/src/commands/schema/init/generate-code.ts` and `pnpm-lock.yaml`. Both fix the same `schema init` defect — an inert whitelist beside `restrict_components: false` was mapped to `allow`, so the next `schema push` switched a disabled restriction back on — but #718 fixes it in the private `toDslField` inside `generate-code.ts` while this branch fixes it in `to-dsl-field.ts`, the shared module it extracted that function into. Taking #718's hunk would reintroduce the bug; dropping it is correct because the extracted module already carries the gate. `to-dsl-field.test.ts` covers both whitelist forms, so a wrong resolution fails the tests.

3. **Defects found and fixed during review.**

   `allow` emission, all three of which produce types that reject content the editor accepts:
   - It was emitted for a field whose `restrict_components` is `false`. The group-whitelist form of that state is never normalized by the backend, so it is live rather than theoretical.
   - It was emitted for any field type, including `multilink`, where the same wire key holds content type names rather than block names. The gate is `bloks` plus `richtext` — a richtext whitelist genuinely does restrict blocks.
   - Entries naming a deleted component were kept. `ApplyAllow` is an `Extract`, so a list of only stale names resolved the field to `never[]`.

   Messages that told the user something untrue:
   - The `--filename` collision warning fired for every `--filename`, asserting a clash with a file nothing writes. It now compares against the legacy default.
   - The unmapped-`field_type` warning told users to place a module where one already sat — the case `schema init` produces, since it writes a `schema` export with no `fieldPlugins` key. Missing and unusable modules are now distinguished, and the warning names the export to rename.
   - All four legacy-only flags shared one rationale about field optionality and the JSON-schema compiler, none of which applies to `--suffix`. Each now carries its own reason.

   Also fixed:
   - `tab` and `section` fields surfaced as `key?: null` on content types. They carry no value, so those keys are dropped — in the codegen template rather than this command's serializer, so hand-written `defineSchema` blocks benefit too.
   - Emitted blocks are sorted by name, so regeneration is byte-stable: MAPI does not promise an order and this file is committed.
   - The field-plugins convention path honours `--path`; the `--field-plugins` error names a near-miss export, since the real failure mode is a misnamed export rather than a missing file; and the shared jiti bootstrap and the `.storyblok/schema` literal are each defined once.

Fixes DX-525


packages/cli/src/commands/types/generate/index.ts
```diff
@@ -23,14 +27,66 @@ const generateCmd = typesCommand
...
+  .option('--future-schema', 'Generate types from the space schema (accurate optionality, block narrowing, and custom field types)')
+  .option('--field-plugins <path>', 'Path to a module exporting your defineFieldPlugin declarations (default: .storyblok/schema/schema.ts)');
 
 generateCmd
   .action(async (options: GenerateTypesOptions, command: Command) => {
-    konsola.title(`${commands.TYPES}`, colorPalette.TYPES, 'Generating types...');
-
     const { space, path, verbose, suffix, filename, separateFiles } = command.optsWithGlobals();
 
+    if (options.futureSchema) {
```
maoberlehner: move this branch into a separate file

packages/cli/src/commands/types/generate/index.ts
```diff
@@ -23,14 +27,66 @@ const generateCmd = typesCommand
   .option('--suffix <suffix>', 'Components suffix')
   .option('--custom-fields-parser <path>', 'Path to the parser file for Custom Field Types')
   .option('--compiler-options <options>', 'path to the compiler options from json-schema-to-typescript')
-  .option('-s, --space <space>', 'space ID');
+  .option('-s, --space <space>', 'space ID')
+  .option('--future-schema', 'Generate types from the space schema (accurate optionality, block narrowing, and custom field types)')
```
maoberlehner: remove: `(accurate optionality, block narrowing, and custom field types)`

packages/cli/src/commands/types/generate/README.md
```diff
@@ -1,17 +1,65 @@
...
+
+### Supported options
+
+`--future-schema` honours `--space`, `--path`, `--filename`, `--separate-files`, `--type-prefix`, `--type-suffix`, and `--field-plugins`. Prefix and suffix apply to every exported type name, not just block names, so `--type-prefix Sb` turns `Block` into `SbBlock`, `Schema` into `SbSchema`, and so on, along with every reference to those names inside the file.
+
+`--strict`, `--custom-fields-parser`, and `--compiler-options` are legacy-only and error when combined with `--future-schema`: there is no `json-schema-to-typescript` compiler involved, optionality comes from the schema's own `required` flags, and custom fields are typed with `defineFieldPlugin` instead of a parser file.
+
+## Legacy generator (deprecated)
+
+### Basic usage
```
maoberlehner: Create a docs PR instead and remove from here

PR #721, opened 2026-07-29
fix(schema): type field default_value per field type

maoberlehner:
`default_value` was declared once on the shared `value-field` spec fragment as `string`, so every field type inherited a `string`-only default. Reported by @diegoscholz:

```ts
defineField('default_open', {
  type: 'boolean',
  default_value: false, // Type 'boolean' is not assignable to type 'string'. ts(2322)
})
```

This also breaks code that `storyblok schema init` itself scaffolds — the serializer emits booleans unquoted, so bootstrapping a space with a boolean field that has a default produces code that doesn't compile.

## What the real type is

Verified against MAPI and the app:

- **Backend** (`storyrails`): the component `schema` is a free-form JSON blob. No validation, no coercion — `false`, `true` and `"true"` all round-trip verbatim (confirmed against a live space).
- **App** (`storyfront`): the boolean schema editor is a toggle emitting a real boolean, and default application does a strict identity check — `if (schema.default_value === true)`. So `default_value: "true"` silently behaves as `false`; a string is actively wrong for booleans.
- `bloks` / `options` accept either the encoded string (JSON for bloks, space-separated for options) or the array itself.
- Everything else is a string, including `number` (the editor writes a text field; `FieldTypeNumber` parses it).

## Change

Moves `default_value` off `value-field.yaml` onto each field type:

| Field type | `default_value` |
| --- | --- |
| `boolean` | `boolean` |
| `bloks` | `string \| Array<object>` |
| `options` | `string \| Array<string>` |
| all other value fields | `string` |

An override left in place on the field type wouldn't work: the fragments merge via `allOf`, so `string` + `boolean` intersects to `never`. The removal is documented in `value-field.yaml` so it doesn't get re-added.

Regenerated the four consumers (`@storyblok/schema`, `@storyblok/api-client`, `@storyblok/management-api-client`, `@storyblok/live-preview`).

Fixes DX-527

### DIFF
--- a/packages/capi-client/src/generated/overlay/_internal.gen.ts
+++ b/packages/capi-client/src/generated/overlay/_internal.gen.ts
@@ -25,6 +25,10 @@ export type TextFieldRoot = BaseFieldRoot & ValueFieldRoot & {
      */
     type: 'text';
     /**
+     * Default value for the field
+     */
+    default_value?: string;
+    /**
      * Maximum length of the input string
      */
     max_length?: number;
@@ -51,6 +55,10 @@ export type TextareaFieldRoot = BaseFieldRoot & ValueFieldRoot & {
      * Field type discriminant
      */
     type: 'textarea';
+    /**
+     * Default value for the field
+     */
+    default_value?: string;
     /**
      * Maximum length of the input string
      */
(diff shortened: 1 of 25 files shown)


dipankarmaikap (approved):
Looks good, @maoberlehner! Just need to fix the Astro schema playground, as it's still using the old `'false'` string instead of the boolean `false`.

PR #722, opened 2026-07-29
feat(schema,cli): redact plugin option secrets in schema commands

maximilian-schwarz:
## Problem

Custom field plugins like the Shopware and Shopify integrations store credentials inside the field's `options: [{ name, value }]` array (e.g. `clientSecret`).

- `schema init` wrote those values verbatim into the generated, versioned files → secrets end up in Git.
- Deleting them by hand doesn't help: `schema push` then diffs the emptied local value against the remote and **clears the secret on the space**.

## Solution

A first-class `secret()` placeholder in `@storyblok/schema`, recognized by the schema command. Secrets are identified purely by the option `name` (the developer-chosen `field_type` is never used), covering the `{ name, value }` options form only.

`secret()` returns a sentinel **string** (`__storyblokSecret__`, or `__storyblokSecret__:ENV`), so it slots into a plugin option's string `value` and the generated files type-check.

- **`secret()`** — preserve-remote: push keeps whatever the space already has, and it is excluded from the diff entirely.
- **`secret('ENV_VAR')`** — read from `process.env` on push, and participates in the diff: when the environment value differs from the value on the space, push reports an update and rotates the secret. Falls back to preserve-remote when the variable is unset.

### Wiring

- **init** — redacts secret option values to `secret()`. Default names + `--secret-names <names>`; opt out via `--no-redact-secrets`.
- **push** — excludes preserve-remote secrets from the diff; env-managed secrets participate via a non-revealing per-run fingerprint (the diff never prints the value). Secrets are hydrated just before the API call (env → remote → omit + warn on create). A placeholder never reaches the Management API.
- **changeset + local component JSON** — real values are masked to placeholders on disk; **rollback** hydrates them from the live space the same way push does.

Name-based array alignment keeps hydration and rotation correct even if option order differs between the local file and the space.

## Tests

Unit tests for the redact / mark / mask / hydrate helpers, integration tests for init codegen, push diff (including env rotation), and `executePush` (preserve + env), plus a jiti round-trip. All green.

## Notes

- On **create** (fresh space) a preserve-remote `secret()` has no source, so the value is left empty with a warning — use `secret('ENV')` to provision it.
- Optional follow-up: surface `schema.secrets` in the config file (`defineConfig`) so the name list needn't be passed per-invocation.

### DIFF
--- a/packages/cli/src/README.md
+++ b/packages/cli/src/README.md
@@ -34,6 +34,9 @@ npm install storyblok@<version>
 | [`migrations generate`](./commands/migrations/generate/README.md) | ✅ | Replaces previous generate-migrations |
 | [`migrations run`](./commands/migrations/run/README.md) | ✅ | Replaces previous run-migrations |
 | [`migrations rollback`](./commands/migrations/rollback/README.md) | ✅ | Replaces previous rollback-migrations |
+| [`schema init`](./commands/schema/README.md) | ✅ | Bootstrap local code-driven schema from a space. Redacts plugin option secrets to `secret()` |
+| [`schema push`](./commands/schema/README.md#secret-handling) | ✅ | Apply local schema changes to a space. Restores redacted secrets from the space or `process.env` |
+| [`schema rollback`](./commands/schema/README.md#secret-handling) | ✅ | Revert a previous `schema push` |
 | [`types generate`](./commands/types/generate/README.md) | ✅ | Replaces previous generate-typescript-typedefs |
 | [`signup`](./commands/signup/README.md) | ✅ | Opens the Storyblok signup page in your browser |
 | `sync` | ⚠️ | Pending new API endpoint implementation for improved performance and reliability (Check below for more details) |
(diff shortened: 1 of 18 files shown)


maximilian-schwarz:
@cursor review

maoberlehner:
Thank you for your contribution! We're currently investigating how we want to handle this.

maximilian-schwarz:
Hi @maoberlehner just wanted to check in to see if there are any updates on this yet. Let me know if you need any additional info from my side. Thanks!

maoberlehner:
> Hi @maoberlehner just wanted to check in to see if there are any updates on this yet. Let me know if you need any additional info from my side. Thanks!

@maximilian-schwarz I don't recommend using Shopware API v2 and recommend upgrading to v3 for a secure way to handle secrets. Similarly, Shopify V2 also does not require storing secrets in Storyblok fields anymore, and I highly encourage you to upgrade.

Storyblok has no secret store, and storing secrets in field types is not secure. Merging this PR would not change that but would promote usage of secrets in Storyblok field types, which I strongly advise against.

If your secrets are non-sensitive pseudo-secrets (e.g., an API key that can only be used to retrieve public data), you're safe putting them in plain text in your Git repo. If your secrets are real secrets, you shouldn't send them to Storyblok at all.

PR #726, opened 2026-07-31
fix(schema): model multilink as a discriminated union

maoberlehner:
- model multilink values as a shared base plus story, URL, email, and asset variants
- make target and story anchor optional but non-null
- preserve Storyfront string-valued custom attributes for story and URL links
- regenerate schema, Content API, Management API, and live-preview overlay types
- add type-narrowing and runtime-validator regression coverage

## Scope

- no CLI multilink type changes
- no API-client runtime link/story normalization

Fixes DX-534

### DIFF
--- a/packages/capi-client/src/generated/overlay/_internal.gen.ts
+++ b/packages/capi-client/src/generated/overlay/_internal.gen.ts
@@ -701,38 +701,9 @@ export type BlockContentInputRoot = {
 };
 
 /**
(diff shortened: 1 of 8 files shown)


dipankarmaikap (approved):
Approving. Should we also add this to the rich text OpenAPI spec?

maoberlehner:
> Approving. Should we also add this to the rich text OpenAPI spec?

@dipankarmaikap yes, if it makes sense: go for it!

PR #728, opened 2026-07-31
feat(richtext): use OpenAPI spec as source of truth for richtext types

dipankarmaikap:
## Summary

Uses the Storyblok OpenAPI overlay spec as the single source of truth for all richtext node/mark types across the richtext package and every framework integration. Removes the hand-rolled Tiptap-derived type generator and replaces it with generated types that stay in sync with the API spec.

### Naming pattern (applied uniformly across all packages)

Every framework-specific richtext type follows the same rename:

| Before | After |
|--------|-------|
| `SbRichText*` | `StoryblokRichText*` |
| `SbAstroRichText*` | `StoryblokAstroRichText*` |
| `SbVueRichText*` | `StoryblokVueRichText*` |
| `SbReactRichText*` | `StoryblokReactRichText*` |
| `SbSvelteRichText*` | `StoryblokSvelteRichText*` |
| `SbAngularRichText*` | `StoryblokAngularRichText*` |

All old `Sb*` names are kept as `@deprecated` aliases — no breaking changes.

The renamed variants cover the same four shapes in each package: `Props`, `RenderContext`, `ComponentMap`, and where applicable `Component` / `DocumentProps`.

### `@storyblok/richtext`
- **OpenAPI codegen pipeline** — adds `scripts/generate.ts` using `@storyblok/openapi-codegen`; commits `src/generated/overlay/types.gen.ts` as reviewed output
- **Removes hand-rolled type generator** — deletes `src/static/generate/richtext-type.ts`
- **New utilities** — `hasContent(node)` type guard; `StoryblokRichTextNodeWithKey` / `StoryblokRichTextMarkWithKey` for keyed rendering; `StoryblokRichTextElementByType` discriminated lookup derived from the spec
- **Extensions** — `richtext-attrs.ts` derives `ExtensionAttrMap` directly from the OpenAPI discriminated union

### Framework packages (`@storyblok/astro`, `@storyblok/vue`, `@storyblok/react`, `@storyblok/svelte`, `@storyblok/angular`)
- Types renamed following the pattern above; deprecated aliases preserved
- All node params switched to `StoryblokRichTextNodeWithKey` for safe keyed iteration
- `attrs` access guarded with `'attrs' in node` checks throughout the rendering tree
- `StoryblokRichTextElementByType` (or the `Omit`-based equivalent where the compiler allows) replaces the removed `SbRichTextElementByType` lookup
- `hasContent` and all richtext helpers re-exported from each package's public API

## Migration guide (for consumers)

All `Sb*` types still work — no immediate action needed. To adopt the new names:

```ts
// Before
import type { SbRichTextNode, SbRichTextDoc, RenderSpec } from '@storyblok/richtext'

// After
import type { StoryblokRichTextNode, StoryblokRichTextDoc, StoryblokRichTextRenderSpec } from '@storyblok/richtext'
```

Same pattern in framework packages:

```ts
// Before (Vue example)
import type { SbVueRichTextProps, SbVueRichTextComponentMap } from '@storyblok/vue'

// After
import type { StoryblokVueRichTextProps, StoryblokVueRichTextComponentMap } from '@storyblok/vue'
```

Fixes DX-493


packages/react/src/core/richtext.ts
```diff
@@ -1,13 +1,65 @@
...
 
 export type {
+  StoryblokRichTextElement,
+  StoryblokRichTextImageOptions,
+  StoryblokRichTextInput,
+  StoryblokRichTextMark,
+  StoryblokRichTextMarkWithKey,
+  StoryblokRichTextNode,
+  StoryblokRichTextNodeWithKey,
+  StoryblokRichTextProps,
```
maoberlehner: This silently changes what `StoryblokRichTextProps` means for `@storyblok/react` consumers. On main it's the document-level props exported from `./rich-text-renderer` (`{ document, optimizeImage, components, data }`), public via `export * from './core/richtext'` in `src/index.ts`. Here it becomes richtext's generic node props, so existing code like `const p: StoryblokRichTextProps = { document }` fails with "requires 1 type argument".

Can we drop it from this re-export (react already has `StoryblokReactRichTextProps`) and keep the old name as `/** @deprecated */ export type StoryblokRichTextProps = StoryblokReactRichTextDocumentProps`?

packages/astro/src/utils/richtext-helpers.ts
```diff
@@ -1,28 +1,56 @@
...
-export type SbAstroRichTextProps<
-  T extends SbRichTextElement,
-> =
-  SbRichTextElementByType<SbAstroRichTextRenderContext>[T];
+/**
+ * @deprecated Use {@link StoryblokAstroRichTextRenderContext} instead. Will be removed in the next major version.
+ */
+export type SbAstroRichTextRenderContext = StoryblokAstroRichTextRenderContext;
+
+export type StoryblokAstroRichTextProps<T extends StoryblokRichTextElement> =
```
maoberlehner, line 29: `StoryblokRichTextProps` carries a required `children: string`, which only makes sense for the string renderer. Astro passes children through the slot (see `RenderNode.astro`), never as a prop. Anyone typing a custom component with this ends up declaring a required `children` that's always `undefined` at runtime.

I think it should be `Omit<..., 'context' | 'children'>`, same as the React one.

packages/svelte/src/lib/richtext-helpers.ts
```diff
@@ -1,25 +1,45 @@
...
+ * @deprecated Use {@link StoryblokSvelteRichTextRenderContext} instead. Will be removed in the next major version.
+ */
+export type SbSvelteRichTextRenderContext = StoryblokSvelteRichTextRenderContext;
 
-export type SbSvelteRichTextProps<
-  T extends SbRichTextElement,
-> =
-  SbRichTextElementByType<SbSvelteRichTextRenderContext>[T]
-  & {
+export type StoryblokSvelteRichTextProps<T extends StoryblokRichTextElement> =
```
maoberlehner, line 30: Same as astro, but worse here: `Omit<..., 'context'>` keeps the required `children: string`, and intersecting `children?: Snippet` on top collapses it to a required `children: string & Snippet`. `src/tests/richtext/CustomLink.svelte:5` destructures that.

It slips through because `test:types` is plain `tsc --noEmit`, so `.svelte` files aren't type-checked at all. Add `'children'` to the `Omit`. Maybe worth wiring up `svelte-check` while we're in here.

packages/richtext/src/render-richtext.ts
```diff
@@ -89,13 +146,12 @@ function renderNode(node: SbRichTextNode, context?: SbRichTextRenderContext): st
...
 /** Renders an image node with optimization applied. */
 function renderOptimizedImage(
-  node: SbRichTextNode,
-  context: SbRichTextRenderContext,
+  node: Extract<RichTextNode, { type: 'image' }>,
+  context: StoryblokRichTextRenderContext,
 ): string {
-  const attrs = node.attrs as Record<string, unknown> | undefined;
-  const src = attrs?.src as string | undefined;
+  const src = node.attrs.src ?? undefined;
```
maoberlehner: nitpick: This was `attrs?.src`. `attrs` now throws instead of degrading. The React renderer kept the guard (`'attrs' in node ? node.attrs : undefined`), so the two renderers now disagree on how defensive they are.

packages/richtext/src/static/generate/richtext-element-types.ts
```diff
@@ -0,0 +1,166 @@
...
+      hasMarks,
+      marksOptional,
+      extraProps,
+    });
+  }
+
+  return entries;
+}
+
+export function generateElementTypes(inputPath: string): string {
```
maoberlehner, line 121: Do we need the AST walk? Everything this emits looks expressible as a mapped type, and `static/types.ts` already builds the same lookup as `RichTextElementMap`:

```ts
export type StoryblokRichTextElementByType<TContext = unknown> = {
  [K in keyof RichTextElementMap]: RichTextElementMap[K] & { _key?: string; context?: TContext };
};
```

That would drop this file, the committed `.generated.ts`, and the generate-before-build coupling.

packages/richtext/src/static/types.type-test.ts
```diff
@@ -1,32 +1,32 @@
...
  * nested nodes (empty paragraphs, leaf nodes) may omit `content` entirely.
  * The types must mirror that: `content` is required on the document root and
  * optional on every other node.
  *
- * These assertions are checked by `pnpm test:types` (`tsc --noEmit`) and are
+ * These assertions are checked by `pnpm typecheck` (`tsc --noEmit`) and are
  * intentionally never executed at runtime.
  */
-import type { SbRichTextDoc, SbRichTextNode } from './index';
+import type { RichTextDoc, RichTextNode } from '../generated/overlay/types.gen';
```
maoberlehner: This file exists to guard the *public* contract from #16. Pointing it at the generated file means it no longer does. Keep it on `StoryblokRichTextDoc` / `StoryblokRichTextNode` from `./index`.

packages/richtext/src/static/util.ts
```diff
@@ -75,3 +80,9 @@ export function attrsToHtmlString(attrs: Record<string, unknown>): string {
 
   return result;
 }
+
+export function hasContent(
```
maoberlehner: Narrowing loses the key typing when this is called with a `StoryblokRichTextNodeWithKey`. Generic keeps it:

```ts
export function hasContent<T extends RichTextNode>(node: T): node is T & { content: NonNullable<T['content']> }
```

packages/richtext/src/generated/overlay/types.gen.ts
```diff
@@ -0,0 +1,499 @@
...
+    copyright: string | null;
+    /**
+     * Asset metadata
+     */
+    meta_data: {
+      alt?: string | null;
+      title?: string | null;
+      source?: string | null;
+      copyright?: string | null;
+      [key: string]: string | null | string | null | string | null | string | null | string | null | undefined;
```
maoberlehner, line 170: `string | null` repeated five times. `PluginFieldValueRoot`'s index signature has the same problem (`unknown | string | undefined`, which just collapses to `unknown`). Cosmetic, but maybe we can fix in `tools/openapi-codegen`?

packages/richtext/package.json
```diff
@@ -66,6 +66,7 @@
     "build": "tsdown",
     "build:watch": "tsdown --watch",
     "test:types": "tsc --noEmit --skipLibCheck",
+    "typecheck": "tsc -p tsconfig.json && tsc -p tsconfig.test.json",
```
maoberlehner: Why do we need an extra `typecheck`? In the rest of the repo, we use `test:types` exclusively. Here we now use `test:types` and `typecheck` and at least to me, it is unclear which one is for what.

packages/vue/src/rich-text-renderer.ts
```diff
@@ -1,45 +1,66 @@
...
+}
+
+/**
+ * @deprecated Use {@link StoryblokVueRichTextRenderContext} instead. Will be removed in the next major version.
+ */
+export type SbVueRichTextRenderContext = StoryblokVueRichTextRenderContext;
 
-export type SbVueRichTextComponentMap = {
-  [K in SbRichTextElement]?: Component<SbVueRichTextProps[K]>;
+export type StoryblokVueRichTextProps = StoryblokRichTextElementByType<StoryblokVueRichTextRenderContext>;
```
maoberlehner, line 37: Vue stays on `StoryblokRichTextElementByType` while react/astro/svelte moved to `Omit<StoryblokRichTextProps<T>, ...>`. Three shapes for one concept. Can we settle on one?

maoberlehner (changes requested):
I'm confused about `types.generated.ts` in `.gitignore`. Shouldn't `src/static`, be git ignored as a whole? As I understand it those types are generated by TipTap and other than with OpenAPI specs, that are private, I don't see a reason why we need to track them in git?

maoberlehner (commented):
Code looks good to me, Claude QA run found the following regresssion:

| input                                            | main                                   | PR                                              |
| ------------------------------------------------ | -------------------------------------- | ----------------------------------------------- |
| `attrs.emoji: ''` + `content: [text 'FALLBACK']` | `<span data-name="x">FALLBACK</span>`  | `<span data-name="x"></span>`                   |

maoberlehner (approved):
Please take a look at the case found by Claude. Fix it or decide it's not worth to fix, otherwise LGTM.

dipankarmaikap:
@maoberlehner Claude's finding is never going to happen.
There is no scenario where an emoji appears on its own. It will always be inside a paragraph, and the emoji itself never has any content—the paragraph does.

packages/vue/src/rich-text-renderer.ts
dipankarmaikap: With the Vue TypeScript compiler, we have to provide a static list at the moment. Unfortunately, there isn't a workaround for this right now.

packages/richtext/package.json
dipankarmaikap: Fixed

packages/richtext/src/generated/overlay/types.gen.ts
dipankarmaikap: Will create a seperate pr

packages/richtext/src/static/util.ts
dipankarmaikap: Fixed

packages/richtext/src/static/types.type-test.ts
dipankarmaikap: Fixed

packages/richtext/src/static/generate/richtext-element-types.ts
dipankarmaikap: This is only needed for Vue.

packages/richtext/src/render-richtext.ts
dipankarmaikap: Fixed

packages/svelte/src/lib/richtext-helpers.ts
dipankarmaikap: Fixed

packages/astro/src/utils/richtext-helpers.ts
dipankarmaikap: Fixed

PR #730, opened 2026-08-03
feat(experiments): keep serverless runtimes alive until event delivery settles

maoberlehner:
The `createExperiments` factory fires adapter events without awaiting them. On serverless platforms (Vercel, Netlify, Cloudflare) the process is frozen as soon as the response goes out, so in-flight deliveries are silently dropped: they never reject, so `onError` never fires. Conversions are hit harder than exposures because `track` typically runs right before the response, which biases measured conversion rates.

This PR hardens delivery without changing existing behavior:

- New `waitUntil` factory option: every pending async delivery is handed to the host platform's `waitUntil` (Vercel `@vercel/functions`, Next.js `after`, Netlify `context.waitUntil`, Cloudflare `ctx.waitUntil`, Nitro `event.waitUntil` all share the `(promise) => void` shape). The handed promise never rejects; failures still route to `onError`.
- New `flush()` method: resolves once every pending delivery has settled, for platforms without a `waitUntil` hook or when delivery must complete before the response. Non-rejecting, so a downed analytics sink still cannot break a request.
- Adapters are still invoked synchronously so the fetch is initiated before the handler returns, and `resolveExperiment` stays sync on the render path. Sync adapters no longer allocate a throwaway microtask.
- Fixed the `Adapter` doc comment that promised awaitable delivery the factory previously made impossible.

Docs PR with Vercel/Cloudflare guidance: storyblok/storyblok-docs-platform#603

Fixes DX-539

### DIFF
--- a/packages/experiments/src/create-experiments.test.ts
+++ b/packages/experiments/src/create-experiments.test.ts
@@ -88,4 +88,95 @@ describe('createExperiments', () => {
 
     expect(() => exp.resolveExperiment({ slug: 'home', visitorId: 'visitor-1' })).not.toThrow();
   });
+
+  it('hands each pending async delivery to waitUntil', async () => {
+    const waitUntil = vi.fn();
+    let delivered = false;
+    let resolveDelivery: () => void;
+    const delivery = new Promise<void>((resolve) => {
+      resolveDelivery = resolve;
+    });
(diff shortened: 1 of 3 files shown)


dipankarmaikap (approved):
LGTM 🚀

PR #732, opened 2026-08-04
feat(cli): add support for .vtt files when using assets push

### LINKED ISSUE #731: [Feature]: Support .vtt subtitles files when using CLI
### Package

storyblok (CLI)

### Problem

We run a website with ~40 languages and multiple videos that we would like to add localized subtitles to, in order to improve accessibility. We would like to upload our `.vtt` subtitle files as Storyblok Assets. While we can do this using the Storyblok Visual Editor or by explicitly using the Storyblok CLI with `storyblok assets push /path/to/specific/file.vtt`, using the more generic `storyblok assets push` ignores the `.vtt` files.

We assume this is because `.vtt` is not present in the [SUPPORTED_ASSET_EXTENSIONS array](https://github.com/storyblok/monoblok/blob/66547b6af129dd31e53e3d5140d70f421a01bd27/packages/cli/src/constants.ts), but as a [widely available and standard way](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API/Web_Video_Text_Tracks_Format) of adding subtitles to videos, we think it should be supported by the CLI.

### Proposed Solution

Adding `.vtt` to the list of supported asset extensions would allow the CLI's `assets push` command to simply include this type of file when pushing assets.

I will submit a PR with the proposed solution

### Alternatives Considered

I've successfully pushed individual .vtt files by using `storyblok assets push /path/to/specific/file.vtt` but this is not a scalable solution for us as we work with hundreds of these files and need to update them occasionally as new translations arrive or improve

### Additional Context

_No response_

knandersen:
The Storyblok CLI `storyblok assets push` command only picks up assets with extensions in the [`SUPPORTED_ASSET_EXTENSIONS`](https://github.com/storyblok/monoblok/blob/66547b6af129dd31e53e3d5140d70f421a01bd27/packages/cli/src/constants.ts) array. Among others, this causes the command to ignore `.vtt` subtitles files.

This PR expands the `SUPPORTED_ASSET_EXTENSIONS` with the `.vtt` extension

Fixes #731

### DIFF
--- a/packages/cli/src/constants.ts
+++ b/packages/cli/src/constants.ts
@@ -131,11 +131,12 @@ export const SUPPORTED_ASSET_EXTENSIONS = new Set([
   '.wma',
   '.m4a',
   '.opus',
-  // Documents: application/msword, text/plain, application/pdf, application/vnd.openxmlformats-officedocument.wordprocessingml.document
+  // Documents: application/msword, text/plain, application/pdf, application/vnd.openxmlformats-officedocument.wordprocessingml.document, text/vtt
   '.pdf',
   '.doc',
   '.docx',
   '.txt',
+  '.vtt',
 ]);
 
 export const directories = {


dipankarmaikap (approved):
Thanks for the PR and for the contribution! 🙌

I'm merging this now since the API already validates whether the asset type is supported for a given space. We'll follow up separately on improving the error messages for failed asset uploads so users get more actionable feedback.

PR #733, opened 2026-08-04
fix(cli): surface API error messages on asset push failure

dipankarmaikap:
## What

Asset push failures now show the server's actual error message in the terminal. The fix lives in `APIError` so all CLI commands benefit automatically.

## Why

When a push fails (e.g. a starter plan rejecting an unsupported file type), users saw only a failure count with no explanation. The API returns a clear message that was being discarded.

## Changes

**`api-error.ts`** — `APIError` now extracts a server-provided message from the response body and sets it as `this.message`, replacing the generic fallback for all ~90 `handleAPIError` call sites. Two guards keep the UX from degrading:
- **401 regression fix** — MAPI echoes `{"error":"Unauthorized"}` on 401. HTTP/2 sends empty `statusText`, so the previous guard let it override the friendlier `API_ERRORS.unauthorized` constant. `HTTP_REASON_PHRASES` provides the canonical phrase as a fallback so it is correctly suppressed.
- **422 object-shaped errors** — `{"error":{"base":["..."]}}` has a non-string error field. `firstFieldValue()` walks the raw response data and sets the first string value as the message (without leaking internal key names like `base:`). `messageStack` keeps the full `key: value` entries for verbose display.

**`pipelines.ts`** — replace per-asset `ui.warn` calls with a `FailureReasonGroup` that collects failures and emits grouped lines after the pipeline finishes. Prevents N identical warn lines interleaved with active progress bars. Uses `short_filename` so CDN URLs don't appear as labels.

**`streams.ts`** — treat `ENOENT` from the outer directory scan as empty instead of forwarding to `onAssetError`. A missing `shared/<id>/` directory is normal for any space that has library access but has never pulled shared assets.

**`transfer/index.ts`** — migrated to the same `FailureReasonGroup` utility, removing the inline grouping logic.

**`error.ts`** — removed `getApiResponseMessage`; callers use `toError(error).message` directly now that `APIError.message` already carries the server string.

## Before / After

**Before:**
```
⚠  Push results: 1 processed, 1 assets failed
  Assets: 0/1 succeeded, 1 failed.
```

**After (unsupported file type):**
```
⚠  Failed to push "demo.vtt": Your space must be verified to allow this asset type. Please reach out to our Support team.
⚠  Push results: 1 processed, 1 assets failed
  Assets: 0/1 succeeded, 1 failed.
```

**After (multiple failures, same cause):**
```
⚠  Failed to push 5 assets — Your space must be verified to allow this asset type. (e.g. "a.vtt", "b.vtt", "c.vtt")
```

**After (invalid folder ID):**
```
⚠  Failed to push "image.png": This asset folder is not valid
```

Fixes DX-541

### DIFF
--- a/packages/cli/src/commands/assets/pipelines.ts
+++ b/packages/cli/src/commands/assets/pipelines.ts
@@ -7,7 +7,7 @@ import { fetchStoriesStream, fetchStoryStream, mapReferencesStream, writeStorySt
 import { validateStoryAgainstSchemas } from '../stories/validate-story';
 import type { Logger } from '../../lib/logger/logger';
 import type { Report } from '../../lib/reporter/reporter';
-import { logOnlyError } from '../../utils/error/error';
+import { getApiResponseMessage, logOnlyError } from '../../utils/error/error';
 import type {
   AppendAssetFolderManifestTransport,
   AppendAssetManifestTransport,
@@ -116,15 +116,6 @@ export const upsertAssetsPipeline = async ({
   const assetProgress = ui.createProgressBar({ title: 'Assets...'.padEnd(PROGRESS_BAR_PADDING) });
   const summary = { total: 0, succeeded: 0, failed: 0 };
 
-  // Extract the human-readable message from the API response body when present
-  // (e.g. { error: "..." } or { message: "..." }), falling back to error.message.
-  const getAssetErrorMessage = (error: Error): string => {
-    const data = (error as any)?.response?.data;
-    return (typeof data?.error === 'string' ? data.error : undefined)
-      ?? (typeof data?.message === 'string' ? data.message : undefined)
-      ?? error.message;
-  };
-
   const steps = [];
   // Use the asset provided via the CLI.
   if (assetBinaryPath && assetData) {
@@ -138,7 +129,7 @@ export const upsertAssetsPipeline = async ({
         summary.failed += 1;
         assetProgress.increment();
         logOnlyError(error);
-        ui.warn(`Failed to read "${assetData?.filename ?? assetBinaryPath}": ${getAssetErrorMessage(error)}`);
+        ui.warn(`Failed to read "${assetData?.filename ?? assetBinaryPath}": ${getApiResponseMessage(error)}`);
       },
     }));
   }
@@ -154,7 +145,7 @@ export const upsertAssetsPipeline = async ({
         summary.failed += 1;
         assetProgress.increment();
         logOnlyError(error);
-        ui.warn(`Failed to read asset: ${getAssetErrorMessage(error)}`);
+        ui.warn(`Failed to read asset: ${getApiResponseMessage(error)}`);
       },
     }));
   }
@@ -182,7 +173,7 @@ export const upsertAssetsPipeline = async ({
     onAssetError: (error, asset) => {
       summary.failed += 1;
       logOnlyError(error, { assetId: asset.id });
-      ui.warn(`Failed to push "${asset.filename ?? asset.short_filename ?? asset.id}": ${getAssetErrorMessage(error)}`);
+      ui.warn(`Failed to push "${asset.filename ?? asset.short_filename ?? asset.id}": ${getApiResponseMessage(error)}`);
     },
   }));
   await pipeline(steps);
--- a/packages/cli/src/commands/assets/pipelines.ts
+++ b/packages/cli/src/commands/assets/pipelines.ts
@@ -7,7 +7,8 @@ import { fetchStoriesStream, fetchStoryStream, mapReferencesStream, writeStorySt
 import { validateStoryAgainstSchemas } from '../stories/validate-story';
 import type { Logger } from '../../lib/logger/logger';
 import type { Report } from '../../lib/reporter/reporter';
-import { getApiResponseMessage, logOnlyError } from '../../utils/error/error';
+import { logOnlyError, toError } from '../../utils/error/error';
+import { FailureReasonGroup } from '../../utils/failure-reason-group';
 import type {
   AppendAssetFolderManifestTransport,
   AppendAssetManifestTransport,
@@ -116,6 +117,11 @@ export const upsertAssetsPipeline = async ({
   const assetProgress = ui.createProgressBar({ title: 'Assets...'.padEnd(PROGRESS_BAR_PADDING) });
   const summary = { total: 0, succeeded: 0, failed: 0 };
 
+  // Collect failures during the pipeline and group by reason afterwards.
+  // This mirrors the transfer pipeline convention (transfer/index.ts) and
+  // prevents 60+ identical warn lines interleaved with active progress bars.
+  const failures = new FailureReasonGroup();
+
   const steps = [];
   // Use the asset provided via the CLI.
   if (assetBinaryPath && assetData) {
@@ -129,7 +135,7 @@ export const upsertAssetsPipeline = async ({
         summary.failed += 1;
         assetProgress.increment();
         logOnlyError(error);
-        ui.warn(`Failed to read "${assetData?.filename ?? assetBinaryPath}": ${getApiResponseMessage(error)}`);
+        failures.record(toError(error).message, assetData?.short_filename ?? assetData?.filename ?? assetBinaryPath);
       },
     }));
   }
@@ -141,11 +147,11 @@ export const upsertAssetsPipeline = async ({
         summary.total = total;
         assetProgress.setTotal(total);
       },
-      onAssetError: (error) => {
+      onAssetError: (error, filePath) => {
         summary.failed += 1;
         assetProgress.increment();
         logOnlyError(error);
-        ui.warn(`Failed to read asset: ${getApiResponseMessage(error)}`);
+        failures.record(toError(error).message, filePath);
       },
     }));
   }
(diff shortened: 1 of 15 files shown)


packages/cli/src/commands/assets/pipelines.ts
```diff
@@ -116,6 +116,15 @@ export const upsertAssetsPipeline = async ({
   const assetProgress = ui.createProgressBar({ title: 'Assets...'.padEnd(PROGRESS_BAR_PADDING) });
   const summary = { total: 0, succeeded: 0, failed: 0 };
 
+  // Extract the human-readable message from the API response body when present
+  // (e.g. { error: "..." } or { message: "..." }), falling back to error.message.
+  const getAssetErrorMessage = (error: Error): string => {
+    const data = (error as any)?.response?.data;
```
alexjoverm: is `any` strictly necessary here?

packages/cli/src/commands/assets/pipelines.ts
```diff
@@ -116,6 +116,15 @@ export const upsertAssetsPipeline = async ({
...
+    return (typeof data?.error === 'string' ? data.error : undefined)
+      ?? (typeof data?.message === 'string' ? data.message : undefined)
+      ?? error.message;
```
alexjoverm: I believe this, separated in different lines/const/conditions, could make it more human-readable

packages/cli/src/commands/assets/pipelines.ts
```diff
@@ -116,6 +116,15 @@ export const upsertAssetsPipeline = async ({
   const assetProgress = ui.createProgressBar({ title: 'Assets...'.padEnd(PROGRESS_BAR_PADDING) });
   const summary = { total: 0, succeeded: 0, failed: 0 };
 
+  // Extract the human-readable message from the API response body when present
+  // (e.g. { error: "..." } or { message: "..." }), falling back to error.message.
+  const getAssetErrorMessage = (error: Error): string => {
```
alexjoverm: @dipankarmaikap this is named as AssetError, but actually isn't this globally applicable to any error message. Isn't this something other error handling places in the CLI can benefit from?

No need to refactor the existing ones, but if we create a util to format error messages, we better think on how to apply it as a core logic rather than tackling a single scenario.

packages/cli/src/commands/assets/pipelines.ts
```diff
@@ -116,6 +116,15 @@ export const upsertAssetsPipeline = async ({
...
+    const data = (error as any)?.response?.data;
+    return (typeof data?.error === 'string' ? data.error : undefined)
+      ?? (typeof data?.message === 'string' ? data.message : undefined)
+      ?? error.message;
+  };
```
maoberlehner: I wonder if we really should only fix this on the asset level. How much effort would it be to have a generic fix that improves error reporting for all commands?

maoberlehner (changes requested):

packages/cli/src/commands/assets/pipelines.ts
dipankarmaikap: Makes sence, On it.

alexjoverm (changes requested):
Markus was faster, but mostly same feedback 😄 

We already have a centralize `APIError` in utils, but seems the constructor is not extracting top-level error/message string fields from response.

packages/cli/src/commands/assets/pipelines.ts
```diff
@@ -129,6 +129,7 @@ export const upsertAssetsPipeline = async ({
         summary.failed += 1;
         assetProgress.increment();
         logOnlyError(error);
+        ui.warn(`Failed to read "${assetData?.filename ?? assetBinaryPath}": ${getApiResponseMessage(error)}`);
```
maoberlehner: Writing to the output while a MultiBar is running makes the MultiBar output ugly. If I remember correctly, in other places we gather warnings and output them all at once when the MultiBar is finished.

packages/cli/src/utils/error/error.ts
```diff
@@ -121,3 +121,16 @@ export function handleError(error: Error | FetchError, verbose = false, context?
...
+/**
+ * Extracts a human-readable message from an error for display in per-item
+ * stream failure warnings (e.g. `ui.warn`).
+ *
+ * For `APIError` instances, `error.message` already reflects the server-provided
+ * string (set by the `APIError` constructor from `response.data.error` /
+ * `response.data.message`). For all other errors it falls back to `error.message`
+ * or `String(error)`.
+ */
+export function getApiResponseMessage(error: unknown): string {
```
maoberlehner: I think the name does not really fit the purpose (nothing about API responses in the body), and it is duplicating work already done by `toError`: you can use `toError(error).message` instead.

packages/cli/src/commands/assets/pipelines.ts
maoberlehner: Additionally:  `readLocalAssetsStream`'s outer catch fires because the local `assets/shared/<libraryId>/` directory doesn't exist. This is normal for any space with shared-library access that has never pulled shared assets.

Although it's not the fault of this PR and I think it should be fixed in `readLocalAssetsStream`, this addition surfaces it and shows a warning for something that is completely normal.

maoberlehner: BEFORE (2/2 assets uploaded):
 Assets  Pushing assets...

 Assets...                [■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■] 100% | 0s | 2/2 processed
ℹ Push results: 0 processed, 1 assets failed

AFTER (same 2/2 assets uploaded):
 Assets  Pushing assets...

⚠️  Failed to read asset: ENOENT: no such file or directory, scandir '/Users/maoberlehner/Workspace/monoblok/.claude/tmp/qa-733/clean-after/assets/shared/185217663908539'
 Assets...                [■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■] 100% | 0s | 2/2 processed
ℹ Push results: 0 processed, 1 assets failed

maoberlehner (changes requested):
Claude QA findings:

## The fix does not fire for real asset-push 422s

The actual asset-create validation shape from MAPI is an **object**, not a string, so `typeof data.error === 'string'` never matches:

```
POST /v1/spaces/290020686946333/assets  {"asset_folder_id":99999999}
→ 422 {"error":{"base":["This asset folder is not valid"]}}

POST /v1/spaces/290020686946333/assets  {"internal_tag_ids":[99999999]}
→ 422 {"error":{"internal_tag_ids":["Invalid internal_tag, there is at least one internal_tag not found in our database"]}}
```

Real CLI output — the new warn adds a name but **no reason**:
```
⚠️  Failed to push "https://a.storyblok.com/f/290020686946333/54cbaa426a/bad1.png": The request was well-formed but was unable to be followed due to semantic errors
```

`storyrails` documents `error` as `oneOf` string | `string[]` | map, plus a separate `errors` map. Only the string branch is implemented. The DX-541 target message *is* a string so that one case works.

## Warn label shows the CDN URL, not the filename

`asset.filename ?? asset.short_filename ?? asset.id` — for a local sidecar, `filename` is the full remote URL, so it always wins. The PR description advertises `Failed to push "demo-one.txt"`; real output is a 140-char line. `short_filename` is the readable field.

## One warn per failed asset, ungrouped, while progress bars are live

60 assets failing with one identical cause → 60 identical 140-char lines. BEFORE printed nothing to the terminal (log file only):

**BEFORE:**
```
 Assets...  [■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■] 100% | 0s | 60/60 processed
ℹ Push results: 0 processed, 1 assets failed
```

**AFTER:** 61 warn lines, then the bars. Contradicts the convention documented at `assets/transfer/index.ts:126` ("Group failures by reason so the output stays a short summary even when thousands of assets fail with the same cause").

## REGRESSION: every 401 degrades to the bare word "Unauthorized"

MAPI's 401 body is `{"error":"Unauthorized"}` (confirmed in `--verbose` `responseData`), so the actionable string is replaced by a single word. This is the most common real failure (expired/wrong PAT) on all ~90 `handleAPIError` call sites that don't pass a `customMessage`.

**BEFORE:**
```
 Error 

▲ error Failed to pull component internal tags
▲ error The user is not authorized to access the API
```

**AFTER:**
```
 Error 

▲ error Failed to pull component internal tags
▲ error Unauthorized
```

`--verbose` is now self-contradictory — `cause` is left generic while `message` is the degraded one:
```
▲ error Unauthorized
▲ error API Error: The user is not authorized to access the API {
  message: 'Unauthorized',
  cause: 'The user is not authorized to access the API',
  responseData: { error: 'Unauthorized' }
}
```

`storyblok user` is unaffected (it passes a `customMessage` for 401) — verified identical before/after.

## No observable change for stories push

Real story 422s are field maps (`{"name":["can't be blank"]}`), no `error` key → code inert. Output byte-identical before/after:
```
▲ error Failed stories (2):
  about (uuid: b9a458b5-…, file: about_….json)
    • Failed to create story
        └─ The request was well-formed but was unable to be followed due to semantic errors
        └─ name: can't be blank
```
The PR description's claim that "Stories failures also now show the server message" is not reproducible against real MAPI. Same for 404 (`["This record could not be found"]` — array shape, unhandled): identical before/after.

## Latent: 422 with both `error` and field errors orders wrong

Not observed live, but spec-permitted (`additionalProperties` alongside `error`). The 422 block runs first and pushes field lines, so `messageStack`'s tail is no longer the generic entry — the replace branch misses and the `else` appends the server message *after* the field errors, leaving the generic boilerplate in place. Fix by splicing the generic entry by index, or computing `serverMessage` before the 422 block.

## Recommendations

Blocking:
1. Don't warn on the missing shared-assets directory — distinguish the per-file catch from the outer directory catch (`streams.ts:569` vs `:574`); the outer one should stay log-only, and the per-file one should include `binaryFilePath`.
2. Handle the object/array forms of `error` plus the `errors` map, or the fix misses the majority of real asset failures.
3. Skip the override when the server string adds nothing over `statusText` (fixes the 401 degradation), and keep `cause` in sync with `message`.

Non-blocking: prefer `short_filename` in the warn label; group warns by reason and render after `stopAllProgressBars()`; update the PR description's before/after block to real output.

packages/cli/src/utils/error/error.ts
dipankarmaikap: Removed

packages/cli/src/commands/assets/pipelines.ts
dipankarmaikap: Fixed

maoberlehner (approved):

PR #734, opened 2026-08-04
feat(experiments): make track stateless so conversions work across requests

maoberlehner:
Makes `track` usable from SSR, where the click that converts arrives in a different request than the render. Non-breaking except for two type-level changes (see below).

Docs: storyblok/storyblok-docs-platform#608
Decision record: `adr/0012-experiments-factory-is-a-stateless-config-binding.md`

## The problem

`track` read assignments from a `Map` inside the factory closure, filled only by `resolveExperiment`:

```ts
const assignments = new Map<number, Assignment>(); // keyed by experiment id only
```

In SSR the factory lives for one render. The conversion arrives in a later request against a fresh instance with an empty map, so **`track` silently recorded nothing**. Refilling the map meant calling `resolveExperiment` again, which fired a second exposure and skewed the denominator, and neither `CreateExperimentsOptions` nor the resolve options could suppress it.

A shared instance was worse than useless: because the map is keyed by experiment id with no visitor dimension, two concurrent visitors overwrite each other and one visitor's conversion gets attributed to the other's variant. Bounded memory, corrupted attribution, non-deterministic under load.

## The fix

Bucketing is a deterministic hash of `visitorId` + experiment id (`assign-variant.ts`), so the assignment was always recomputable and never needed remembering. The map was a cache, not state.

`track` now takes an explicit `visitorId` and works from any request, runtime, or instance. The factory holds no per-visitor state, so one instance is safe to create at module scope and share.

```ts
// Render (any request)
const { slug, delivered } = exp.resolveExperiment({ slug: "home", visitorId });

// Conversion (any other request, any runtime)
await exp.track("signup", visitorId, { value: 4900 });
```

## Added

- **`track(goal, visitorId, options?)`** returning a promise, with `value` and `props`. `void` to `Promise<void>` is safe for callers who ignore it, and the promise never rejects (failures still go to `onError`), so an un-awaited call cannot produce an unhandled rejection.
- **`assignments(visitorId)`**: one assignment per running experiment the visitor is bucketed into, delivering nothing. This is the attribution step `track` performs implicitly, made public so a caller can scope or deduplicate before anything is recorded.
- **`createEvent(goal, assignment)` / `createConversion` / `defineGoal` / `send`**: building one conversion payload without delivering it, and delivering one that arrived from the browser. Callers previously had to hand-assemble the payload the factory built internally as `resolved.exposure` and then discarded. `createConversion` is the bare, pure equivalent, joining `assignVariant` and `resolveExperiment` as a standalone function.
- **`exposure: false`** on `resolveExperiment`, plus the `exposure` descriptor and a `delivered` promise on its result. `resolveExperiment` has to stay synchronous because its return value is what you render, so it returns its delivery instead of taking another callback.
- **`visitorId` and `value` on `ExperimentEvent`**, `visitorId` and the experiment's identity on `Assignment`. Without a visitor on the event, a destination cannot join a conversion to its exposure or count each visitor once per variant. Carrying the experiment's identity rather than only its id is what makes building an event a pure, total function with no lookup and no `undefined` to handle.
- **`findExperimentBySlug`** (also de-duplicates the lookup that existed twice internally), **`beaconAdapter`**, and **`baseUrl`** on `fetchAdapter`, which now rejects an unresolvable url at construction rather than failing on every delivery where `onError` swallows it. A relative url is accepted in a browser, where it resolves against the current page.

## Naming

`defineGoal` returns a **`ConversionGoal`**: the *input* to `track` and `createEvent`, as distinct from the `ExperimentEvent` they *produce*, which additionally carries the experiment, the variant, and the visitor.

## The assignment is the unit

`track` does three things: bucket the visitor, build an event, deliver it. Keeping all three welded together made `createEvent(goal, visitorId)` return one conversion per running experiment, because a goal name does not say which experiment it belongs to. That put a plural return behind a singular name, and it left a caller who wanted to scope a goal or honor a per-visitor cookie filtering a list of *built events*, when the thing being scoped is the *assignment*.

So the two steps are public, and the plural one is the plural thing:

```ts
// Every experiment this visitor is in. Nothing delivered.
const assignments = exp.assignments(visitorId);

// Keep what should count, then build and send one event each.
const scoped = assignments.filter(a => notYetCounted(a.experiment.id));
await Promise.all(scoped.map(a => exp.send(exp.createEvent("signup", a))));
```

`track(goal, visitorId)` is unchanged and is now documented as sugar for exactly that loop. It still records against every running experiment, which is correct when one experiment is live and over-reports when several are: the difference is that the over-reporting is now visible in the API rather than only in a docs warning, and the fix is ordinary code at the call site instead of a join in the sink.

A goal-to-experiment mapping in `createExperiments`, mirroring the backend's `ExperimentAssignedMetric`, would make this declarative. It is deliberately left out: the platform will eventually serve that mapping in the experiments payload, and a filter covers the same need today without adding config that later has two sources of truth. ADR-0012 records it.

## Backward compatibility

`track(goal, props?)` still works and is deprecated. The two forms are discriminated on **`typeof`**, not on the presence of a `visitorId` key: `track('signup', { visitorId })` was a real 1.x call, and key-sniffing would silently reinterpret it as the new shape with no type error. There is a test for exactly that.

`waitUntil` stays a construction-time option; per-call delivery is handled by the returned promises instead.

**Two type-level breaks:**

1. `visitorId` is required on `ExperimentEvent` and `Assignment`, and `Assignment` gains a required `experiment`. Adapters only read events and are unaffected; this breaks code that *constructs* either by hand, which `assignments` and `createEvent` now remove the reason to do. It surfaced in exactly two places in this repo, the `adapters.test.ts` fixture and four literals in the e2e spec, which is the blast radius. Optional would mean destinations cannot rely on the field, defeating the point. `Assignment.experimentId` stays as a deprecated alias for `experiment.id`, so code that *reads* it keeps compiling.
2. Argument-bag types are no longer exported from the package root: `AssignVariantOptions`, `ResolveExperimentOptions`, and `FactoryResolveOptions`. A call site passes an object literal, so annotating these is rare. Option objects that callers genuinely assemble and pass around (`CreateExperimentsOptions`, `TrackOptions`) stay exported.

**Two runtime breaks**, both turning a silent failure into a loud one, so neither is caught by the type checker:

1. `track(goal, secondArgument)` throws a `TypeError` when the second argument is present but is neither a string nor an object. That value is nearly always a `visitorId` that went missing, an unset cookie reaching `track(goal, cookies.get("sb_vid")?.value)`, and letting it through routed the call to the deprecated path and attributed the conversion to whichever visitor the instance resolved last. On the module-scope instance this PR recommends, that is a different visitor's request, and the corruption is invisible in the sink. Arity separates the cases, since a default parameter value applies to an explicitly passed `undefined` as well as to an omitted argument. `track(goal)` stays valid. The cost: forwarding an optional props bag, `track(goal, maybeProps)`, now throws when the bag is absent, which type-checks today. Loud breakage of an already-deprecated form beats silent cross-visitor corruption. The guard throws synchronously rather than rejecting, so `track` keeps its promise-never-rejects contract.
2. `fetchAdapter` resolves its url at construction and throws on a relative path it cannot resolve, where 1.x constructed fine and then failed on every delivery into `onError`. Eager validation and a module-scope adapter shared between server and client code pull in opposite directions: `fetchAdapter("/api/events")` in such a module now throws on the server even when only the browser delivers through it. Deferring the throw would restore the silent per-delivery failure, so the check stays eager and the JSDoc names the ways out, which are `baseUrl`, an absolute url, building the adapter where it is used, or `beaconAdapter` on the client.

`props` is deliberately **not** renamed to `properties` (which is what most analytics platforms call it) to avoid breaking adapters reading `event.props`. Noted in the `2.0 cleanup` comment.

`createEvent` changing from `(goal, visitorId) => ExperimentEvent[]` to `(goal, assignment) => Conversion` is not a break: it was added on this branch and has never shipped. `1.1.0` has no `createEvent`.

## Deliberately not changed

Exposures are still delivered once per `resolveExperiment` call and **not deduplicated**, so a reload or a layout and page that both resolve produce more than one event.

This does not distort the experiment, because the unit of analysis is the visitor rather than the event: the sink counts distinct `visitorId` per variant, which it needs regardless, since a visitor returning tomorrow is legitimately exposed again. Any cache the SDK kept would live in one process, and consecutive requests from one visitor land on different serverless instances, so it would suppress some duplicates, miss others, and imply a guarantee the package cannot make. The tools that do deduplicate rely on a per-request context object, which is precisely what this PR removes.

`exposure: false` covers prefetches and bots by handing back the descriptor to send once the page is really viewed. Documented under "Counting exposures" rather than silently assumed.

Fixes DX-542


packages/experiments/test/specs/experiments-round-trip.spec.e2e.ts
alexjoverm: @maoberlehner seems this still uses the previous signature, not the new one defined. Is it on purpose?

https://github.com/storyblok/monoblok/pull/734/changes#diff-aaffadf16f70517fbaef5dbb55e7e6b0b716e372e651ea5fbff6bb6d9887711cR241-R245

alexjoverm (commented):
@maoberlehner posted couple of questions to look into

packages/experiments/test/specs/experiments-round-trip.spec.e2e.ts
maoberlehner: Cleaned up the tests so that we have one for the new approach while also making sure the backward compatible approach keeps working for now.

alexjoverm (approved):
Great one Markus!

PR #736, opened 2026-08-05
feat(cli): add schema validate command

maoberlehner:
Adds `storyblok schema validate <entry-file>`, which checks a local code-defined schema statically. No login, no space, no API calls.

It reports missing or duplicate block names, field names, and datasource slugs and names, plus cross-references that resolve to nothing: `allow` entries, field `datasource`, and `custom` field plugins. `--level error|warning` sets the display threshold and `--format pretty|json` selects the output format.

## Shared validation layer

This PR also introduces `packages/cli/src/lib/validation/`, the layer both `schema validate` and `stories validate` commands build on: issue filtering by level, grouping by entity, pretty and JSON formatting, and the run summary written to the reporter.

The layer wraps `validateStory` alongside `validateSchema`, and exports `entityToRef` and `entityToHeader`, which this command does not use yet. That is deliberate: `stories validate` is the second consumer and lands in the stacked PR on top of this branch. All of it is covered by `validation.test.ts`.

Schema module loading moves out of `schema push` into `src/utils/schema/classify-exports.ts` so both commands classify a schema module the same way, and `isRecord` is deduplicated into `src/utils/object.ts`.

`UI.writeMachineOutput()` writes `--format json` straight to stdout behind a one-time EPIPE guard, so `storyblok schema validate ... --format json | head -5` no longer crashes with a stack trace and a clobbered exit code.

## Related PRs

| PR | What it covers |
| --- | --- |
| **#736 (this one)** | `schema validate` and the shared validation layer |
| #737 | `stories validate`, stacked on this branch |
| storyblok/storyblok-docs-platform#606 | Reference docs for both commands |
| #669 | The original combined PR, superseded by this split |

## Notes for reviewers

This branch is the first half of a split. The second half, `stories validate` (#737), is stacked on top of it and should be reviewed after this one.

Fixes DX-452


packages/cli/src/lib/validation/adapter.ts
```diff
@@ -0,0 +1,85 @@
+/**
```
alexjoverm, line 1: suggestion(non-blocking): from an architecture perspective, to me this validation looks more like a core module of the CLI rather than simple utils, and contains contextual logic. 

Because of that, I'd suggest to place it under `lib/validation` (also good opportunity, to clog less the `utils`, which IMO already contains things that are closer to modules rather than utils (i.e.: the fetch/api/rate limitting/session/creds could be centralized in an api core module, as they also have app logic - but that's a different story to revisit in future :D))

packages/cli/src/lib/validation/validation.test.ts
```diff
@@ -0,0 +1,446 @@
...
+import type { ValidationRunResult } from './types';
+import { parseFormat, parseLevel } from './types';
+import { countIssues, filterIssuesByLevel } from './filter';
+import { formatJson } from './format-json';
+import { formatPretty } from './format-pretty';
+import { entityToHeader, entityToRef, groupIssuesByEntity } from './group';
+import { writeValidationReport } from './report';
+import type { Reporter } from '../../lib/reporter/reporter';
+
+const error: ValidationIssue = {
```
alexjoverm, line 13: praise: at first I thought we need more tests at unit level, but actually testing the validation module here at integration-level is actually more valuable and maintainable

packages/cli/src/README.md
```diff
@@ -31,6 +31,7 @@ npm install storyblok@<version>
 | [`assets push`](./commands/assets/push/README.md) | ✅ | Upload assets and asset folders to your Storyblok space |
 | [`stories pull`](./commands/stories/pull/README.md) | ✅ | Download stories from your Storyblok space |
 | [`stories push`](./commands/stories/push/README.md) | ✅ | Upload stories to your Storyblok space |
+| `schema validate` | ✅ | Validate a local code-defined schema. Offline — no login, no space |
```
alexjoverm: (blocking): @maoberlehner see latest commit merged in main. This file doesn't exist anymore

alexjoverm (changes requested):

packages/cli/src/commands/schema/validate/index.ts
```diff
@@ -0,0 +1,99 @@
...
+      try {
+        loaded = await loadSchemaEntry(entryFile);
+      }
+      catch (maybeError) {
+        const error = toError(maybeError);
+        // Record a failure so the report reflects the aborted run, not SUCCESS.
+        reporter.addSummary('validation', { total: 1, succeeded: 0, failed: 1 });
+        handleError(error, verbose);
+        logger.error('Schema validate failed to load entry file', { error: error.message });
+        process.exitCode = 2;
```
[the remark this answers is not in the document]
maoberlehner: Fixed in e2501d7. Both overrides dropped; `handleError` owns the mapping.

The option-parse path was redundant — `parseLevel`/`parseFormat` throw `CommandError`, which already exits 2. The entry-load path was the real one, and it splits rather than flattens:

- missing entry file, or one exporting no definitions → `CommandError` → 2 (unchanged)
- entry file throws on import (syntax error, unresolvable dependency) → plain `Error` → 1

I kept 1 there rather than wrapping the jiti failure in a `CommandError` to preserve exit 2: wrapping discards the original error, and `--verbose` needs it to name the offending line in the schema file.

Added a test pinning the CommandError side (`should exit 2 when the entry file does not exist`), which was previously only covered by a message assertion.

packages/cli/src/lib/validation/adapter.ts
maoberlehner: Agree. Moved.

alexjoverm (approved):

dipankarmaikap (approved):

PR #738, opened 2026-08-05
chore(agents): fix drift in agent steering docs

maoberlehner:
Audit and cleanup of the agent steering docs: `AGENTS.md` (root and `packages/cli`), `docs/`, `.agents/skills/`, and `.claude/rules/`. Every path, script, symbol, and skill name referenced in those files was checked against the tree.

## Broken references

- **Dead paths in the assets test plan example.** It named three scripts that do not exist (`asset-create.sh`, `png-create.sh`, `scenarios/has-nested-asset-folders.sh`). Replaced with the real entry points. Also dropped its `cd ./packages/cli`: the scripts resolve `.env.qa-engineer-manual` from `$PWD` and the script paths are repo-root relative, so every command in that file was broken from inside the package directory, not just the three.
- **`artifact-naming.md` claimed to cover every artifact** but omitted `/plan` and `/triage`, both of which write files. Added.
- **`docs/testing-patterns.md` pointed at `qa-engineer-integration`**, a name only `packages/cli`-scoped work could resolve.

## One directory, two spellings

Skills live in `.agents/skills/`, and `AGENTS.md` said so, but every reference *inside* the skills said `.claude/skills/`. Those resolved only through the tracked `.claude/skills -> ../.agents/skills` symlink, so the repo taught both paths. All references now use `.agents/skills/`, including the one functional use (`seed-scenario.sh` resolved `skill_dir` through the symlink) and three package test guides. The symlink stays, since Claude Code discovers skills through it.

## Skill descriptions

Rewrote nine `description:` fields as triggering conditions ("Use when...") instead of workflow summaries. A description that summarizes the process becomes a shortcut agents take *instead of* reading the skill body.

## packages/cli/.claude removed

It held one skill, `qa-engineer-integration`, whose patterns were already documented in `packages/cli/AGENTS.md`, plus a `.gitignore` for a `tmp` directory nothing writes to. Four things in it were not documented elsewhere and moved out first:

- `resolveCommandPath` + `directories` constants for test paths, replacing a hardcoded `.storyblok/...` in the cli example
- `onUnhandledRequest: 'error'` on the msw server, now in `docs/testing-patterns.md` as general guidance (17 of 85 cli test files already pass it; the documented examples did not)
- precondition naming, including failure preconditions
- the `should` test-title convention, now a repo-wide testing principle (936 of ~975 `it(` calls in the cli already follow it)

## Also

- `docs/cli-architecture.md` deleted; the parts that survived verification folded into `packages/cli/AGENTS.md`. Three of its claims were stale and deliberately not carried over: `utils/ui.ts` does not exist, command registration happens in `src/index.ts` rather than `program.ts`, and `konsola` has no references in the package.
- Root `AGENTS.md` restructured: the no-commit rule scoped to `main`, `../storyblok-docs-platform` documented as a sibling repo, two dead pointers removed (`../storyrails/monoblok.md` and `../storyfront/monoblok.md` were supposed to offer opt-in steering, but the same can be achieved by modifying custom AGENTS/CLAUDE.md files).


.agents/skills/triage/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: triage
-description: Triage GitHub issues or Linear tickets — classify type, assess priority, check reproducibility, detect duplicates. Use when asked to triage, assess, or categorize issues.
+description: Use when asked to triage, assess, or categorize GitHub issues or Linear tickets
 model: haiku
```
alexjoverm: @maoberlehner what do youthink we change this to sonnet? Over time, I realize that while haiku is usually ok for triagging, sometimes it misses some "deeper" considerations. And sonnets might be a good middleground on lighter-yet-good-enough

docs/docs-platform.md
```diff
@@ -1,14 +1,6 @@
 # Storyblok Docs Platform
 
-**Repo:** storyblok-docs-platform (via `additionalDirectories`)
```
alexjoverm, line 3: @maoberlehner I understand this isn't necessary in order for claude finding the other repos?

packages/cli/AGENTS.md
```diff
@@ -16,6 +17,10 @@
 | Structured logs | `src/lib/logger/` |
 | Machine-readable command reports | `src/lib/reporter/` |
 
+## Module initialization
```
alexjoverm, line 20: Great addition, I realized over time that claude tends to not-know about this core area

.agents/skills/blitz/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: blitz
```
alexjoverm, line 2: @maoberlehner what do you think about this skill? Do you use it often from your side?

While I initially thought is a great idea, over time I realized it does too much and I usually take a bit more granular approach when it comes to working on multiple issues

alexjoverm (approved):

.agents/skills/blitz/SKILL.md
maoberlehner: The only skills/features I find myself manually triggering deliberately are `/review` and `/qa-engineer-manual`. Otherwise, I let the agent decide to load what it finds useful, and according to my stats, this skill was not chosen by it.

docs/docs-platform.md
maoberlehner: The agent will ask for permission when reaching outside of its working directory. `additionalDirectories` is the mechanism behind this, but it knows this by itself.

.agents/skills/triage/SKILL.md
maoberlehner: changed! Agree, Haiku should only be used for the simplest kind of tasks.

PR #739, opened 2026-08-06
fix(cli): remove conditional isVitest references

dipankarmaikap:
## What

Removes all `isVitest` / `isVitestRunning` references from the CLI package — an antipattern that coupled production code to test-environment detection.

## Why

Test-environment guards in production code are an antipattern. The spinner verbosity should be controlled by the existing `--verbose` global config option, not by sniffing `process.env.VITEST`.

## Changes

- **`src/lib/ui/ui.ts`** — replaced `verbose: !isVitest` with `verbose: getActiveConfig().verbose`; swapped the `isVitest` import for `getActiveConfig` from the config store
- **`src/utils/index.ts`** — deleted the `isVitest` export entirely
- **`src/commands/create/index.test.ts`** — removed dead `isVitest: true` entry from the utils mock (UI is already fully mocked via `vi.mock('../../lib/ui')`)
- **`src/commands/login/index.test.ts`** — removed entire `vi.mock('../../utils')` block; the only entry was a misnamed `isVitestRunning: true` key that had no effect
- **`src/commands/types/generate/index.test.ts`** — same as above

## Behaviour

Spinner `verbose` is now `false` by default (matching `GlobalConfig.verbose`), meaning spinner frames are silent unless the user passes `--verbose`. In tests the default config is unchanged (`verbose: false`), so no `isVitest` guard is needed.

Fixes DX-533

### DIFF
--- a/packages/cli/src/commands/create/index.test.ts
+++ b/packages/cli/src/commands/create/index.test.ts
@@ -43,7 +43,6 @@ vi.mock('../../utils', async (importOriginal) => {
     requireAuthentication: vi.fn(),
     handleError: vi.fn(),
     toHumanReadable: vi.fn(),
-    isVitest: true,
   };
 });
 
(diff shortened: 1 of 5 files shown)


maoberlehner (approved):
Nice one!!

PR #742, opened 2026-08-10
fix(cli): generate richtext types from OpenAPI spec and fix StoryblokRichTextDoc name

dipankarmaikap:
## Summary

Fixes the CLI's richtext type generation by sourcing `StoryblokRichTextDoc` from the OpenAPI spec (same approach as the richtext package) and corrects the type name used in generated component imports.

## Changes

**`packages/cli/package.json`**
- Add `generate:openapi` npm script (`tsx scripts/generate.ts`)
- Add `@storyblok/openapi-codegen` and `tsx` as devDependencies
- Add NX `generate:openapi` target with `{projectRoot}/package.json` in inputs (same cache-invalidation fix as the richtext package — so the task re-runs on version bumps)

**`packages/cli/scripts/generate.ts`** _(new)_
- Fetches OpenAPI spec and emits `src/generated/overlay/types.gen.ts` using `@storyblok/openapi-codegen`

**`packages/cli/src/generated/overlay/types.gen.ts`** _(new)_
- Generated richtext types from the OpenAPI spec

**`packages/cli/src/types/storyblok.ts`**
- Re-export `RichTextDoc` as `StoryblokRichTextDoc` from generated types
- Keep `StoryblokRichtext` as a deprecated alias for backwards compatibility
- Remove the hand-written `StoryblokRichtext` interface

**`packages/cli/src/commands/types/generate/utils.ts`**
- Add `TYPE_NAME_OVERRIDES` map and `storyblokTypeName()` helper so `richtext` resolves to `StoryblokRichTextDoc` instead of the incorrect `StoryblokRichtext`
- Apply `storyblokTypeName()` in all import/detect helpers

**`packages/cli/src/commands/types/generate/utils.test.ts`**
- Update assertions to match new `StoryblokRichTextDoc` type name

## Type of change
- [x] Bug fix

Fixes DX-494


packages/cli/src/commands/types/generate/utils.ts
```diff
@@ -1,5 +1,20 @@
...
+const TYPE_NAME_OVERRIDES: Record<string, string> = {
+  richtext: 'StoryblokRichTextDoc',
+};
+
+function storyblokTypeName(type: string): string {
+  if (TYPE_NAME_OVERRIDES[type]) {
+    return TYPE_NAME_OVERRIDES[type];
+  }
+  return `Storyblok${toPascalCase(type)}`;
+}
```
maoberlehner, line 16: In `generate/actions.ts` we still use `const componentType = `Storyblok${toPascalCase(propertyType)}`;` I think it should use `storyblokTypeName()` too?

maoberlehner (changes requested):
`packages/cli/src/commands/types/generate/README.md` still names StoryblokRichtext.

packages/cli/src/commands/types/generate/utils.ts
dipankarmaikap: Fixed

maoberlehner (approved):

PR #748, opened 2026-08-11
fix(experiments): rename adapter factories to create* names

maoberlehner:
The package uses "adapter" for two different things. An adapter is `(event) => unknown` (the `Adapter` type). The functions exported from `@storyblok/experiments/adapters` are `(config) => Adapter` — factories, not adapters. The names, the JSDoc ("An adapter that POSTs each event as JSON"), and the docs site all blurred the two, which is confusing when you follow a symbol to its definition.

## What

- `fetchAdapter` → `createFetchAdapter`, `beaconAdapter` → `createBeaconAdapter`. The `create*` prefix is what the package already uses for `createExperiments` and `createConversion`, so this makes the adapters entrypoint consistent with the rest of the API rather than introducing a new convention.
- `FetchAdapterOptions` → `CreateFetchAdapterOptions`, `BeaconAdapterOptions` → `CreateBeaconAdapterOptions`, matching the `<functionName>Options` pattern of `CreateExperimentsOptions` / `CreateConversionOptions`.
- JSDoc now says a factory *creates* an `Adapter`; the `Adapter` type doc states the two-term distinction explicitly.
- Error message prefixes follow the rename, so a message someone hits in production greps back to the exported symbol.

## Backward compatibility

All four old names remain exported as `@deprecated` aliases, so this is non-breaking and can ship in a minor. They are removed in the next major.

## Follow-up

The docs site uses the old names throughout and calls the factories "adapters". Terminology fix is going onto storyblok/storyblok-docs-platform#608; the rename itself needs a docs pass once this is released, with the version badge for whichever minor it lands in.

Fixes DX-546

### DIFF
--- a/packages/experiments/src/adapters.ts
+++ b/packages/experiments/src/adapters.ts
@@ -2,7 +2,7 @@ import type { Adapter, ExperimentEvent } from "./types";
 
 export type { Adapter } from "./types";
 
-export interface FetchAdapterOptions {
+export interface CreateFetchAdapterOptions {
   /** Override the `fetch` implementation (e.g. for testing or a custom client). */
   fetch?: typeof globalThis.fetch;
   /** Extra headers merged onto the request. */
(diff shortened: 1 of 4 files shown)


packages/experiments/src/adapters.ts
```diff
@@ -91,40 +91,69 @@ export function fetchAdapter(url: string, options: FetchAdapterOptions = {}): Ad
...
+ * factory. Kept for backward compatibility, removed in the next major.
+ */
+export type BeaconAdapterOptions = CreateBeaconAdapterOptions;
+
+/**
+ * @deprecated Renamed to `createFetchAdapter`, which reflects that it builds an
+ * `Adapter` rather than being one. This alias is kept for backward
+ * compatibility and will be removed in the next major.
+ */
+export const fetchAdapter = createFetchAdapter;
```
alexjoverm, line 152: @maoberlehner while it's re-exposed using an alias, the PR is named as `refactor:` - that won't trigger a release. An alias-like change needs to be driven as a `fix:`

alexjoverm (commented):
Approving as doesn't need code re-review, but just a comment to apply at PR title/commit prefix

alexjoverm (approved):

PR #753, opened 2026-08-13
fix(schema): check defineField options against the field type

maoberlehner:
> [!IMPORTANT]
> **Stacked on #752 (`fix/schema-deny-leaks-to-wire`)** and targets that branch, not `main`.
> Review that one first. This PR depends on it in two ways: `component_denylist` can only point at
> `deny` once `deny` reaches the wire, and `component_group_denylist`'s bucket depended on that
> branch's folder-ref decision.

## Two defects

**1. No excess property checking.** `defineField` accepted *any* key. The field literal is inferred
*as* the type parameter, so the only check was plain assignability to the constraint, which ignores
excess properties. Typos and options meaningless for the field type compiled clean, were pushed to
the Management API verbatim, and silently did nothing:

```ts
defineField('x', { type: 'bloks', component_group_whitlist: ['a'] }); // typo, ignored
defineField('x', { type: 'text',  component_group_whitelist: ['x'] }); // meaningless on text
defineField('x', { type: 'asset', restrict_components: true });        // meaningless on asset
defineField('x', { type: 'bloks', totally_bogus_key: 123 });           // anything goes
```

The DSL keys had the same hole, in the place it matters most: `allow`, `deny` and `datasource` were
exempt on every variant, so `defineField('x', { type: 'text', allow: ['teaser'] })` compiled and
reached the wire, while the raw `component_whitelist` form on that same field was already rejected.

**2. `allow`/`deny` and the raw wire keys were silently both legal.** `mapFieldToWire` spreads
`...rest` and *then* assigns from `allow`, so a hand-written `component_whitelist` was overwritten
without warning. Setting both is always a mistake.

## Approach

- `NoExtraKeys<T>` gives excess property checking the concrete target it lacks, keyed off the matched
  `Field` variant so the check is per field type. `type: 'custom'` is exempt, since plugin option
  keys legitimately pass through verbatim.
- Each DSL key is legal only on the variants that declare the wire key it maps onto, read off the
  variant's own declarations rather than a hand-kept list of field types. `FieldInput` is scoped by
  the same rule, so it stays assignable to `defineField` on every variant, with a test that guards
  that.
- `NoRestrictionConflict<T>` makes mixing `allow`/`deny` with the five keys they derive a compile
  error. It treats a present-but-`undefined` `allow` as absent, so a conditionally built restriction
  next to a raw wire key still compiles: nothing gets derived at runtime. `validateSchema` reports
  the same conflict as a `conflicting_restriction` issue for consumers authoring schemas in plain
  JavaScript. Both read from one `DERIVED_RESTRICTION_KEYS` list, so they cannot drift.
- Rejected keys resolve to `Invalid<TReason>`, an unsatisfiable `unique symbol` branded type whose
  argument carries the reason into the error:

  ```
  Type '["teaser"]' is not assignable to type '["teaser"] & Invalid<"\"component_whitelist\" is
  derived from \"allow\"/\"deny\": set one or the other, not both">'.
  ```
- `CheckedField<T>` is exported so a wrapper around `defineField` can type its own parameter. The
  checks are mapped types over the type parameter, and TypeScript cannot prove an unresolved `T`
  satisfies a mapped type of itself, so forwarding a still-generic field otherwise fails with
  `TS2345`. Typing the wrapper's parameter with `CheckedField<T>` also moves the check out to the
  wrapper's call site, where the literal is actually written.
- `restrict_type` typos are reported by `validateSchema`, not by the type. The API never validates
  that key, so a space can hold anything there and `schema init` has to emit what it finds.
  Narrowing the type would buy typo protection by making a stored value untypeable, which is the
  failure this area exists to remove, so the warning catches the typo on the write path and leaves
  the read path intact.
- The restriction constants live in `src/restrictions.ts`, so the validators no longer import the
  helpers layer at runtime to read one list.

The wire keys stay **legal** as a lower-level escape hatch. They are not removed, and
`restrict_components: false` and `restrict_type: 'tags'` plus the tag lists still compile.

## `schema init` fixes in the same area

Manual QA against a QA space surfaced two `schema init` bugs in `packages/cli`, fixed in the
`fix(cli)` commits here since they land in the same feature area and this PR's docs point straight at
them.

**The tag dimension was dropped.** `resolveFieldRestriction` classified a tag-restricted field as
`kind: "none"`, the branch for fields with no restriction in force, which drops
`restrict_components` and `restrict_type` as byproducts `allow`/`deny` re-derive on push. Nothing
re-derives them for the tag dimension, because `allow`/`deny` cannot express tags at all. So
`schema init` followed by `schema push` pushed back a field whose tag lists the editor never reads,
silently restricting by block name instead.

The fix classifies a tag restriction as its own kind, keyed on `restrict_type` **and** a non-empty
tag list. `restrict_type` alone is not enough: the editor clears all six lists on a dimension switch,
and the Management API backstops that for `bloks` fields only, so a `richtext` can hold
`restrict_type: 'tags'` next to a live `component_whitelist` from legacy or direct API writes. The
editor gates every dimension on a non-empty list and reads the name list in exactly that case, so a
tag dimension that selects something keeps its wire form with the flags intact, while an empty one
falls through to whichever list is actually in force.

**Init was not idempotent across a push.** Push always sends the five optional metadata keys with
their reset value, turning an unset field from `null` into `""` in the space, and init emitted that
back, so "init, commit, push" produced a spurious diff on the next init for fields nobody had set.

Verified end to end against the `has-restrictions` scenario, which covers all nine restriction
flavors: the generated code typechecks clean under the stricter `defineField`, no field round-trips
to a semantically different value, and a second push reports every entity `unchanged`. The *first*
push after `init` is not a no-op on an editor-authored space, because the editor always writes the
paired empty list and `init` drops it, but every one of those removals is inert, since the editor
gates each list on a non-empty length. Regression tests cover the lossless round-trip, the
stale-name-list case, and the gate.

## Notes for review

**`@deprecated` cannot produce an inline strikethrough on the wire keys, and that is a TypeScript
limitation.** Probed with the language service on TS 5.8.3 and 6.0.3:

- A key declared on *both* intersection members with only one tagged reports as **not** deprecated,
  order independent, unions included. TypeScript requires *every* declaration to carry the tag. Hence
  `WireRestrictionDocs`, a distributive `Omit` over `Field` that re-picks the five keys, so
  `FieldInput` is the only declaration site. `restrict_type` and the tag keys correctly stay
  undeprecated.
- TypeScript reports no deprecation on object literal property *writes*, and none on property
  *reads* either: not for interfaces, type aliases, intersections, unions, or destructuring. What the
  tags deliver is the struck-through **autocomplete** entry with its reason and example, plus the
  reason in the conflict error, and nothing beyond that. An inline nudge on a written
  `component_whitelist:` key would need a lint rule.

## Accepted breaking change

This is a type-level break shipped as `fix(schema):`, a patch bump that every consumer on `^0.4.0`
picks up automatically. An existing `defineField` call with an excess key, a typo, or `allow`
alongside a derived wire key becomes a build failure. That is the point of the PR, and 0.x is the
right time for it, but it needs to be called out in the release notes rather than reading as a plain
fix. The docs-site schema page also still describes the wire props as escape hatches with no mention
of the new `allow`/`deny` exclusivity.

## Known gaps, out of scope here

- `schema push` never calls `validateSchema` (only `schema validate` does), so the new
  `conflicting_restriction` diagnostic protects JS authors only if they run `validate` first. The
  compile-time rejection is the real guard.
- `restrict_type` is itself overwritten by `allow`/`deny` at push time (`map-to-wire.ts`), but is
  deliberately not in `DERIVED_RESTRICTION_KEYS`, so neither the type nor the validator flags
  `allow` + `restrict_type: 'tags'`. The tag list ends up inert. Worth a follow-up.

Fixes DX-550


dipankarmaikap:
## 🤖 AI Findings

### 1. `MemberFor<T>` fallback should be `never`

```typescript
// define-field.ts
type MemberFor<T> = T extends { type: infer TType } ? Extract<Field, { type: TType }> : Field;
```

The `else` branch falls back to the entire `Field` union. Since `FieldInput` is always discriminated by `type` this branch is unreachable in practice, but `never` would be a more defensive and self-documenting choice than silently widening to the full union.

---

### 2. `RESTRICTABLE_FIELD_TYPES` and `DENIABLE_FIELD_TYPES` can drift silently

`RESTRICTABLE_FIELD_TYPES` (`Set(["bloks", "richtext"])`) lives in `generate-code.ts` while `DENIABLE_FIELD_TYPES` (`["bloks", "richtext"]`) lives in `restrictions.ts`. They hold the same values for closely related (but distinct) semantics. If a future field type gains restriction support both need updating, but nothing links them so one could be missed. A cross-reference comment, or deriving one from the other, would make the dependency explicit.

---

### 3. `schema push` never calls `validateSchema`

The new `conflicting_restriction` runtime check only fires when a user explicitly runs `schema validate`. A plain-JS author who sets `allow` alongside a raw wire key gets no warning at push time — the wire key is silently overwritten. The compile-time rejection is the real guard, but a follow-up to surface this at push time (at least as a warning) would close the JS-author gap entirely.

---

### 4. `allow` + `restrict_type: 'tags'` compiles but produces inert tag lists

`restrict_type` is correctly excluded from `DERIVED_RESTRICTION_KEYS` (it's the only route to the tag dimension). The side-effect is that `allow: ['teaser'], restrict_type: 'tags'` compiles and pushes cleanly — at push time `restrict_type` is overwritten with `''` and the tag lists become inert without any warning. The PR notes this as a known gap. Worth a follow-up issue so it doesn't get lost.

dipankarmaikap (approved):
Looks good to me and it passed all my manual QA testing. I’m approving it.

I found a few minor points suggested by AI and added them as comments, so you can judge for yourself whether they’re worth addressing.

PR #755, opened 2026-08-14
fix(cli): restore type generation after Vite+ migration

bentbrain:
## Summary

- resolve the packaged declaration file through the CLI package metadata
- keep generated Storyblok types self-contained by excluding Vite+ runtime imports and source maps
- cover `types generate` through the built CLI artifact

## Why

The Vite+ migration changed the published declaration from `dist/index.d.ts` to `dist/index.d.mts`, while `types generate` continued reading the old filename. This makes `storyblok@4.22.1 types generate` fail with `ENOENT`.

Reading the new declaration unchanged would also copy CLI-only imports into generated user types, so this extracts the bundled declaration regions before writing `storyblok.d.ts`.


dipankarmaikap (approved):
Thanks for creating the PR and catching this!

maoberlehner:
Superseded by #757, which carries your two commits unchanged plus follow-up fixes. Opened as a separate PR only because this one is from a fork, so I could not push to its branch. Thanks @bentbrain the `ENOENT` diagnosis was spot on, and chasing it turned up two further bugs in the generated types (a richtext type name that was never imported, and a non-nullable `StoryblokAsset.id`) that predate the Vite+ migration.

PR #756, opened 2026-08-14
fix(schema): match field types, options and QA fixtures to the real wire formats

maoberlehner:
Groundwork extracted from the review of #752 and #753, split out so both of those show only their own change. They are now stacked on top of this branch.

Everything here was checked against one of two sources. Never against a round trip: the Management API stores component schemas and story content as opaque blobs, accepts almost any JSON and echoes it back, so pushing a fixture and reading it back proves storage, not shape.

1. **Sibling-repo source** — `storyfront` for what the editor writes and reads, `storyrails` for what the backend normalizes and enforces.
2. **Operator-authored data** — a block with one field of every type, a block per restriction shape, and stories covering every value shape, created by hand in a QA space and read back via MAPI.

## What's here

**`fix(schema): declare the field options the editor writes`**
`schema init` passes every wire key through verbatim, so an option a space holds but the spec does not declare produces a field the generated types cannot describe. Two were missing: `richtext` `link_scope`/`max_length`/`rtl`, and `options` `exclude_empty_option` (declared on `option` but not its multi-select sibling).

**`feat(schema): support the image and file field types`**
`image` and `file` predate the asset object and store a bare protocol-relative URL string. The editor no longer offers them when adding a field, but spaces still hold them and neither was modelled, so `schema init` emitted `defineField('x', {type: 'image'})`, which failed on the discriminant.

`add_https` and the crop options (`image_crop`, `image_width`, `image_height`, `keep_image_size`) belong to `image`/`file` alone. An `asset` and a `multiasset` created in the editor stay bare, and `storyfront` agrees: `FieldTypeAsset` computes a protocol from `add_https`, but its only consumer, `processFilename`, runs behind a `deprecated` prop that only `FieldTypeImage` and `FieldTypeFile` pass. On an `asset` field the key is inert. `image_width`/`image_height` are `integer | string`: clearing the input writes an empty string rather than removing the key.

**`fix(migrations): remap only the option sources that hold a reference`**
`optionsFieldRefMapper` keyed on `internal_users`, `internal_tags` and `internal_datasources`. None are real option sources — the editor offers `internal_stories`, `internal`, `external`, `internal_languages` and inline options — so those branches were unreachable. Of the real ones only `internal_stories` holds a cross-space reference; the others hold values that mean the same thing in every space, so remapping them would corrupt content. The singular `option` type had no mapper at all.

**`test(repo): match seed fixtures to the wire formats the editor writes`**
The corpus held shapes the editor never writes, and every one passed QA:

- Asset sidecars were named `<file>.png.meta.json` while the CLI resolves `<basename>.json` and swallows ENOENT. **Every seeded asset in every scenario had always uploaded with no metadata, silently.**
- Story `uuid`s were `"1"`…`"10"` and a multilink's `id` was the numeric story id. `mapRefs` keys on the uuid, so the one scenario built to exercise remapping could not have caught a broken remap.
- A component declared `source: "internal_datasources"` — the fixture was shaped to match the `map-refs.ts` bug above, so QA confirmed the bug.
- Folder stories carried `{component: "page"}` instead of `content_types`/`default_root`.
- `blog.json` paired `source: "internal_stories"` with `datasource_slug`, which are mutually exclusive, and it seeds into every scenario.

New coverage: `has-restrictions` holds the first fixtures in the repo containing a `component_denylist`, a `restrict_type`, or any group or tag list — while both stacked PRs are about exactly those. `has-diverse-components` gains `kitchen_sink` for the field types nothing exercised. `has-rich-content` gains a `linktype: "story"` link mark and an embedded `type: "blok"` node, both of which `mapRefs` walks.

**`docs(repo): record that the Management API never validates a shape`**
`AGENTS.md` and the QA skill now state the opaque-blob caveat and give an ordered recipe for grounding a shape: existing seeds, then storyrails, then storyfront, and only ask the operator to author it in the UI when the source does not settle it — an agent cannot produce that evidence itself.

**`feat(schema): type the conditional settings a field can carry`**
`conditional_settings` was an untyped array of objects, so every key inside a rule was `unknown`: a typo in a condition produced no error. The shape comes from `storyfront`, which both writes it (`FieldConditions/index.vue`, `FieldConditionItem.vue`) and evaluates it (`utils/validateConditionalField.ts`) — six validations, two rule matches, two modifications. Nothing is required, because the editor persists a setting the moment a rule row is added, before any of it is filled in. `kitchen_sink` gains the first conditions in the corpus.

---

## Follow-up from review

**The field type union was still incomplete.** The Management API permits 22 field types and rejects anything else on component save; the overlay declared 19, while `packages/mapi-client`'s own generated types already carried all 22. `link`, `group` and `commerce` are added, with the same argument as `image`/`file`: spaces hold them, the editor no longer offers them, and `schema init` passed the wire `type` through into a `defineField` call that failed on the discriminant.

**Two more declarations were narrower than the wire.** `modifications[].display` accepted only `hide`, but the server's conditional-required check compares against `hidden` and its own fixtures author that spelling; both are now declared and the divergence is documented. `max_length` was `integer`, but the schema form persists the number input's raw value, so spaces hold `"60"` as readily as `60`. The three text types that already declared `integer` are widened too rather than left inconsistent, and `checkStringLength` coerces.

**"Further entries are inert" was wrong on the server.** True of the editor, which reads only the first `conditional_settings` entry. The server's conditional-required check requires every setting to match, so an extra entry can only make a field harder to leave empty. That description is inlined as JSDoc into four packages, and its audience is exactly the people who can produce a multi-setting array.

**The `option` mapper fix reached only one of the two forks.** `packages/cli` carries its own copy of the story ref-mapper and had no `option` entry, so a single-select story reference pushed to another space kept the source-space uuid while the multi-select field beside it remapped correctly. Ported rather than shared: the CLI does not depend on `@storyblok/migrations`.

**`has-restrictions` could not be seeded.** It is the first scenario shipping `groups.json`/`tags.json` sidecars; `count_staged` learned to skip them but the `component_group_uuid` guard did not, so the seed refused and left the space empty. Both checks now classify by data shape, the way the CLI does. Staging also purges first, so a killed run cannot leave another scenario's components behind.

**`AGENTS.md` overstated the opaque-blob rule**, which was the load-bearing premise of the section. The API does validate a field's `type` against a closed list, so a bogus `type` returns a 422 naming every permitted value, which is cheaper than reading a sibling repo. Story content is closer to the original claim, but only until a space enables field constraints. Also reflowed to oxfmt, which `pnpm format:check` had been failing on.

**Fixture coverage.** Every option this PR declared went in without one. `kitchen_sink` now carries `link`, `group`, `image_width`/`image_height` as `""` (the cleared-input case the spec calls out), a string `max_length`, a singular `option` sourced from `internal_stories`, plus `link_scope`, `rtl` and `exclude_empty_option`. `SCENARIOS.md` states the real count (20 of 22) and says why `custom` and `commerce` are out.

**New: `hideWhen` / `requiredWhen`.** Typing `conditional_settings` made the shape checkable but left it an 18-line literal per rule, four levels deep, duplicated across three playground files. These collapse it to one line. They need a docs-site entry before release. The playground comments they replaced were also wrong: an unchecked boolean counts as empty, but an *untouched* one is absent, and the editor reads an absent field as no match, so the dependent field is visible on a fresh story.

**Breaking-change note.** `RefMaps` loses `users`, `tags` and `datasources`. The branches were provably unreachable, but it is a type-level break shipped under a `fix:` patch bump.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,8 +31,7 @@ We use `nx` and `pnpm` workspaces. Use commands like `pnpm nx build <package>` a
 - `../storyfront` (headless CMS frontend) - Consult when matching UI/app behavior and you need
   information about the visual editor, bridge protocol, or rendering in the Storyblok UI, and
   whenever you need the real shape of a component schema field or a story content value. The editor
-  is what writes them: `packages/openapi/src/FieldType.ts` declares the shapes, `Schema/*/index.vue`
-  is what the editor **writes**, `EditorForm/FieldType*/` is what it **reads**.
+  is what writes them.
 - `../storyblok-bridge` (Storyblok Bridge) - Consult when verifying bridge behavior: the
   `postMessage` protocol between the editor and a preview iframe, the editable-block attributes it
   writes into the page, the overlay/click-to-edit UI, or the bridge lifecycle.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -46,18 +46,22 @@ These sibling repos may not be available; ignore them if absent.
   reference them, their paths, file names, or internal implementation details in commit messages, PR
   titles and descriptions, issue comments, code comments, or any other public-facing text. Describe
   the observable API or behavior instead.
-- **IMPORTANT:** The Management API stores component schemas and story content as opaque blobs. It
-  accepts almost any JSON and echoes it back, so a successful round trip proves storage, not shape.
-  Never treat "I pushed it and read it back" as evidence that a shape is real: you authored the
-  input, so the check could not have failed. Only two things can ground a shape:
-  1. **Sibling-repo source** - `../storyfront` for what the editor writes and reads,
-     `../storyrails` for what the backend normalizes and enforces.
+- **IMPORTANT:** The Management API validates a component schema field's `type` against a closed
+  list and rejects an unknown one outright. Everything else it treats as an opaque blob: the rest of
+  the schema unconditionally, and story content unless the space has field constraints enabled. So a
+  successful round trip proves storage, not shape. Never treat "I pushed it and read it back" as
+  evidence that a shape is real: you authored the input, so the check could not have failed. Only
+  two things can ground a shape:
+  1. **Sibling-repo source** - `../storyfront` for what the editor writes and reads, `../storyrails`
+     for what the backend normalizes and enforces.
   2. **Operator-authored data** - ask the user to create it by hand in the Storyblok UI, then read
      it back via MAPI. You cannot produce this yourself, so treat it as the last resort, for when
      the source does not settle it.
 
-  The API can still prove server-side *transformation* (normalization, defaults, rejection), just
-  never that a shape is correct.
+  Whatever the API actively enforces, it can prove: push a field with a bogus `type` and the 422
+  names the permitted values, which is cheaper than reading either sibling repo. The same goes for
+  normalization and server-set defaults. What it cannot prove is that a shape it merely stored is
+  one any real space holds.
 
 ## Conventions
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -29,7 +29,10 @@ We use `nx` and `pnpm` workspaces. Use commands like `pnpm nx build <package>` a
 - `../storyrails` (Storyblok backend) - Consult when verifying REST/MAPI/CAPI schemas, error shapes,
   or endpoint behavior; `../storyrails/spec/integration/openapi/` is the source of truth.
 - `../storyfront` (headless CMS frontend) - Consult when matching UI/app behavior and you need
-  information about the visual editor, bridge protocol, or rendering in the Storyblok UI.
+  information about the visual editor, bridge protocol, or rendering in the Storyblok UI, and
+  whenever you need the real shape of a component schema field or a story content value. The editor
+  is what writes them: `packages/openapi/src/FieldType.ts` declares the shapes, `Schema/*/index.vue`
+  is what the editor **writes**, `EditorForm/FieldType*/` is what it **reads**.
 - `../storyblok-bridge` (Storyblok Bridge) - Consult when verifying bridge behavior: the
   `postMessage` protocol between the editor and a preview iframe, the editable-block attributes it
   writes into the page, the overlay/click-to-edit UI, or the bridge lifecycle.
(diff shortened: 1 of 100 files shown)


tools/openapi-codegen/specs/mapi/components/field-types/asset-field.yaml
```diff
@@ -30,6 +30,13 @@ allOf:
       width:
         type: integer
         description: Field width in pixels
+      add_https:
+        type: boolean
+        description: |
+          Whether to prepend `https:` to the asset URL. The schema editor offers
+          this on the legacy `image` and `file` types only, but the asset field
+          component reads it from whatever schema it is handed, so an `asset`
+          field carrying it is honored at runtime.
```
maoberlehner: > but the asset field component reads

what asset field component?

packages/migrations/src/map-refs.ts
```diff
@@ -3,9 +3,6 @@ import type { Component, Story } from "./types";
 export interface RefMaps {
   assets?: Map<unknown, string | number>;
   stories?: Map<unknown, string | number>;
-  users?: Map<unknown, string | number>;
-  tags?: Map<unknown, string | number>;
-  datasources?: Map<unknown, string | number>;
```
maoberlehner, line 8: Explain to me how it was determined that this is  not necessary and double check.

AGENTS.md
```diff
@@ -29,13 +29,28 @@ We use `nx` and `pnpm` workspaces. Use commands like `pnpm nx build <package>` a
...
   lives there, not here: package READMEs stay minimal and link to the docs site.
 
 These sibling repos may not be available; ignore them if absent.
 
+**IMPORTANT:** The Management API stores component schemas and story content as opaque blobs. It
+accepts almost any JSON and echoes it back, so a successful round trip proves storage, not shape.
+Never treat "I pushed it and read it back" as evidence that a shape is real: you authored the input,
+so the check could not have failed. Only two things can ground a shape:
+
+1. **Editor-authored data** - create it by hand in the Storyblok UI, then read it back via MAPI.
```
maoberlehner: agent can't do this, rephrase along the lines of "operator creates it by hand"

tools/openapi-codegen/specs/mapi/components/field-types/asset-field.yaml
```diff
@@ -30,6 +30,13 @@ allOf:
       width:
```
maoberlehner, line 30: general question: do we have support for conditions? One field in test space has a condition activated.

dipankarmaikap (approved):

packages/live-preview/src/generated/types/field.ts
```diff
@@ -154,11 +154,30 @@ interface FieldTypeValueMap {
...
   table: TableFieldValue;
   section: never;
   tab: never;
+  group: never;
+  /**
+   * Owned by a commerce integration rather than the editor, and exempt from
+   * server-side content type-checking, so the integration is the only thing that
+   * knows the shape. Narrow it where you know which one wrote the field.
+   */
+  commerce: unknown;
```
alexjoverm, line 180: future idea (non-blocking): @maoberlehner we're actually owning the [Commerce integrations, supporting the ones here](https://www.storyblok.com/docs/concepts/e-commerce#official-integrations). Probably in the future we can see if we can provide default types for the supported ones?

AFAIK, under the hood they all add a custom field type (or set of them), but not sure if all look the same and have the same properties.

alexjoverm (approved):

PR #759, opened 2026-08-14
test(astro): add unit tests for lib helpers

MatheusMartinho:
## What

Add unit tests for the public helpers in `packages/astro/src/lib/helpers.ts`, which currently have no coverage:

- `useStoryblokApi`: returns the initialized client and throws when initialization has not happened
- `getLiveStory`: returns the preview story and falls back to `null`
- `getPayload`: returns story and server data, and handles missing preview data
- `initStoryblokBridge`: generates the bridge init script with and without a serialized config

## Why

The `tests/` suite already covers the utils (`toCamelCase`, `sanitizeJSON`, `isEditorRequest`) and the Vite plugins, but the `lib/helpers.ts` entry points were untested. They are pure functions plus one global accessor, so the tests need no token or API access.

## Verification

- `vp test run` in `packages/astro`: 9 test files, 143 tests passing
- `pnpm run test:types` in `packages/astro`: clean
- `pnpm nx lint @storyblok/astro` and `pnpm format`: clean

cc @maoberlehner

### DIFF
--- a/packages/astro/tests/helpers.test.ts
+++ b/packages/astro/tests/helpers.test.ts
@@ -1,0 +1,84 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { getLiveStory, getPayload, initStoryblokBridge, useStoryblokApi } from "../src/lib/helpers";
+import type { ISbStoryData, StoryblokClient } from "../src/types";
+
+function makeStory(name = "Home"): ISbStoryData {
+  return {
+    name,
+    id: 1,
+    uuid: "story-uuid",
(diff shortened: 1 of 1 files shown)


MatheusMartinho:
Gentle ping on this one too, whenever there is time. It only adds unit tests for the public helpers in `packages/astro/src/lib/helpers.ts`, which had no coverage, and touches no runtime code. Same as #760, the two required checks are still waiting for the workflows to be approved. Happy to rebase if needed.

PR #693, opened 2026-07-14
fix(cli): resolve tsconfig aliases in config files

Ehtasham-Yasin:
Summary

* Resolve tsconfig.json path aliases when loading storyblok.config.ts
* Pass aliases to c12/jiti
* Add a regression test for alias imports

Closes #511

PR #760, opened 2026-08-14
fix(cli): load compiler options and custom fields parser with jiti

### LINKED ISSUE #307: 🐛 [Bug Cli V4]: Unable to load compiler-options as .ts file
### Package

storyblok (CLI)

### Bug Description

When trying to use a compiler-options.ts file when generating types the command is throwing as error.

### Steps to Reproduce

1. Create a compiler options file called `compiler-options.ts` in root.
2. Run generate types like so `storyblok types generate  --space [spaceId] --compiler-options ./compiler-options.ts`

### Expected Behavior

Types should be generated and compiler options should be applied

### Actual Behavior

An error is thrown `Unknown file extension ".ts" for [path]/compiler-options.ts`

### Code Sample

```javascript

```

### Environment

```shell
System:
    OS: macOS 15.6.1
    CPU: (14) arm64 Apple M4 Max
    Memory: 1.26 GB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.14.0 - ~/.volta/tools/image/node/22.14.0/bin/node
    npm: 10.9.0 - ~/.volta/tools/image/npm/10.9.0/bin/npm
  Browsers:
    Chrome: 140.0.7339.133
    Safari: 18.6
  npmPackages:
    @storyblok/react: ^5.0.0 => 5.0.0 
    next: ^15.5.2 => 15.5.2 
    react: ^19.1.0 => 19.1.0 
    storyblok: ^4.6.1 => 4.6.1
```

### Error Logs

```shell

```

### Additional Context

_No response_

MatheusMartinho:
Fixes #307.

## Before this PR

`storyblok types generate --compiler-options ./compiler-options.ts` fails with `Unknown file extension ".ts"`. The file was loaded with a bare dynamic `import()` in `loadCompilerOptions`, which only understands what the running Node version understands:

- `.ts` files throw `Unknown file extension ".ts"` before Node 22.18 (where type stripping became the default)
- `.json` files throw `needs an import attribute of "type: json"` on every Node version, so the `.json` format that worked in CLI v3 also has no working path in v4
- only `.js`/`.mjs` files load reliably, which matches what users report in #307

`loadCustomFieldsParser` uses the same bare `import()` for `--custom-fields-parser` and fails the same way, even though the option is documented with a `.ts` example path.

## After this PR

Both files are loaded through a small shared helper, `load-user-module.ts`, that evaluates the file with jiti and returns its default export. This is the same loading strategy the CLI already uses for `storyblok.config.ts` (via c12/jiti, see #693) and for schema entry files (`loadSchemaModule`). `.ts`, `.js`, `.mjs`, `.cjs`, and `.json` files now all load consistently on every supported Node version, and the helper honours the same `JITI_TSCONFIG_PATHS=false` escape hatch as the config loader.

## Tests

- `load-user-module.test.ts` follows the real-filesystem pattern from `lib/config/loader.test.ts`: it writes fixture files to a temp directory and asserts that compiler options load from `.ts` (the regression in #307), `.json`, `.mjs`, and `.cjs`, and that a custom fields parser function loads from `.ts`.
- `actions.test.ts` previously intercepted the dynamic import through mocked `pathe.resolve`/`pathToFileURL` and a `vi.mock("/mocked/path")` module stub; it now mocks the loader module directly, which also removes the Windows-specific `pathToFileURL` workaround.

## Verification

- `vp test run` in `packages/cli`: 92 test files, 1288 tests passing
- `pnpm nx lint storyblok`, `pnpm nx test:types storyblok`, and `pnpm format`: clean

cc @maoberlehner

### DIFF
--- a/packages/cli/src/commands/types/generate/actions.test.ts
+++ b/packages/cli/src/commands/types/generate/actions.test.ts
@@ -50,43 +50,33 @@ vi.mock("pathe", async (importOriginal) => {
   };
 });
 
-// Mock pathToFileURL so dynamic imports resolve consistently on all platforms.
-// On Windows, pathToFileURL adds a drive-letter prefix (file:///D:/…) which
-// prevents vi.mock('/mocked/path') from intercepting the import.
-vi.mock("node:url", async (importOriginal) => {
-  const actual = await importOriginal<typeof import("node:url")>();
+// Create a mock for the custom fields parser and the user module loader that
+// hands it to the code under test. vi.hoisted makes both available inside the
+// hoisted vi.mock factory below.
+const { mockCustomFieldsParser, mockLoadUserModule } = vi.hoisted(() => {
+  const customFieldsParser = vi.fn().mockImplementation((key, field) => {
+    if (field.field_type === "native-color-picker") {
+      return {
+        [key]: {
+          properties: {
+            color: { type: "string" },
+          },
+          required: ["color"],
+          type: "object",
+        },
+      };
+    }
+    return {};
+  });
   return {
-    ...actual,
-    pathToFileURL: (p: string) => ({ href: p }),
+    mockCustomFieldsParser: customFieldsParser,
+    mockLoadUserModule: vi.fn().mockResolvedValue(customFieldsParser),
   };
 });
 
-// Create a mock for the custom fields parser
-const mockCustomFieldsParser = vi.fn().mockImplementation((key, field) => {
-  if (field.field_type === "native-color-picker") {
-    return {
-      [key]: {
-        properties: {
-          color: { type: "string" },
-        },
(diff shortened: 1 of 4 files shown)


MatheusMartinho:
Gentle ping whenever someone has a moment. This fixes #307: `storyblok types generate --compiler-options ./compiler-options.ts` fails with `Unknown file extension ".ts"` on Node versions before 22.18, `.json` files fail on every version, and `--custom-fields-parser` has the same problem. The two required checks (`build (linux)` and `licenses`) are still waiting to run, I think because the workflows need a maintainer's approval for a fork. Happy to rebase or adjust anything.

PR #761, opened 2026-08-17
fix(capi-client,js-client): invalidate cached content by the version it belongs to

### LINKED ISSUE #105: Published responses cached regardless of caching settings

maoberlehner:
Supersedes #508, keeping @intellix's commit and their use case intact.

Docs counterpart: [storyblok/storyblok-docs-platform#627](https://github.com/storyblok/storyblok-docs-platform/pull/627). The caching page recommends the pattern this PR makes work, and describes it in a way that breaks once an access token has a TTL.

## The problem

Published content goes stale and never recovers. Invalidation is reactive: the client only learns a new `cv` from a response it receives, but published responses are served *from* the cache, so no request goes out and no new `cv` arrives. Closed loop. Reproduced against a live space: warm the cache, publish a story, re-read it, and the old content comes back with zero network calls, indefinitely.

`/cdn/spaces/me` is the way out: tiny response, `s-maxage=2` against `s-maxage=604800` on content endpoints, and [our own docs recommend it](https://www.storyblok.com/docs/concepts/caching#retrieve-the-latest-cache-version). But it reports the version under `space.version` with no `cv` field, and the invalidation code only read `cv` — so the one endpoint that could break the loop was the one both clients ignored.

## Why the obvious fix regresses

#508 did `response.data.cv ?? response.data.space?.version`. Those are **not** the same number.

When an access token has a Minimum Cache TTL configured (Space Settings → Access Tokens), content endpoints return a `cv` floored to a multiple of the TTL, while `cdn/spaces/me` keeps reporting the space's raw version. Measured live against a token with `min_cache: 3600`:

- `cdn/stories` returns `cv=1786950000`; `cdn/spaces/me` returns `space.version=1786950860`
- requesting stories with `cv=1786950860` gives **HTTP 301** back to `cv=1786950000`

End to end with `cache.clear: 'auto'`, alternating poll and stories, three rounds, no content changes:

| build | tracked cv | `flushCache()` | story cache |
|---|---|---|---|
| `main` | stable `1786950000` | 0 | works |
| #508 as-is | oscillates `…860` and `…000` | **5 in 6 requests** | destroyed |

The TTL defaults to `0`, where the two values coincide, so the common path is fine. But it is a self-serve field on the access token screen, so "nobody sets it" is not safe.

## What the API actually guarantees

Every rule below rests on behaviour verified against the live CDN, not on assumptions:

| | verified behaviour |
|---|---|
| 1 | `/cdn/spaces/me` reports `space.version`, no `cv`, `max-age=0, s-maxage=2`. |
| 2 | A published request **without** a `cv` is redirected (301) to the same URL carrying the current `cv`. One extra hop, never stale. |
| 3 | A published request with any `cv` the edge does not hold (`0`, stale, even a future one) gets that same redirect. |
| 4 | A published request with an old `cv` the edge **still holds** is served that old snapshot for up to a week. The only path to stale content, and narrower than "any old `cv`": a first request for a non-current `cv` misses and the origin redirects it, so only a `cv` already warm at that POP serves the old snapshot. |
| 5 | A response body's `cv` identifies the snapshot that was actually served. |
| 6 | Draft requests ignore `cv` entirely and are never redirected. A draft-only save moves `cv` and `space.version` together. |
| 7 | `/cdn/spaces/me/` is served identically and a `cv` on it is ignored, though the parameter fragments its edge cache. |
| 8 | Without a Minimum Cache TTL, `space.version` and `cv` report the same raw version. With one, the `cv` is floored permanently. They are different units. |
| 9 | `space.version` is monotonic at the origin, but a two-second per-POP cache means a poll can be answered with a lower one. |
| 10 | A publish takes 5 to 6.4 s to appear in `space.version`, well beyond its `s-maxage=2`. Polling notices a publish within seconds, not immediately. |
| 11 | `/cdn/links`, `/cdn/links/{uuid}` and `/cdn/tags` report no `cv` at all, with 13 links and 2 tags in the space so an empty collection cannot explain it, and `/cdn/stories` returning a `cv` for zero results as the control. `/cdn/datasources` reports one. |
| 12 | A draft request reports the same `cv` a published one does: the same raw version without a Minimum Cache TTL, the same floored version with one (`1787558400` for both, against `space.version=1787560240`). |
| 13 | A trailing slash is served identically on collection and member endpoints of both the Content Delivery and the Management API, not only on `/cdn/spaces/me`. |

Fact 2 shapes the design: the origin corrects a wrong or missing `cv` for the price of one redirect. Invalidation never has to be exactly right, only never wrong in the direction of fact 4.

## The model

**Published content is an immutable snapshot addressed by its `cv`.** An entry is valid exactly as long as the `cv` it was served under is still the current one. That is a property of the entry, not an event in time.

`@storyblok/api-client` implements this directly:

- **Entries are tagged.** `CacheEntry.cv` records the `cv` the response reported, or, for endpoints that report none (`/cdn/tags`, `/cdn/links`), the `cv` the request was issued under. Every response teaches a `cv` whether or not it is cached: a draft response is never stored, but it reports the same version a published one does (fact 12), so it invalidates published entries like any other signal — the only one an app that reads both but never polls has.
- **Two watermarks, one per signal.** `knownCv` from response bodies, `knownSpaceVersion` from `/cdn/spaces/me`. Both advance by monotonic max only, so a stale edge read never moves them (facts 5 and 9), and they are never compared to each other (fact 8) except on a first sighting, where there is no earlier space version to compare against. A third value, `highestCv`, records the highest `cv` the record has ever held and is never dropped: `knownCv` is reset by an invalidation, and that is exactly when the next response is the one that teaches the version everything afterwards is measured against, so the floor a stale edge read is recognised against has to outlive the reset. Never sent as a `cv`.
- **They live in the cache provider**, under the reserved key `sb:versions:v1:<tokenId>`, so they share fate with the entries they govern and every client sharing a provider shares them — which is what makes polling work for a per-request client in a serverless deployment. A missing record reads as "cv unknown" rather than falling back to TTL alone, since it can be evicted while the entries it governs survive.
- **Entry keys are scoped to the space.** The access token selects the space and travels in the `token` parameter rather than in the query, so without it two clients sharing one provider read each other's content. Keys carry a non-cryptographic hash of the token rather than the token itself, since they reach listings, `MONITOR` output and metrics labels.
- **A read is a hit when the entry is TTL-fresh and its tag equals `knownCv`.** A publish sets `knownCv` to undefined, which makes every tagged entry unreachable at once and sends the next request out bare, taking the redirect from fact 2. Nothing is flushed: unreachable entries expire by TTL and eviction, and flushing would empty a provider other clients keep their own entries in.
- **A response is discarded when the `cv` it was issued under is no longer the known one**, unless its own body reports the `cv` that superseded it. It was answered for a version since superseded, so it may neither be stored nor teach a `cv`. The check is against shared state, so it holds across clients and processes; the exception keeps a response that merely lost a race, since its body proves it carries the current snapshot. One case has no `cv` to compare — a request issued while none was known, racing an explicit `flushCache()` — so the record also carries a `generation` that `flushCache()` bumps, and a response whose generation no longer matches is discarded. That is the epoch, shared through the provider rather than held on the instance, and bumped only on an explicit flush, so noticing a publish still costs no refetch of what was in flight.
- **A caller-pinned `cv` is honoured literally.** Keyed by its own `cv`, immune to publishes, expiring by TTL alone, never teaching a watermark. The snapshot it receives is indistinguishable on the wire from a stale edge read (fact 4), so the stale-read rule is gated on caller intent rather than on the response. Only a positive `cv` pins: `0` is the sentinel `storyblok-js-client` records for "no version known", and reading it as a choice of snapshot keeps it on the wire and files the entry under a key no publish signal reaches.
- **Only a positive `cv` is a version.** `0` is rejected where the response body is read, not only where the watermark is written, which keeps "no numeric sentinel reaches the wire" true end to end (fact 3).

`flushCache()` keeps its role for webhook-driven invalidation under `cache.flush: 'manual'`, and now resets the watermark record along with the provider — keeping `highestCv`, a fact about the space rather than about any entry, and the bumped `generation`.

The reasoning is recorded in `adr/0014-content-cache-invalidation-by-version-watermarks.md`.

### What this removes

Modelling invalidation as an event — when a signal moves, flush — destroys evidence instead of recording it, so every edge case needs compensation and the compensation is where the defects live. Gone with it:

- the **cache epoch**, replaced by the issued-under check above,
- the **settle protocol** (pending revalidation, `cv` stripping, in-flight promise coordination, a substituted network-first strategy), replaced by `knownCv = undefined` plus the next natural request,
- **flushing from the correctness path** entirely.

Roughly 150 lines of concurrency machinery, and the defect class that came with it.

### One rejected alternative

Deciding from the `cv` a request was issued with, rather than from the one its body reports, looks like one removable cause. Implemented, it breaks a real case: under a Minimum Cache TTL the `cv` is floored and legitimately stops moving while `space.version` advances, so recording supersession on the `cv` scale from a `space.version` signal mixes the two units fact 8 says are incomparable, and content is then never cached again. A late pre-publish response and a fresh floored refetch report the same `cv`; only issue time separates them. The issued-under check stays, narrowed to the cases above.

### `storyblok-js-client`

It keeps its flush-based mechanism: it is widely deployed and its custom cache providers observe its key shapes. It adopts the same semantics, nine fixes:

- **A falsy tracked `cv` is never serialized.** The sentinel `0` is not a version the API knows (fact 3), so `cv=0` cost a redirect and cached the response under a key no later request read, leaving freshly published content invisible for the whole cache lifetime. `clearCacheVersion()` still records `0` internally, so the public API is unchanged.
- **Cache keys are canonical**, built from params sorted at every depth, because the serializer walks in insertion order: `parseParams` assigns the `cv` before `resolve_level`, so a rebuilt params object appending the `cv` last describes one request under two keys. Only cache keys are sorted; query strings keep the caller's order.
- **The response is cached after any flush it triggered, under the key the next read will build.** It used to be stored first and then dropped by its own flush. Ordering alone was not enough: the tracked `cv` is part of the key, so a response reporting a new one landed under the pre-flush `cv` while the next read looked under the new one.
- **A caller-pinned `cv` is outside the version tracking entirely** — it neither teaches a `cv` nor flushes, and it is not stale for reporting the version it was pinned to. Judged by the tracked `cv` instead, the one request that asked for an older snapshot was the one never cached: measured against a real space, the edge served that snapshot and every read of it went back to the network.
- **The `cv` in that key is only rewritten where the client chose it.** A caller-pinned `cv` belongs to the caller's key, and `cache.cv: 'manual'` puts no `cv` in a key at all — writing the tracked one into either takes those requests out of the cache entirely, the opposite of what the mode exists for.
- **A `cv` is never adopted backwards**, and such a response is not cached: an older edge object answering (fact 5) would pin every later request to content the space has moved past. Floored at the highest `cv` the API ever reported rather than at the tracked one, which a flush zeroes — the response after a flush is the one that teaches the version everything afterwards is measured against, so an unfloored comparison there re-adopts the very `cv` the flush dropped. `setCacheVersion()` does not raise that floor: only a version the API reported is evidence of what it served, and nothing lowers the floor, so a caller value too far ahead would make every later response look stale for good.
- **A retry after a 429 re-reads the `cv` before it goes out**, and **`flushCache()` moves the epoch before it empties the provider**. Both are the same hazard: a response that outlives a flush re-sending the `cv` that flush dropped, or writing itself into the cache mid-clear.
- **Both space-version comparisons use `>`.** Inequality flushed for a version from a POP that had not caught up (fact 9).
- **The recorded space version is a maximum too**, so after a regression the next poll back at the current version does not flush again.
- **Neither `get()` nor `getAll()` mutates the caller's params.** They stamped the version, token and `cv` onto their object, carrying request state into their next call.
- **A `cv` counts as pinned by the caller by truthiness**, which is the test `parseParams` applies before substituting the tracked one. Read as `!== undefined`, a `cv: 0` was treated as a choice of snapshot while the request carried the tracked `cv`, and the entry landed under a key holding the zero.
- **The signal is scoped to the cache it would act on**, identified by the provider object rather than the client instance, so instances writing to the same cache share it while an instance with its own provider keeps its own. Keyed per instance, per-request clients over one external provider each saw a fresh state and, under a Minimum Cache TTL, each flushed the shared cache on its first poll. The baseline a first sighting compares against is a separate, never-cleared map of the highest `cv` seen, because a flush zeroes the tracked `cv` for the whole token and would otherwise erase the comparison for every cache that had not polled yet. The flush epoch lives in that same per-cache record, so one instance's `flushCache()` no longer discards in-flight responses belonging to unrelated tokens.

Both clients also normalise trailing slashes in the request path (fact 7). A spelling that fails the path comparison leaves the poll carrying a `cv`, its response cached, and the space version never read; `get('/cdn/spaces/me')` used to build `//cdn/spaces/me` and do exactly that.

One unrelated bug is fixed here because it sits two lines from the subject: a request retried after a 429 re-entered `cacheResponse` without the caller's `fetchOptions`, so the second attempt went out with different credentials, headers, signal and caching hints than the first.

## Behaviour changes

In `@storyblok/api-client` (pre-1.0):

1. **`cache.flush: 'auto'` no longer calls `provider.flush()`.** Entries that can no longer be served expire by TTL or eviction, bounded by the existing 1000-entry cap. The trade is not emptying a shared provider on someone else's behalf.
2. **An invalidated entry is not served as a `network-first` fallback.** This matches `main`, where a flush deleted those entries outright; it was the settle machinery that had briefly added such a fallback. A publish followed by a failing origin therefore surfaces the error rather than stale content.
3. **`CacheEntry` gains an optional `cv`, and entry keys change shape.** Custom providers store entries opaquely and need no change, but they must not write to the reserved key. An external provider carried across the upgrade refills once. The `cv` is load-bearing, though: a provider that rebuilds entries field by field and drops it degrades invalidation to TTL alone, and a dropped tag reads back exactly like an entry from an endpoint that reports no `cv`. `CacheEntry`/`CacheEntryInput` are exported so an adapter can name the shape it has to round-trip, and the contract says so — a runtime warning was tried and removed: it cannot fire in the per-request-client deployment it exists for, and false-positives on a shared provider where another client legitimately stores an entry untagged.

In `storyblok-js-client`, cache keys change shape too — sorting the params changes every key — so a persisted custom provider (Redis, say) takes one cold-cache round after deploy. The in-memory cache is unaffected, since it starts empty. No public API changes.

## Known issues and deliberate choices

Each of these is a trade-off taken knowingly, not an open defect.

- **A Minimum Cache TTL bounds freshness regardless of the client.** The floored `cv` does not move on publish, so the edge object stays a HIT with a growing `age` under `s-maxage=604800`: with `min_cache: 3600`, five reads over 12 s after a publish all returned the pre-publish story. No client-side model can fix that, this one included. The TTL is the freshness contract; these rules only make sure the client adds no staleness on top of it.
- **Every draft request reads the watermark record once**, where `main` touched the provider not at all. The price of taking the publish signal from draft traffic; it writes only when a version actually moved.
- **A space with a Minimum Cache TTL pays one needless revalidation per watermark record**, from the first-sighting comparison against the `cv`. It answers itself and does not recur.
- **In `storyblok-js-client` that first-sighting state is per process**, since it lives beside the client rather than in the provider, and a flush is its only reaction. Under a Minimum Cache TTL, where the floored `cv` never equals the raw space version, every cold start therefore flushes the shared cache once — a module-scope provider does not help, because a new process brings a new module scope. The price of leaving that client on the flush model.
- **A response issued while no `cv` was known cannot be recognised as superseded by a publish alone.** It has no issue-time `cv` to compare, and the generation only moves on an explicit `flushCache()`. An unlucky one carries pre-publish content into the cache for one TTL.
- **One extra provider `get` per cacheable request** for the watermark record, issued in parallel with the entry read. Free for the in-memory provider, one round trip for an external one, still far cheaper than the origin fetch it prevents. Non-cacheable requests do not pay it: only the poll has anything to record, and a draft response may neither teach a `cv` nor be cached.
- **An endpoint reporting no `cv` is only invalidated when it was fetched while a `cv` was known.** Fetched during an unknown-`cv` window, it falls back to TTL alone.
- **The watermark record is read, merged and written back without a lock**, and `CacheProvider` offers no compare-and-swap to build one from. The write re-reads the generation first and gives up when it moved, so a `flushCache()` is never merged away, which narrows the window to the provider write itself. Two concurrent requests can write over each other: a poll that dropped `knownCv` for a publish can lose that drop to a content response that read the record before it, and the space-version watermark falls back with it. It costs one repeated round of invalidation — the next poll recognises the same publish again — and never serves an entry past its version, since the `cv` the losing write records is the newer one. The window is a provider round trip wide.
- **The js-client signal is keyed by the provider object**, so it is shared only while that object outlives the client. A handler building its provider inline per request gets a fresh state each time and never compares one space version against a previous one — the serverless case the design exists for. Build the provider once at module scope. Documented rather than changed, since moving that state into the provider is what this PR deliberately does not do to js-client.
- **`clear: 'onpreview'` only clears on draft requests.** What matters is the version of the poll, not the endpoint: a published poll never flushes, a draft poll of `/cdn/spaces/me` does and the published cache recovers normally. Documented semantics, left as-is, documented on `ISbCache.clear` and locked with a test for each half.
- **js-client clients without a cache share one flush epoch.** There are no entries to keep apart, so the shared state guards a cache holding nothing. One such client's `flushCache()` can stop another's in-flight response from teaching a `cv`, worth one redirect on that request.
- **js-client multi-token invalidation stays asymmetric.** A clear empties the whole cache but drops only the signalling token's `cv`, so every other token refills from whatever the edge still holds. Pre-dates this PR; documented on `ISbCache.clear`, since fixing it means bulk-zeroing a module-level map shared across instances.
- **js-client keeps flushing rather than tagging entries.** The tagged model would work there too, but its cache keys are observed by user-supplied providers, so it is a separate, breaking-ish change.

## Verification

```
api-client: 219 passed (17 files)
js-client:  237 passed, 1 skipped (8 files)
lint, types, build: clean
```

Behaviour is asserted through the public entry points and msw, on request counts and emitted query strings rather than on internal state — which is how several earlier defects hid behind green tests.

api-client covers the publish path, every no-op path (unchanged, regressed, floored `cv`, non-numeric version, a `space.version` embedded in another endpoint's body, `flush: 'manual'`), pinned-`cv` isolation, a failed refetch, trailing-slash spellings, and the two capabilities the previous model could not have had: three sequential per-request clients over one shared provider settling at one refetch total, and client B's in-flight pre-publish response not being served after client A noticed the publish. A provider that rebuilds entries and loses the `cv` is asserted to warn exactly once across three reads, and one that round-trips them not at all.

js-client asserts that a second identical published request reaches the network once for each of the six shapes a `cv`-bearing cache key can take — plain, `resolve_relations`, `cache.cv: 'manual'`, a caller-pinned `cv`, a falsy `cv`, and a read whose own response flushed the cache on its way in. Each fails with its fix reverted.

Live verification against a dedicated space, both clients built and driven through the CDN, produced the facts table above, using a purpose-made token with `min_cache: 3600` for the flooring scenario. Every rule stated here was confirmed end to end, including recovery after publish, which never happens on `main`.

## Relation to #773

Independent, and they compose: #773 fixes error-result handling in the cache strategies, this PR fixes which version a cached entry belongs to. No defect overlaps.

They touch the same two files, so applying #773 on top of this branch produces exactly one conflict, the strategy call site, resolved by adding `getFailure` to the argument object. Verified locally: the merged tree is green. **Land #773 first**, then rebase this branch onto it; the reverse order forces #773 to rebase onto a `client.ts` that has changed substantially around its hunks. Together: an entry that is still valid falls back on a 503 as #773 intends, an entry invalidated by a publish surfaces the error per behaviour change 2.

## Follow-ups

- **API:** have `cdn/spaces/me` report the same cache version the content endpoints return, so the value it exposes is directly usable as a `cv`. Both clients' existing `cv` paths would then work with no client code at all.
- **Docs:** [storyblok/storyblok-docs-platform#627](https://github.com/storyblok/storyblok-docs-platform/pull/627). The caching page states "the `version` property in the response is the latest `cv` value" and separately documents TTL, without noting that the two stop matching once a TTL is set. Worth merging alongside this, together with the `cache.clear: 'auto'` requirement, the api-client behaviour changes, the 5 to 6.4 s a publish takes to reach `space.version`, and the fact that a Minimum Cache TTL caps freshness on its own.

Fixes #105

### DIFF
--- a/packages/capi-client/src/resources/spaces.test.ts
+++ b/packages/capi-client/src/resources/spaces.test.ts
@@ -2,6 +2,7 @@ import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from "vitest
 import { setupServer } from "msw/node";
 import { http, HttpResponse } from "msw";
 import { createApiClient } from "../index";
+import type { CacheEntry, CacheEntryInput, CacheProvider } from "../utils/cache";
 
 const server = setupServer();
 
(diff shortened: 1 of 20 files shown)


alexjoverm (changes requested):
Great one Markus! Pasting here some AI findings:

### Major: `@storyblok/api-client` keeps the stale `cv` after a space-version flush

**File:** `packages/capi-client/src/client.ts:309-320`

**Problem:** `updateSpaceVersion` calls `cacheProvider.flush()` directly and leaves `currentCv` untouched. The js-client path deliberately does the opposite: `flushCache()` calls `clearCacheVersion()`, and the PR description justifies it explicitly — *"clearing is necessary — reusing the old `cv` would hit the CDN's cached old response."*

Verified with a probe against the PR branch (msw, inspecting the outgoing query):

```
cv values sent: [ null, '1000', '1000' ]
//                          ^ after the space-version-change flush, the refetch still carries cv=1000
```

`createCacheKey(method, path, rawQuery)` builds the local key from the query **before** cv is applied, so the response fetched under the stale cv is re-cached under the same key. The next poll sees an unchanged `space.version` and does not flush again, so the client stays stale until the next publish. That is exactly the closed loop #105 describes, still open for `@storyblok/api-client` whenever the edge still holds the object for `?cv=<old>` (`s-maxage=604800`, and active purge-on-publish is not yet enabled for every space).

The two clients cannot both be right. Either reusing the old cv is safe (then js-client's `clearCacheVersion()` and its documented 301 cost are unnecessary) or it is not (then api-client is under-fixed).

**Suggestion:** reset the tracked cv in the space-version flush path so the next request omits `cv` and gets the origin `301` to the current one:

```ts
if (cacheFlush === "auto" && (isFirstSighting || hasSpaceVersionChanged)) {
  await cacheProvider.flush();
  currentCv = undefined; // force the next published request to re-resolve the cv
}
```

(`applyCvToQuery` is skipped when `currentCv === undefined`, so the request goes out without `cv` and the API redirects to the current one — same effect as js-client's `cv=0`, without inventing a sentinel.) Add a regression test asserting the outgoing `cv` param is absent on the request that follows a space-version flush. If you keep the current behavior, please say which live observation supports it and add a comment, because it reads as an oversight next to the js-client side.

### Minor: the first-sighting flush is avoidable for the default (TTL = 0) token

**File:** `packages/js-client/src/index.ts:796-802`, `packages/capi-client/src/client.ts:305-308`

**Problem:** the PR accepts "one wasted flush + one 301 per process" to cover the publish that may have landed between the first content request and the first poll. For a token with no Minimum Cache TTL — the default — that cost is unnecessary: the returned `cv` is not floored, so a content response's `cv` and a poll's `space.version` are literally the same number.

**Suggestion:** on first sighting, compare the two. Equal means nothing was published in between, so skip the flush. Unequal means either a publish or a TTL floor, so flush once as today. This makes the heuristic exact for the majority of tokens and keeps the TTL case safe. Both regression tests already in the PR still pass under that rule (the TTL test never has an equal pair).

### Minor: js-client does not type-guard `space.version`

**File:** `packages/js-client/src/index.ts:792, 794`

**Problem:** `response.data` is `any` (`ISbResult.data: any`), so `response.data.space?.version` is untyped and lands in `spaceVersions`, which is declared as `CachedVersions` (`[key: string]: number`). `extractSpaceVersion` in the api-client guards with `typeof space.version === "number"`; the js-client does not. Probe: a string `"1786950860"` followed by a numeric `1786950860` triggers a spurious flush. Separately, `if (params.token && spaceVersion)` uses truthiness (so `0` is treated as absent) while the two comparisons immediately below correctly use `!== undefined`.

**Suggestion:**

```ts
const rawSpaceVersion = response.data.space?.version;
const spaceVersion = typeof rawSpaceVersion === "number" ? rawSpaceVersion : undefined;

if (params.token && spaceVersion !== undefined) { ... }
```

### Minor: js-client reacts to `space.version` on any URL

**File:** `packages/js-client/src/index.ts:792`

**Problem:** the new block is shape-based, while the surrounding cache code already special-cases the literal `"/cdn/spaces/me"` twice (lines 715 and 764). Not a live bug — `space.version` is only ever exposed by the public CDN space payload, never by the MAPI space responses — but the guard is looser than it needs to be, and the magic string is now effectively repeated three times.

**Suggestion:** extract `const SPACES_ME_PATH = "/cdn/spaces/me";` and gate all three sites on it.

### Minor: api-client hides the space-version tracking inside `updateCv`

**File:** `packages/capi-client/src/client.ts:322-326`

**Problem:** `updateSpaceVersion` only runs on the `nextCv === undefined` early-return branch of `updateCv`. It works today because `/cdn/spaces/me` has no `cv`, but the dependency is implicit and buried in a function named for the cv, and it silently becomes dead if the API follow-up in the description lands (spaces/me reporting a real `cv`). It also makes the divergence with the js-client implementation harder to spot.

**Suggestion:** call both trackers from the two call sites (`requestWithCache`'s non-cached path and `cacheSuccessResult`), or rename to something like `trackResponseVersions(result)` that internally does cv-then-space. Add a cross-reference comment in each package pointing at the other, since the same heuristic now exists twice with different guards.

### Minor: coverage gaps and parity between the two test suites

- api-client has a `flush: "manual"` test; js-client has no `clear: "manual"` equivalent. Probe also shows `clear: "onpreview"` + published polling **never** flushes (`isCacheClearable` requires `params.version === "draft"`). That means the documented polling pattern is a no-op for `onpreview` users with a public token — worth locking with a test and mentioning in the docs PR.
- Neither package covers the "poll before any content request" ordering, which is the recommended startup shape. Probe confirms api-client correctly does not flush there; it should be locked.
- No test asserts what happens to the tracked `cv` around a space-version flush (see the Major).
- `spaceVersions` is a module-level global with no reset hook and no accessor (`cacheVersions` at least has the public `cacheVersions()` method). The new tests stay isolated only because each uses a unique token string. Worth a comment on the declaration or the describe block so the next test author does not trip on it.

alexjoverm (changes requested):
@maoberlehner few extra major findings

### Major: a failed revalidation silently discards the ambiguous signal for good

**File:** `packages/capi-client/src/client.ts:505-519`

`pendingRevalidation` is cleared before `fetchFn` resolves, and `currentSpaceVersion` was already advanced in `updateSpaceVersion`. If the revalidation request fails, nothing retries it and no future poll can detect the change.

```ts
if (pendingRevalidation !== undefined) {
  const { cvBefore } = pendingRevalidation;
  pendingRevalidation = undefined;      // consumed before we know it succeeded

  const result = await fetchFn(query);  // may reject or return result.error
```

Verified: warm the cache, publish, poll (ambiguous sighting), then make the revalidation return 500. The stale entry is served on every subsequent read with zero further network calls:

```
REVAL status: 500  reqs: [null,null]
AFTER body: {"a":"old"}  reqs: [null,null]
AGAIN body: {"a":"old"}  reqs: [null,null]
```

### Major: the revalidation branch bypasses the configured cache strategy

**File:** `packages/capi-client/src/client.ts:505-519`

The branch calls `fetchFn` directly instead of going through `strategy({ key, cachedResult, loadNetwork })`. Two user-visible consequences on that one request:

- `network-first` loses its documented fallback to the cached entry on a network error. Verified against a baseline: the same failure resolves to the cached value normally, but throws when it lands on the revalidation.
- `swr` becomes blocking for that request instead of serving stale and revalidating in the background.

```
BASE (network-first, no poll):  resolved -> {"a":"old"}
NF   (network-first, on reval): THREW -> TypeError Failed to fetch
SWR  (on reval):                THREW -> TypeError Failed to fetch
```

The blocking behavior is arguably intended (the comment says the request "bypasses the cache"), but losing the `network-first` fallback is not, and neither is documented.

### Major: js-client — a leading slash silently disables the whole feature

**File:** `packages/js-client/src/index.ts:57, 197, 732, 781, 813`

`get()` builds ``const url = `/${slug}` ``, so `get('/cdn/spaces/me')` produces `//cdn/spaces/me`, which fails every `url === SPACES_ME_PATH` equality check. `isCDNUrl` only tests `url.includes('/cdn/')`, so it still takes the CDN path. Three things go wrong at once, all silently:

```
A  get('cdn/spaces/me'):   calls=[/cdn/stories/home, /cdn/spaces/me, /cdn/spaces/me]  cvs={tok-a: 0}
B  get('/cdn/spaces/me'):  calls=[/cdn/stories/home, //cdn/spaces/me?cv=100]          cvs={tok-b: 100}
```

In B the poll carries a `cv` (fragmenting the edge cache of the endpoint this PR makes load-bearing), the response is written into the published memory cache, so the second poll never leaves the process, and `space.version` is never read, so no flush ever happens. The user sees a polling loop that appears to work and invalidates nothing, forever.

Needs path normalization

packages/capi-client/src/resources/spaces.test.ts
```diff
@@ -129,3 +129,606 @@ describe("spaces.get()", () => {
...
+    expect(linkRequests).toBe(2);
+
+    // The first entry must still be cached: nothing reported a new cv, and a content
+    // response's `space.version` is not a flush signal.
+    await client.get("v2/cdn/links", { query: { version: "published" } });
+
+    expect(linkRequests).toBe(2);
+  });
+
+  it("should not attach a cv to the poll request", async () => {
```
dipankarmaikap: This is the same test as in line 211; we can remove this.

packages/capi-client/src/resources/spaces.test.ts
```diff
@@ -129,3 +129,606 @@ describe("spaces.get()", () => {
...
+      });
+      await client.get("v2/cdn/links", { query: { version: "published" } });
+      await client.spaces.get();
+    }
+
+    expect(linkRequests).toBe(1);
+    expect(store.size).toBe(1);
+  });
+
+  it("should ignore a space.version reported by another endpoint", async () => {
```
dipankarmaikap: This is the same test as in line 179; we can remove this.

packages/capi-client/src/resources/spaces.test.ts
```diff
@@ -129,3 +129,606 @@ describe("spaces.get()", () => {
...
+    const client = createApiClient({ accessToken: "test-token" });
+
+    await client.get("v2/cdn/links", { query: { version: "published" } });
+    await client.spaces.get();
+
+    expect(spaceUrls).toHaveLength(1);
+    expect(new URL(spaceUrls[0]).searchParams.has("cv")).toBe(false);
+  });
+
+  it("should flush on the first poll when the cv does not match the space version", async () => {
```
dipankarmaikap: This is the same test as the one on line 235, so we can remove this as well. Also, this should be `revalidates` rather than `flush`, but I think we can safely remove it altogether.

dipankarmaikap (approved):
Approving this PR. I left three minor comments, all related to tests that can simply be removed. Once those are addressed, this is ready to be merged.

alexjoverm (changes requested):
### Major: concurrent requests each settle the same pending revalidation

**File:** `packages/capi-client/src/client.ts:504-543`

`pendingRevalidation` is read at `loadNetwork` entry (`:506`) and only cleared once a response is in hand (`:520`) — deliberately, to fix the "failed revalidation drops the signal" finding. But nothing marks the settle as *in flight*, and the strategy is selected per call (`:540`). Every cacheable request started inside that window therefore:

1. uses `revalidationStrategy` (network-first) and so bypasses its cache hit, and
2. captures the same `revalidation` record and runs the flush check itself.

With N parallel published reads after an ambiguous sighting (a normal SSR page firing `Promise.all([...])`) that is N origin requests instead of 1, and — when the `cv` moved — up to N flushes interleaved with N `cacheProvider.set` calls, so entries stored by earlier siblings are wiped by later ones and get re-fetched on the next read. No stale data is served, but the PR's central cost argument ("Same one-extra-request cost as a flush, scoped to the instance that made it") only holds for strictly sequential reads.

**Suggestion:** keep the record but add an in-flight marker, e.g. `pendingRevalidation: { cvBefore: number; inFlight?: Promise<void> }`. The first request sets `inFlight` before awaiting and clears it on failure; concurrent requests see `inFlight` and use the configured `strategy` (optionally awaiting it first so they observe a post-flush cache). Add a test firing `Promise.all` of 4–8 reads after an ambiguous poll and asserting one origin request and one flush.

### Minor: duplicated and stale tests in `spaces.test.ts` (rebase artifact)

**File:** `packages/capi-client/src/resources/spaces.test.ts:179/379, 211/411, 235/435`

Three redundant tests, ~70 lines:

- `179-210` and `379-410` — "should ignore a space.version reported by another endpoint": byte-identical (verified by hash).
- `211-234` and `411-434` — "should not attach a cv to the poll request": byte-identical.
- `235` "should **revalidate** on the first poll when the cv does not match the space version" vs `435` "should **flush** on the first poll…". The second is a leftover from the pre-revalidation revision; its comment ("a cv that no longer matches the space version gives it away, so flush once defensively") describes behavior the api-client no longer implements. Its assertion (`linkRequests === 2`) passes under either rule, so it is a test that cannot fail for the reason it documents — the worst kind to leave behind.

Vitest happily runs duplicate test names, so CI cannot catch this. Delete `379-434` and `435-459`.

### Moderate: js-client — the flush drops the `cv` for the wrong token

**File:** `packages/js-client/src/index.ts:838-851` with `:906-910, 954-958`

The space-version signal is keyed by `params.token` (a documented per-request override), but the flush it triggers is `this.flushCache()` → `clearCacheVersion()` → `cacheVersions[this.accessToken] = 0`. When a request supplies its own `token` — which every new test in this PR does — the flush empties the shared memory cache but leaves `cacheVersions[params.token]` at the old value, so the very next request still goes out with `?cv=<old>`. That is the failure mode the drop exists to prevent, and the PR's own test acknowledges it in a comment ("The token is the client's own access token so that `flushCache` clears the cv this test tracks").

The pre-existing `cv` path (`:855-862`) has the same asymmetry, so this can be a scoped follow-up — but as written the new feature silently under-delivers for any process using more than one token, which is the normal shape for preview + published side by side.

**Suggestion:** clear the cv for the token that produced the signal (`cacheVersions[params.token] = 0`) in addition to `flushCache()`, or document the limitation on `ISbCache.clear` next to the polling note.

PR #763, opened 2026-08-17
fix(nuxt): make the bridge option optional in useAsyncStoryblok

maoberlehner:
Supersedes #231, which had the right premise but is 401 commits behind `main` and conflicts in every file it touches (the Nuxt 4 playground restructure, the Oxfmt reformat, and the `src/types.ts` addition all landed since).

## What

`useAsyncStoryblok` already defaults `bridge` to `{}` at runtime and inherits `resolve_relations` / `resolve_links` from the `api` options:

```ts
const { api, bridge = {}, ...rest } = options;
```

…but `UseAsyncStoryblokOptions` still declared `bridge` as required, so callers who don't configure the bridge were forced to write `bridge: {}` for nothing. Made it optional.

Also in this PR:

- **Removed `src/types/index.ts`.** `module.ts` does `export * from "./types"`, which resolves to `src/types.ts` — the directory was unreachable dead code, and its copy of `UseAsyncStoryblokOptions` had already drifted (`api: ISbStoriesParams` instead of `MaybeRefOrGetter<ISbStoriesParams>`). This is why #231's "single source of truth" import no longer compiles on `main`.
- **Dropped `bridge: {}`** from the three playground pages that carried it and from the reactive-params JSDoc example. `playground/app/pages/index.vue` keeps a populated `bridge` so the explicit path stays exercised.
- **Added `src/type-tests.test-d.ts`**, a compile-time guard run by `test:types`. Verified it fails when `bridge?:` is reverted to `bridge:`.

Widening required → optional is backward compatible; existing `bridge: {}` call sites keep working.

Closes #231


packages/nuxt/src/type-tests.test-d.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Type-level regression guards, checked by `pnpm test:types`.
+ *
+ * These are compile-time only; nothing here is bundled into `dist`.
+ */
+import type { UseAsyncStoryblokOptions } from "./runtime/composables/useAsyncStoryblok";
+
+/** `bridge` is optional: the composable defaults it to `{}` at runtime. */
+export const bridgeIsOptional: UseAsyncStoryblokOptions = {
```
alexjoverm, line 9: question (non-blocking): I didn't know about this kind of types testing. Is it worthy? On which cases in particular?

(no need to answer here - just triggering the question so you can show us around in the weekly ;))

alexjoverm (approved):

packages/nuxt/src/type-tests.test-d.ts
JakubAndrzejewskiSB: I myself wasn't used to these kind of tests. I believe the TS will inform us instantly if we make some mistake based on the types.

This can be added but it will be probably rewritten quite a lot once I will add the proper Test support next week :)

JakubAndrzejewskiSB (approved):
Nice work!

packages/nuxt/src/type-tests.test-d.ts
maoberlehner: The usual approach is to do https://vitest.dev/guide/testing-types.html but this was the next best thing with no Vitest (yet). But as soon as Vitest is in, we should use `expectTypeOf` as we do in other packages already.

PR #764, opened 2026-08-18
test(repo): add Visual Editor manual QA tooling

maoberlehner:
Manual QA tooling for the Storyblok Visual Editor, for `@storyblok/nuxt` and `@storyblok/astro`.

The Visual Editor embeds the app in an iframe and pushes live content updates into it over the
bridge. Nothing but the real editor exercises that chain, and **its failure mode is silence**: a
broken bridge throws nothing, the page renders the story it fetched and then never changes. So every
assertion here asserts an observed change, never the absence of an error.

The reusable Playwright harness lives in a new `tools/visual-editor-qa` package
(`@storyblok/visual-editor-qa`), not copied into each framework package, so `@storyblok/astro` already
reuses the same harness `@storyblok/nuxt` does. Each package keeps only its own `qa.config.ts` values
and the specs that need the real editor. The `qa-engineer-manual` skill owns the checklist and the
session/space scripts every package shares.

## The skill

- **`visual-editor-checklist.md`**, plus a gate in `SKILL.md` that loads it. Framework-agnostic: it
  checks for the `data-blok-c` / `data-blok-uid` attributes every SDK wrapper emits, and it owns the
  per-run setup and the teardown that every package shares.
- **`scripts/save-storyblok-session.mjs`** opens a headed browser so a person logs in once, 2FA and
  all, and saves the session to `.storyblok-qa/session.json` at the repo root. It watches the browser
  rather than the terminal, so it needs no TTY and an agent can start it while a person only logs in.
  No credential reaches the harness, only the session state, which is gitignored. One saved session
  serves every package harness and every throwaway script.
- **`examples/visual-editor-run.mjs`** is a self-contained one-off check: copy it to `.claude/tmp/`,
  edit the constants at the top, run it, throw it away. It covers the standalone load, the editor
  render, and a live update, and prints the standard `{ outcome, ... }` JSON. `@playwright/test` is
  now a root devDependency so a script in `.claude/tmp/` can import it.
- **`scripts/configure-space.sh`** reads and sets a space's preview domain. Read-only unless you pass
  `--confirm`, and it refuses a non-https domain.
- **`scripts/seed-scenario.sh`** now refuses to seed a component missing `component_group_uuid`. The
  CLI otherwise skips it silently: the push reports success, the component never lands, and the
  stories that use it render nothing.

## The shared harness

- **`tools/visual-editor-qa`** holds the page object owning every Storyblok-app selector, config
  helpers, and preflight/playwright building blocks. Framework-agnostic, source-only, consumed
  directly by both packages' `test/visual-editor` directories.

## The packages

- **Seed scenario** for each package under `test/scenarios/has-playground-content/`, mirroring the
  content tree its own playground expects. The relation field carries the local story UUIDs;
  `storyblok stories push` remaps them to the remote ones, so no post-seed linking step is needed.
- **Playwright harness** in each package's `test/visual-editor/`: config, an auth setup, preflights,
  and its own `specs/live-editing.spec.ts`. `@storyblok/nuxt` has the three checks only the real
  editor can make; `@storyblok/astro` adds nested-block, relation, save/publish, and
  preventDefault/disabled-preview coverage on top. Everything the playground renders standalone
  belongs in the existing Cypress suite, not here.
- **`test/GUIDE.md`** in each package with its port, scenario, and quirks.
- Two Nuxt playground components, `PopularArticles.vue` and `Article.vue`. Both are required for
  observability rather than decoration: `StoryblokComponent` renders nothing for a Storyblok
  component with no Vue counterpart, so a resolved relation and an article story were invisible to
  any assertion.

## How to run it

```bash
node .agents/skills/qa-engineer-manual/scripts/save-storyblok-session.mjs   # once; log in, it saves by itself

set -a && source ./.env.qa-engineer-manual && set +a
bash .agents/skills/qa-engineer-manual/scripts/configure-space.sh --domain https://localhost:3200/ --confirm
bash .agents/skills/qa-engineer-manual/scripts/seed-scenario.sh \
  --scenario has-playground-content --scenario-dir packages/nuxt/test/scenarios

export NUXT_PUBLIC_STORYBLOK_ACCESS_TOKEN="$STORYBLOK_PREVIEW_TOKEN"
pnpm --filter @storyblok/nuxt qa:editor
```

Same flow for `@storyblok/astro` (`STORYBLOK_ACCESS_TOKEN`, port `4321`, `qa:editor`/`qa:stop`); see
its `test/GUIDE.md` for the daemon quirks specific to `astro dev`.

Restore the space's preview domain afterwards. Nothing does it for you, and the next person to open
the space in the editor gets a blank preview with no error.

## Things worth knowing

- **A blank preview frame is usually Chrome, not the bridge.** The editor is served from a public
  origin and embeds the playground from localhost, which Chrome's Local Network Access policy blocks
  with `net::ERR_BLOCKED_BY_LOCAL_NETWORK_ACCESS_CHECKS`, rendering a `chrome-error` page that looks
  exactly like a dead bridge. The harness disables the checks via Chromium launch args.
- **No locally-trusted certificate is needed.** `ignoreHTTPSErrors` covers the preview iframe as well
  as the page, so `qa:dev` serves nuxi's own self-signed cert and mkcert, `qa:certs`, and
  `NODE_EXTRA_CA_CERTS` are gone. Verified by a full green run without any of them.
- **The editor previews a story at its `full_slug`**, so it loads `/vue/articles/first-article`, which
  the catch-all route serves, not the playground's own `/articles/:slug`. Specs target what the
  editor actually loads.
- **Block selection is retried, not slept on.** The bridge's click handler is not ready the instant
  the preview paints, so the first click is sometimes swallowed. `selectBlock` asserts the form's
  field id changed, which is the only evidence the form switched; a swallowed click used to edit the
  story-level field silently and time out 30s later looking like a dead bridge.

## Not in this PR

- Region support is partial: `mapiBaseUrl`/`STORYBLOK_MAPI_URL` covers the harness, but
  `cleanup-remote.sh`, `list.sh`, and the skill's own docs still hardcode the EU host.
- Live editing of a richtext field is not covered. Driving the editor's contenteditable is brittle,
  and the editor's preview path for that story renders `body`, not `richText`. Noted in the GUIDE.

Fixes DX-551


dipankarmaikap (approved):
Looks good to me. One thought: the Astro example has a hard-coded preview token. We should remove that and perhaps add a small script under `tools/visual-editor-qa` to fetch `space.first_token` automatically.

PR #765, opened 2026-08-18
fix(repo): publish workspace dependencies as caret ranges

maoberlehner:
`pnpm publish` replaces the `workspace:` protocol at pack time, and `workspace:*` becomes an **exact** version. nx never rewrites it in the source manifest — `preserveLocalDependencyProtocols` defaults to `true` ([`config.js:161`](https://github.com/nrwl/nx/blob/master/packages/nx/src/command-line/release/config/config.ts)), and `nx release publish` delegates to `pnpm publish` for exactly this reason ([`release-publish.impl.js:43`](https://github.com/nrwl/nx/blob/master/packages/js/src/executors/release-publish/release-publish.impl.ts): *"pnpm publish provides ideal support"*).

So every published package pins its Storyblok dependencies to whatever the workspace happened to be **at its own publish moment**, and that snapshot is frozen forever.

`nx.json` sets `updateDependents: "never"`, so releasing a dependency does not bump its dependents. The two combine: the pin never refreshes on its own, and consumers cannot receive a patch until every package in the chain is released again.

## Current damage

**6 of 24** published runtime pins are already behind:

| package | dependency | pinned | latest |
|---|---|---|---|
| `@storyblok/js@6.3.1` | `storyblok-js-client` | `7.7.3` | 7.7.4 |
| `@storyblok/js@6.3.1` | `@storyblok/richtext` | `5.2.1` | 5.3.0 |
| `@storyblok/nuxt@11.1.3` | `@storyblok/vue` | `11.2.2` | 11.3.0 |
| `@storyblok/migrations@0.4.0` | `@storyblok/management-api-client` | `0.7.0` | 0.7.1 |
| `@storyblok/migrations@0.4.0` | `@storyblok/richtext` | `5.2.1` | 5.3.0 |
| `@storyblok/js@6.3.1` | `@storyblok/preview-bridge` | `^2.1.6` | 2.2.1 ✅ resolves |

The last row is the control: `@storyblok/preview-bridge` is the one Storyblok dependency that is *not* a workspace dep, it already uses a caret, and it is the only one that stays current on its own.

## Why this is worth fixing now

This is the mechanism that trapped `@storyblok/nuxt` 6.x users on the `cv=undefined` bug:

```
@storyblok/nuxt@6.2.4 → @storyblok/vue@^8.1.11 → @storyblok/js@3.2.3 → storyblok-js-client 6.10.8 (exact)
```

The fix shipped in `storyblok-js-client@6.10.11`, but the exact pin meant those users could not resolve it without a major SDK upgrade. `^6.10.8` would have delivered it automatically on their next install.

To be clear about scope: this does **not** fix anything for anyone currently broken — published manifests are immutable. It stops the next one.

## The change

Runtime `dependencies` of published packages → `workspace:^` (23 deps across 11 packages).

Deliberately **not** changed:
- `devDependencies` (43 of them) — never installed by consumers, so the range is irrelevant; `workspace:*` stays correct for internal linking.
- Private playgrounds — never published.
- `peerDependencies` — none use the workspace protocol.

## Review follow-ups

Code review of the sweep surfaced seven items. Each is a separate commit with its full rationale in the commit body.

**The caret is not safe while a package also bundles the dependency.** `@storyblok/vue` and `@storyblok/astro` externalized only `vue` / `astro`, so their bundles carried an inlined copy of `@storyblok/js`, `@storyblok/richtext`, *and* `storyblok-js-client` while declaring caret ranges for two of them. The exact pin was the only thing keeping the inlined copy and the installed copy identical; a range lets them drift, so the executed code is the frozen copy while `dist/index.d.ts` and any direct consumer import resolve against a newer one. Both now externalize their workspace deps, matching what `@storyblok/react` and `@storyblok/js` already did. `@storyblok/vue`'s UMD output declares them as globals using the same names `@storyblok/js` already uses; its bundle drops to 8 kB.

**`storyblok-js-client` was referenced by declarations that never declared it.** `useStoryblok` had no explicit return type in `@storyblok/vue` and `@storyblok/react`, so TypeScript resolved the default type argument and emitted `import('storyblok-js-client').ISbComponentType` into the published `.d.ts`. That package is not in either manifest, so consumers on pnpm's isolated layout or Yarn PnP hit TS2307 from inside our own types. Both now annotate the return type, which leaves the resolved type unchanged and routes the reference through the declared `@storyblok/js`.

**`preserveLocalDependencyProtocols` is now explicit in `nx.json`.** The whole scheme depends on it, it is only an implicit default, and if it ever flips back `nx release version` rewrites all 23 specifiers to literal versions inside a generated release commit that nobody reviews line by line.

**`pnpm check:workspace-protocol` enforces the invariant in CI.** Nothing did before, while ~60 legitimate `workspace:*` occurrences remain in `devDependencies` as the shape a contributor copies. The guard fails when a runtime section of a non-private package uses anything other than `workspace:^`, and exempts `devDependencies` on purpose.

**Docs and an ADR.** `RELEASING.md` still described tarballs pinning exact versions, so its alpha-guard section was rewritten around ranges, plus a new "Workspace Dependencies" section. `adr/0014` records the decision, the carve-outs, the rejected alternatives, and the consequences.

**Accepted, not solved: pre-release channels lose the exact-build guarantee.** A range built from a pre-release version, such as `^7.8.0-next.0`, also matches stable `7.8.0` and `7.9.0`, so an artifact published from a pre-release branch can resolve a stable dependency instead of the pre-release it was built against. This only affects edges where a package *and* its workspace dependency both release from the same pre-release channel (`storyblok`, `@storyblok/angular`, `@storyblok/migrations`, `@storyblok/astro`); `@storyblok/nuxt` → `@storyblok/vue` is unaffected because `@storyblok/vue` releases from `main` only. Pre-release consumers are already opted into instability and the alternatives (per-branch overrides, publish-time channel checks) cost more than the risk, so the condition and the per-edge escape hatch are documented in `RELEASING.md` and the ADR instead.

## Verification

`pnpm pack` on `@storyblok/js`, reading the manifest out of the actual tarball:

```json
{
  "@storyblok/preview-bridge": "^2.1.6",
  "@storyblok/richtext": "^5.3.0",
  "storyblok-js-client": "^7.7.4"
}
```

vs. what is on npm today for the same package: `"@storyblok/richtext": "5.2.1"`, `"storyblok-js-client": "7.7.3"`.

`pnpm pack` on `@storyblok/vue` confirms the range and the externalization together — the declared dependency and the executed code are now the same copy:

```json
{ "@storyblok/js": "^6.3.1", "@storyblok/richtext": "^5.3.0" }
```

and `dist/storyblok-vue.mjs` imports `@storyblok/js`, `@storyblok/richtext`, and `vue`, rather than inlining them.

- `pnpm nx affected -t build,lint,test:unit,test:types` → 24 projects, exit 0 (includes the Astro, Next, and React playground builds)
- `pnpm check:workspace-protocol` → passes; verified it also fails on a planted `workspace:*` runtime dep
- No `storyblok-js-client` reference remains in any `@storyblok/vue` or `@storyblok/react` published declaration
- `publint` on all changed packages → no errors (only pre-existing `engines.node` / `sideEffects` suggestions)
- `pnpm install` → lockfile changes are specifier-only; every workspace dep still resolves to `link:../<pkg>`, local dev unaffected

Cypress e2e suites were not run locally.

## Notes

**Version bumps.** The protocol change itself bumps no versions: `commitlint.config.js` documents that meta scopes are "intentionally NOT packages, so such commits bump no package version", so the caret reaches npm only as each package is released for its own reasons. The review follow-ups do change source in `@storyblok/vue`, `@storyblok/astro`, and `@storyblok/react`, so those three will bump on the next release, which is intended — the externalization is a real packaging fix that needs to ship. This is also why the release-guard comment fires on this PR.

**Caret on `0.x` is narrower than it looks.** `^0.7.1` allows `>=0.7.1 <0.8.0`, so for these four, only patches flow and a minor still requires a dependent release:

`@storyblok/api-client@0.7.1` · `@storyblok/management-api-client@0.7.1` · `@storyblok/live-preview@0.5.1` · `@storyblok/schema@0.4.0`

Affects `@storyblok/angular`, `@storyblok/migrations`, and `storyblok` (CLI). Fully solved only when those packages reach 1.0. Flagging it rather than leaving it as a silent gap.

**Unrelated, spotted while editing `RELEASING.md`:** the "Supported Pre-release Channels by Package" table lists `@storyblok/migrations` as "`alpha` only", but `packages/migrations/package.json` has no `release.branches` field, which means it releases from every branch. Left alone here since it is out of scope for this PR.

## Trade-off

A consumer without a lockfile can now resolve a newer `storyblok-js-client` than the SDK was tested against. That is the standard caret bargain, and we control both sides of it, but it is a real change in what we ship: the exact pin was, in effect, a tested-together guarantee. Recorded in `adr/0014` along with the rest of the consequences.

Fixes DX-553


alexjoverm (changes requested):
(just blocking as a gate-keeper for now)

@maoberlehner I'm not fully understanding why some versions are left behind. And not sure if moving forward independently from #768 makes sense if they're somehow connected (?).

Could you host an internal session so you can drive us through the problem? They're core structural changes - it'll be important we all are fully knowledgable and aligned before jumping in the implementation :) also ideally we have a ticket shaped with those conclusions/direction by then

alexjoverm (approved):

PR #766, opened 2026-08-19
chore(repo): lint and format the whole repo on commit and in CI

maoberlehner:
A markdown-only commit on another branch broke `Check formatting` in CI, because the
hook never looked at the file.

- **Pre-commit hook:** drop the glob from `lint-staged.config.js`. It matched only
  `{packages,tools}/**/*.{js,ts,jsx,tsx,astro,json}`, so a change to a root script, an
  `.agents/` script, a `.vue` file, or any markdown ran neither the linter nor the
  formatter. Every staged file now goes through, whatever its path or extension.
- **Add the root Oxlint pass to the hook.** `nx affected` can only reach files that
  belong to a project, so repo config, the release scripts, and the agent scripts were
  unreachable no matter how the glob was written. That is the gap the root pass fills.
- **Keep `nx affected` for the project lint**, rather than switching to `run-many`: a
  lint error in a project you never touched should not block your commit.
- **`--concurrent=false`** in `.husky/pre-commit`, so the formatter still runs after
  `oxlint --fix` (an autofix lands unformatted otherwise).
- **CI:** add the same root Oxlint pass. `Check` lints only the projects in the matrix,
  so nothing checked the root files there either. It needs `@storyblok/lint-config`
  built first, since `oxlint.config.ts` imports it.

## Cost

Measured locally: the root pass is 0.8s, `vp fmt` 0.7s, and `nx affected -t lint` for one
package ~5s (it pulls in that project's `build` deps).

## Verification

- Staged a root-level file with an unused variable: the hook blocks the commit and
  reverts cleanly. The old config would not have looked at it.
- Staged deliberately unformatted markdown: the hook reformats it before the commit.
- `pnpm exec oxlint --disable-nested-config` exits 0 on `main`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -67,9 +67,12 @@ jobs:
       # Every project lints itself in the `Check` step below. This one covers
       # what no project owns: repo config, the release scripts, the agent
       # scripts. Same command the root `lint` script runs.
+      #
+      # `oxlint.config.ts` imports the built `@storyblok/lint-config`, so that
+      # project has to be built first: this step runs before `Check`.
       - name: Check root lint
         if: matrix.os == 'ubuntu-latest'
-        run: pnpm exec oxlint --disable-nested-config
+        run: pnpm nx build @storyblok/lint-config && pnpm exec oxlint --disable-nested-config
 
       - name: Check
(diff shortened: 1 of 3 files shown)


alexjoverm (approved):
Nice one!

PR #767, opened 2026-08-19
feat(cli): cli find command

alexjoverm:
### Context

Teams keep asking the same shape of question about a space, and none of it is answerable from the
CLI today:

- Which pages still use this block — is it safe to delete?
- What breaks if I remove this story?
- Which images shipped without alt text?
- What's live but pointing at something that isn't?
- What's sitting published-with-pending-edits before we cut a release?

`stories pull` and `migrations run` take `--starts-with` and `--query` — a path prefix and
root-level content fields. Anything about nested blocks, references, publish state or workflow is
out of reach, so today it's the web UI, or a throwaway script against the API, written again per
question and per space.

### The command

`find` answers exactly one question — **which stories** — and prints the answer as JSONL, one story
per line.

```bash
# Every image with no alt text, anywhere in the space
storyblok stories find --space 12345 --capi-filter --skip-content \
  --where "$..[?(@.fieldtype == 'asset' && @.alt == '')]" | jq -r '.full_slug'

# Is this block still used?
storyblok stories find --space 12345 --includes-block legacy_hero --skip-content | wc -l

# Links pointing at a slug that has since moved
storyblok stories find --space 12345 --check-references --capi-filter \
  --where "$._ref_issues[?(@.type == 'stale_url')]"
```

Filters combine with AND, and each is resolved wherever it's cheapest — blocks, tags, publish state
and workflow stage on the server; `--where` (JSONPath) locally, for what the API can't express.

Printing one **self-contained story per line** is the other half of the design. The answer composes
with whatever comes next — `jq`, a file, `xargs`, eventually another CLI command reading it on
stdin. `find` owns the question; something else owns the change. Results stream as they match, a
reader that leaves stops the run, and a slow consumer paces the producer rather than being buffered
at.

### Making it fast enough to actually use

The story listing carries no content, so every story a content filter has to see costs one request
at 6/s — a four-thousand-story space is eleven minutes before you've asked anything interesting.
Two opt-in flags fix that: one reads content from the CDN in bulk (25 per request at 50/s) and only
fetches the stories that matched, the other skips content entirely when the question doesn't need
it. The bulk stage only ever *prunes* — anything it can't decide falls through and is settled the
slow way — so the result set is identical either way.

| Over a 4,016-entry space | before | after |
| --- | --- | --- |
| Every image with no alt text (1,770 hits) | ~11 min | **~9s** |
| Full reference audit (296 stories with issues) | ~11 min | **9.4s** |

### Also included

- A reference integrity check — broken targets, unpublished targets, cached URLs that have gone
  stale after a restructure
- The JSONL pipe extracted as shared CLI infrastructure, so the next command to join the pipe
  doesn't grow its own copy of it
- Progress and diagnostics properly separated from results, so a run is pipeable without a quiet
  flag
- A scenario harness that measures the command against a real space

### Where the reasoning lives

- [ADR-0015](adr/0015-cli-stories-find-command.md) — why the Management API answers every query and
  the CDN is only an accelerator, and where the flag surface deliberately stops
- [ADR-0016](adr/0016-cli-jsonl-pipe-contract.md) — the line contract between commands, and why `-`
  is explicit rather than sniffed from stdin
- [Command README](packages/cli/src/commands/stories/find/README.md) — full reference, use cases and
  caveats


adr/0013-cli-stories-find-command.md
```diff
@@ -0,0 +1,119 @@
...
+  `--translation-status` filtering. CAPI does not expose per-language `published_at` or
+  `unpublished_changes`.
+
+CAPI's exclusive features (`language` content parameter, date-range filters, `resolve_relations`)
+are either not needed for discovery filtering or expressible via `--query`/`--where`.
+
+### 2. No `--ignore-case` flag
+
+The MAPI `text_search` is always case-insensitive. The backend uses PostgreSQL `ILIKE` and
+`.downcase()` (`storyrails/app/models/concerns/text_searchable.rb:25-59`). The `like` operator in
```
maoberlehner: Storyrails details we might consider to exclude.

packages/cli/src/commands/stories/find/actions.ts
```diff
@@ -0,0 +1,231 @@
...
+        "--capi-filter needs at least one --where filter: without one, every listed story is a match and none can be pruned.",
+      );
+    }
+    // Parsed here as well as at build time so a malformed value fails as a usage
+    // error, next to the flags it belongs with.
+    parseCapiParams(options.capiParams);
+  }
+}
+
+export function buildQueryParams(
```
maoberlehner, line 73: Does this exist already? If not, maybe useful elsewhere, too?

packages/cli/src/commands/stories/find/capi.ts
```diff
@@ -0,0 +1,190 @@
...
+  const space = await fetchSpace(spaceId);
+  const token = space?.first_token;
+
+  if (!token) {
+    throw new CommandError(
+      `No Content Delivery API token available for space ${spaceId}. --capi-filter reads content through the CDN, which needs the space's preview token.`,
+    );
+  }
+
+  const client = new Storyblok({
```
maoberlehner: Should we swap `js-client` dependency with the new CAPI client (also in `assets/actions.ts`)?

packages/cli/src/commands/stories/find/README.md
```diff
@@ -0,0 +1,708 @@
+# `storyblok stories find`
```
maoberlehner, line 1: Helpful during review, should be deleted before merge.

packages/cli/src/commands/stories/find/command.ts
```diff
@@ -0,0 +1,732 @@
...
+  filters: ClientFilter[];
+};
+
+/**
+ * Resolves the CDN token and builds the batch fetcher.
+ *
+ * Done before the pipeline starts so a space without a usable token fails as a
+ * usage error, rather than after a few thousand stories have been listed.
+ */
+async function prepareCapiFilter({
```
maoberlehner: nitpick: in other commands, we've moved supporting functions to separate files. Might be worth doing the same here.

maoberlehner (changes requested):
Looking forward to the final `find` command! Found a couple of issues in a first shallow review round. Will take a closer look at using it next.

maoberlehner:
Claude QA findings:

## Findings

**1. `--query` silently ignores unparseable input** (correctness)
`--query "not-a-query"` returns the **entire space** with exit 0 and no warning — `parseFilterQuery` `continue`s past clauses with no `=` or fewer than two `[…]` groups, yielding `{}`. This is exactly what `assertSupportedOptions`' own comment calls out: "the user gets a full, plausible result set that answers a different question." Fix: error when a non-JSON `--query` produces zero clauses. (Pre-existing in `stories pull`, but `find` is the command people will script against.)

**2. `--check-references` reports at most one issue per reference, so `unpublished` masks `stale_url`**
`detectIssues` `continue`s after the unpublished branch. Reproduced: a richtext link with `cached_url: "/faq-old"` pointing at unpublished `faq` (actual `faq`) is reported only as `unpublished`. In a draft-heavy space, the documented "outdated URLs after a restructure" audit (`--where "$._ref_issues[?(@.type == 'stale_url')]"`) will under-report. Emit both, or document the precedence.

**3. `--capi-params version=published` mixes draft decisions into a "what is live" answer**
Stories with no published content are "undecided" and get fetched and evaluated against **draft** MAPI content, so a `version=published` run reported `home` and `faq` — neither has ever been published. The published-content narrowing itself is correct (`contact`, published without a hero, was correctly excluded even though its draft has one). Suggest skipping undecidable stories when a non-`draft` version is requested, or documenting that it must be paired with `--publish-status published` (verified: that combination returns the correct single result).

**4. `find | head -n` does not stop the run**
The CLI ignores the closed stdout and keeps fetching the whole scope: still running 8s after `head -3` exited, and `| head -5` over 150 stories takes the full 25.6s. Lines *do* stream promptly (first five within ~0.1s of each other), so only termination is missing — the README example implies otherwise. Exit on EPIPE.

**5. Exit-code semantics are surprising and undocumented**
A run with two failed CAPI batches exits **1** while producing complete, correct output (43/43 stories, fallback worked as documented). A script using `find > out.jsonl || exit` would throw away a correct result. Document 0 / 1 (run errors, results may still be complete) / 2 (usage).

**6. Smaller DX items**
- `--query '{"a":'` → exit **1** with bare `Unexpected end of JSON input`, no mention of the flag. Every other validation error is an excellent `CommandError` at exit 2.
- `--search-mode semantic` is advertised in `.choices()` but always errors. Drop it or mark it unavailable in the help text.
- `--verbose` prints every fetched story object to stderr — a 4k-story run floods the terminal and duplicates the stdout payload.
- Empty result sets render progress as `0/1 processed`.
- `--references hello` (non-UUID) returns 0 results at exit 0; a shape check would fail fast like the rest of the command does.
- `find --help` lists no global options, so `--api-rate-limit`/`--verbose`/`--path` — which the README leans on — are undiscoverable. (CLI-wide, not new here.)

**7. Doc bug: `--includes-block` is AND, not OR**
The README example reads "Stories using a hero **or** a pricing table anywhere in their content." `--includes-block hero,cta` returned only the 2 stories containing **both**, not the 4 containing either (verified against content). Fix the wording.

pnpm-lock.yaml
```diff
@@ -183,34 +110,34 @@ importers:
...
+        specifier: 1.76.0
+        version: 1.76.0(oxlint-tsgolint@7.0.2001)(vite-plus@0.2.8(@arethetypeswrong/core@0.18.2)(@opentelemetry/api@1.9.0)(@types/node@20.19.37)(@vitest/coverage-v8@4.1.10(@vitest/browser@4.1.10(msw@2.12.10(@types/node@20.19.37)(typescript@6.0.3))(vite@6.4.1(@types/node@20.19.37)(jiti@2.6.1)(less@4.6.4)(lightningcss@1.33.0)(sass@1.99.0)(terser@5.46.0)(tsx@4.21.0)(yaml@2.9.0))(vitest@4.0.18))(vitest@4.0.18))(@vitest/ui@4.1.10)(esbuild@0.28.1)(happy-dom@20.8.9)(jiti@2.6.1)(jsdom@27.4.0(@noble/hashes@1.8.0))(less@4.6.4)(msw@2.12.10(@types/node@20.19.37)(typescript@6.0.3))(publint@0.3.22)(sass@1.99.0)(svelte@5.55.0)(terser@5.46.0)(tsx@4.21.0)(typescript@6.0.3)(unrun@0.2.28(@emnapi/core@1.11.1)(@emnapi/runtime@1.11.1)(synckit@0.11.12))(vite@6.4.1(@types/node@20.19.37)(jiti@2.6.1)(less@4.6.4)(lightningcss@1.33.0)(sass@1.99.0)(terser@5.46.0)(tsx@4.21.0)(yaml@2.9.0))(yaml@2.9.0))
       rxjs:
         specifier: ~7.8.0
         version: 7.8.2
       typescript:
-        specifier: 'catalog:'
+        specifier: ~6.0.3
         version: 6.0.3
```
maoberlehner, line 216: Claude finding:
Reverts resolutions present at the merge base: typescript@6.0.3 → 5.8.3, @babel/core@7.29.7 → 7.29.0, plus a re-added packages/eslint-config importer block. main now pins typescript: 6.0.3 via catalog, so merging as-is either reverts TS 6 or breaks --frozen-lockfile. Rebase and regenerate with only the json-p3 addition.

packages/cli/src/commands/stories/find/streams.ts
```diff
@@ -0,0 +1,244 @@
...
+}
+
+/**
+ * Replaces the per-story MAPI content fetch with one CAPI page per 25 stories,
+ * for the sole purpose of deciding which stories are still worth fetching.
+ *
+ * The stage only ever *prunes*. A story CAPI does not answer for — a folder, a
+ * story the CDN has not got, a batch that failed — passes through undecided, so
+ * the exact answer still comes from MAPI content downstream. That is what keeps
+ * the result set identical to a run without the flag, and why the `--where`
```
maoberlehner: Claude finding:
doc/behavior mismatch** — docblock says `--where` runs twice ("authoritatively against the MAPI story"), but `command.ts:416` wires `isAlreadyMatched` from `onCandidate`, and `filterStoriesStream` short-circuits on it. Doc is simply wrong. Behavioral impact is narrower than reported: `DEFAULT_CAPI_PARAMS` is `version: draft` and the client cache is `type: none`, so it only bites with an explicit `--capi-params version=published` (or CDN edge staleness). Fix the comment + README:215-area claim, or drop `isAlreadyMatched`.

packages/cli/src/commands/stories/find/README.md
```diff
@@ -0,0 +1,708 @@
...
+### `--skip-content`
+
+If nothing in the query needs content, do not fetch it. The content phase disappears entirely and
+the run becomes the page walk, which moves 100 stories per request instead of one: the same
+167-story scope that takes 29.1s with content takes **4.0s** without it.
+
+The output is the story listing —
+[the story object](https://www.storyblok.com/docs/api/management/stories/the-story-object) as the
+listing returns it, with slugs, ids, publish state and timestamps, but **no `content` field**. The
+`content_summary` digest is included, and is often enough on its own.
```
maoberlehner: Claude finding:
`content_summary`** — MAPI returns `{}` without `with_summary` (`types.gen.ts:10314`) and `buildQueryParams` never sets it. Claim is false as written.

packages/cli/src/commands/stories/find/actions.ts
```diff
@@ -0,0 +1,231 @@
...
+  if (options.startsWith !== undefined) {
+    params.starts_with = normalizeStartsWith(options.startsWith) || undefined;
+  }
+
+  // Filter query and container block (both contribute to filter_query)
+  if (options.query || options.containerBlock) {
+    params.filter_query = {
+      ...(options.query ? parseFilterQuery(options.query) : {}),
+      ...(options.containerBlock ? { component: { in: options.containerBlock } } : {}),
+    };
```
maoberlehner: Claude finding:
`containerBlock` is spread after `parseFilterQuery(query)` into the same object, so `--query="[component][in]=hero" --container-block product` silently drops `hero`.

packages/cli/src/utils/error/api-error.ts
```diff
@@ -23,6 +23,7 @@ export const API_ACTIONS = {
   update_component_preset: "Failed to update component preset",
   delete_component_preset: "Failed to delete component preset",
   delete_component: "Failed to delete component",
+  find_stories: "Failed to find stories",
```
maoberlehner: Claude finding:
`find_stories` defined, zero references anywhere in `packages/cli/src`; find goes through `pull_stories`/`pull_story`.

maoberlehner (changes requested):

adr/0013-cli-stories-find-command.md
alexjoverm: good catch 😓

packages/cli/src/commands/stories/find/actions.ts
alexjoverm: Great point, it's definitely a reusable util!

packages/cli/src/commands/stories/find/capi.ts
alexjoverm: 100%

packages/cli/src/commands/stories/find/README.md
alexjoverm: I'll actually keep this comment unresolved, so the CI won't let me merge till fixed

alexjoverm:
@maoberlehner thanks for the feedback and first review! Some answers to the general findings:

1. _--query silently ignores unparseable input_ -> not sure if there is anything we can/should do here.. at the end, the command maps to the one on MAPI. Will check if it's API behaviour, or something wrong on our end.
2. _--check-references reports at most one issue per reference, so unpublished masks stale_url_ -> weird, I actually have multiple issues per reference in my local runs. Not sure if either was fixed later, but will double check just in case 👍 
3. _--capi-params version=published mixes draft decisions into a "what is live" answer_ -> yeah... this is one of those cases that wouldn't make sense a user do that, was thinking to cover it more from docs. IMO this also applies to other cases, as the command is too flexible that user can do multiple non-sense things if they don't know properly what they're doing. Or maybe we can reduce that flexibility a bit, will think about it. 
7. _Doc bug: --includes-block is AND, not OR_ -> good catch! def all are AND

Will check rest of findings, thanks!

packages/cli/src/commands/stories/find/capi.ts
```diff
@@ -0,0 +1,246 @@
...
+    // makes this explicit rather than incidental.
+    cache: { provider: NO_CACHE_PROVIDER },
+    // A failed batch has to reach the stage's error callback to be counted and
+    // reported, so an HTTP error must reject rather than resolve as `{ error }`.
+    throwOnError: true,
+    // Off, and stated rather than left to the default. Turning it on would make
+    // `--capi-params resolve_relations` put each related story where its UUID
+    // was, so a `--where` expression could reach into a referenced story's
+    // content; as it stands the CDN's `rels` sidecar is ignored.
+    inlineRelations: false,
```
alexjoverm, line 206: @maoberlehner what do you think about this? I see different pros/cons.. on one side, leaving it as `false`  might make it more predictable as the payload to filter will be exactly as the CAPI returns it, in the `rels` (and this command has enough caveats for the user to understand already) - on the other side, inlining relations can make the --where filters easier in the cases where they want to filter based on the content of a resolved relation (which I expect to be a rare case).

Just as an example - for this example, with inlineRelations: false would require 2 passes and an intermediate output file:

```bash 
# Pass 1 — which quote stories have a C-level role?
storyblok stories find -s $S --container-block new_quote --capi-filter --skip-content \
  --where "\$[?search(\$.content.role, 'C[TEIM]O')]" \
  | jq -r '.uuid' > authors.txt      

# Pass 2 — which lp pages point at any of them?
storyblok stories find -s $S --starts-with lp --capi-filter --skip-content \
  --where "\$..[?(@.component == 'quotes_item' && search(@.quote, '$(paste -sd'|' authors.txt)'))]"
```

compare to just a single statement using `inlineRelations: true`.

```bash 
storyblok stories find -s $S --starts-with lp --capi-filter --skip-content \
  --capi-params "resolve_relations=quotes_item.quote" \
  --where "\$..[?(@.component == 'quotes_item' && search(@.quote.content.role, 'C[TEIM]O'))]"
```

Both achieve the same output.

packages/cli/src/commands/stories/find/run.ts
```diff
@@ -0,0 +1,172 @@
...
+          : "Stopped early on purpose: the command reading this output took what it needed and closed the pipe. ") +
+          "This is not an error — the run exits 0. The counts below cover only the part of the scope that ran.",
+      );
+    }
+
+    // An empty result under a bare `--skip-content` is ambiguous: the filters
+    // genuinely matched nothing, or they were written against the content that
+    // was never fetched. Only the user can tell the two apart, so name the
+    // possibility exactly when it applies rather than rejecting the combination.
+    if (skipContent && !capi && filters.length > 0 && filtered.succeeded === 0) {
```
maoberlehner, line 100: Claude:

> `--skip-content` + `--where` returns partially correct results with no warning and exit 0.
>
> `--skip-content` sets `with_summary: true` (`actions.ts:177`), so each listed story carries `content_summary` — a truncated content copy: root fields survive, nested `body`/bloks are dropped. `--where` runs against the whole story object, so `$..` descends into it and matches on the truncated half.
>
> With `--where '$..[?@.component == "hero" || @.category == "design"]'` I get 6 stories: `about`, `blog-color-theory`, `blog-design-tips`, `blog-responsive`, `faq`, `home`. Adding `--skip-content` gives 3 — `blog-color-theory`, `blog-design-tips`, `blog-responsive` — and reports `Results: 3 stories matched (10 listed, no content fetched)` with exit 0. `about`, `faq` and `home` are dropped silently.
>
> This guard only fires on `filtered.succeeded === 0`, so any expression matching *something* off `content_summary` gets no warning at all. That is the dangerous case, because the answer looks plausible.
>
> Two more things to fix with it. The warning text is factually wrong: the listing does not carry "story metadata only" under `--skip-content`, it carries `content_summary` too, which is the whole cause. And the README says "An expression that does read content matches nothing, since the content is not there to read", plus a table row asserting recursive descent does not survive `--skip-content` — it partially survives. Nothing tests this interaction; `content_summary` appears once, in `actions.test.ts:242`, only asserting the param is set.

packages/cli/src/commands/stories/find/check-references.ts
```diff
@@ -0,0 +1,269 @@
...
+    // Written through the sink rather than in a loop of its own, for the two
+    // properties that come with it: a slow reader paces the emit instead of
+    // having it buffered ahead of them, and a reader that leaves stops it — a
+    // plain loop would go on deciding every remaining candidate and counting
+    // matches nobody ever received.
+    await pipeline(Readable.from(reportIssues()), output.sink, { signal: output.signal });
+  } catch (error) {
+    // Same contract as `find`: a reader that has taken what it wanted and left
+    // ends the run cleanly. See `runFind`.
+    if (!isDownstreamClosed(error)) {
```
maoberlehner, line 196: Claude:

> `--check-references` + `--limit` fails the run. `--limit` aborts with `LimitReachedError`, but this catch only forgives `isDownstreamClosed`, so the `AbortError` reaches `handleError`:
>
> `storyblok stories find -s <space> --check-references --limit 1` prints `Results: 2 stories with reference issues (10 checked, 0 external targets resolved)` and then `▲ error The operation was aborted`, exit 1.
>
> The contract is stated at `README.md:182` and `README.md:481` ("A limited run exits **0**"). The `runFind` path gets it right — same space, `--limit 1` exits 0 with `✔ Stopped early on purpose`. Reuse `isDeliberateStop` (`run.ts:68`) plus the `stoppedByLimit` message.
>
> Worth knowing this reads as a hard failure: a blackbox tester hit it cold and concluded the reference audit was broken.

packages/cli/src/commands/stories/find/check-references.ts
```diff
@@ -0,0 +1,269 @@
...
+      for (const { story, refs } of candidates) {
+        const issues = detectIssues(refs, uuidToMeta);
+        if (issues.length === 0) {
+          continue;
+        }
+        const enriched: Story & { _ref_issues: RefIssue[] } = { ...story, _ref_issues: issues };
+        if (!applyClientFilters(enriched, whereFilters)) {
+          continue;
+        }
+        matched += 1;
```
maoberlehner, line 182: Claude:

> The summary over-counts on an early stop. `matched += 1` runs before the line is written, so stories decided while the sink tears down still get counted.
>
> Same `--limit 1` run as above: exactly 1 JSONL line on stdout, while stderr reports `2 stories with reference issues` and `10 checked, 9 with references, 2 with issues`. The unlimited baseline on that space is 9 and 9.
>
> Count it where `createJsonlOutput` counts lines, or after the yield resumes.

packages/cli/src/commands/stories/find/pipeline.ts
```diff
@@ -0,0 +1,252 @@
...
+                logOnlyError(error, { batchSize: size });
+              }
+            },
+          }),
+        ]
+      : []),
+    ...(skipContent
+      ? []
+      : [
+          fetchStoryStream({
```
maoberlehner, line 207: Claude:

> `--sort` does not sort. This runs ~12 fetches concurrently and pushes in completion order, so ordering is broken even without `--limit`: a plain `--sort slug:asc` comes back as `about, post-10, post-13, post-11, post-1, post-12, post-17, post-19, …`, and `--sort name:asc` over 150 stories gives `about, about, Blog, API Reference, API Reference, Blog, …`.
>
> With `--limit` it is also not the top N, and not deterministic. Three runs of `--sort slug:asc --limit 10`:
>
> - `about, post-10, post-1, post-11, post-12, post-14, post-16, post-17, post-22, post-18`
> - `about, post-1, post-10, post-12, post-11, post-14, post-15, post-18, post-13, post-16`
> - `about, post-1, post-13, post-10, post-11, post-12, post-16, post-18, post-17, post-15`
>
> The first drops `post-13` and `post-15` and includes `post-22`. The scramble window is about 6, so `--limit 5` happens to return the right set and the bug only shows from `--limit 10` up.
>
> Correct under `--skip-content` (single-threaded): that returned the exact sorted top 10, identical across runs.
>
> Claims to reconcile: `README.md:438-439` and `README.md:913`. Since the README pitches `--sort updated_at:desc --limit 10` as the way to get the ten most recent, this is roughly the second thing a new user tries after plain search.

packages/cli/src/commands/stories/filter-query.ts
```diff
@@ -23,27 +32,80 @@ export function parseFilterQuery(input: string): FilterQuery {
...
+    result[field] = { ...result[field], [operation]: clause.slice(eq + 1) };
+  }
+
+  if (ignored.length > 0) {
+    throw new CommandError(
+      `Invalid --query clause${ignored.length > 1 ? "s" : ""}: ${ignored.join(", ")}\n${CLAUSE_SYNTAX_HINT}`,
+    );
   }
+
   // Field/operator names are free-form user input; MAPI validates them at runtime.
```
maoberlehner, line 61: Claude:

> MAPI does not validate them — it ignores unknown operators, so a mistyped operation silently disables the filter instead of erroring.
>
> `-q '[category][eq]=technology'` returns `Results: 10 stories found` — the whole space, where 2 stories actually match. `-q '[category][bogusop]=x'` does the same.
>
> `[eq]` instead of `[in]`/`[is]` is the exact typo people make, and "returns everything, exit 0" is the worst way to fail. `--references` already got a dedicated guard for this failure class; an operator allowlist here would match it.

packages/cli/src/commands/stories/filter-query.ts
```diff
@@ -11,10 +15,15 @@ export type FilterQuery = NonNullable<StoryListQuery["filter_query"]>;
...
  * string straight through instead yields a malformed `filter_query=<string>`
  * param that the API silently ignores.
+ *
+ * Input that parses to nothing is a usage error rather than an empty filter. A
+ * `filter_query` that never reaches the wire does not narrow anything, so the
+ * command would answer a different, much larger question than the one asked —
+ * with a plausible-looking result set and a zero exit code to hide it.
  */
 export function parseFilterQuery(input: string): FilterQuery {
   const trimmed = input.trim();
```
maoberlehner, line 29: Claude:

> This guarantee does not hold — the trim guard just below returns `{}` rather than raising, and it is not the only hole.
>
> `-q '{}'`, `-q '&'`, `-q '   '` and `-q ''` all return `Results: 10 stories found`. `''` short-circuits on the falsy check in `buildStoryScopeParams` (`query-params.ts:43`), `'   '` returns `{}` here, `'&'` has every clause `continue`d so `ignored` stays empty, and `'{}'` goes through `parseAsJson` unchecked. Only the malformed-clause path throws.
>
> Minor on its own, but it is the one guarantee the code states and then does not keep.

packages/cli/src/commands/stories/find/README.md
```diff
@@ -0,0 +1,1211 @@
...
+# German specifically, which the API filter cannot express
+storyblok stories find --space 12345 \
+  --where "$.stages[?(@.language == 'de' && @.workflow_stage_id == 42)]"
+```
+
+Prefer the flag where it fits: the `--where` form reads every story in the space to answer.
+
+### `--sort`
+
+Order the results. The API applies this, so it decides which stories are on the first page, which is
```
maoberlehner: Claude:

> This is the claim that does not hold — see my comment on `fetchStoryStream` in `pipeline.ts`. The API does apply the sort, but the concurrent fan-in reorders the results before they reach stdout, so `--sort` alone comes back locally scrambled and `--sort` with `--limit` is neither the top N nor sorted. Same for the "a sorted run streams" claim at line 913.
>
> It holds under `--skip-content`, which is single-threaded.

packages/cli/src/commands/stories/find/README.md
```diff
@@ -0,0 +1,1211 @@
...
+
+The summary reports what it saved. Here 170 of 191 stories never needed an individual request, and
+the 32.3s run above finished in 4.9s:
+
+```
+Filtering via CAPI: 21/191 candidates, 170 pruned before fetch, 0 undecided, 0 batch(es) failed.
+Fetching content: 21/21 succeeded, 0 failed.
+```
+
+Good to know:
```
maoberlehner: Claude:

> This list needs an i18n caveat: `--capi-filter` silently prunes when the predicate targets a field-level translation, because CDN content drops every `<field>__i18n__<lang>` key.
>
> `--capi-filter --where "$[?($.content.title__i18n__de)]"` gives 0 lines, exit 0, `Results: 0 stories matched (1 of 14 listed fetched from MAPI, 13 pruned by the CAPI filter)`. The same `--where` without the flag returns the matching story.
>
> Grounded, not an unknown-language artifact: MAPI holds `title__i18n__de` and `title__i18n__fr` for that story, and the CDN draft document is `{_uid, title, component, _editable}` — both suffixed keys gone, including `fr`, which is a configured language in the space.
>
> Meanwhile "Translation coverage" (line 945) recommends exactly this predicate under `--capi-filter` (lines 957, 961, 965). The promise in this section is line 738 ("Same question as without the flag, a fraction of the requests") and lines 733-736 ("nothing reaches stdout untested"), and the source comment at `streams.ts:96-97` says it "keeps the result set identical to a run without the flag".
>
> Minimum: a caveat here. Better: detect `__i18n__` in a `--where` expression and either refuse or drop the stage to non-pruning.

packages/cli/src/commands/stories/find/streams.ts
```diff
@@ -0,0 +1,260 @@
...
+ * for the sole purpose of deciding which stories are still worth fetching.
+ *
+ * The stage only ever *prunes*. A story CAPI does not answer for — a folder, a
+ * story the CDN has not got, a batch that failed — passes through undecided, so
+ * the exact answer still comes from MAPI content downstream. That is what keeps
+ * the result set identical to a run without the flag.
+ *
+ * A story this stage *matches* is settled here: `onCandidate` records it, and
+ * the terminal stage takes that as the verdict (`isAlreadyMatched`) rather than
+ * re-running the same expressions against the MAPI story. The two documents are
```
maoberlehner, line 101: Claude:

> "The two documents are the same draft content" is not true for `--capi-params version=published` or `language=de`, and it is this sentence that justifies skipping the second evaluation.
>
> The version case is gated behind `--publish-status published|changed` (`actions.ts:86-90`), but satisfying the gate still gets it wrong. Story published as `PUBLISHED_MATCH`, draft since changed to `draft_other`, then `--publish-status changed --capi-filter --capi-params 'version=published' --where "$[?($.content.title == 'PUBLISHED_MATCH')]"` emits `{"slug":"qa-divergent","title":"draft_other"}`.
>
> The emitted document does not satisfy the predicate that selected it. `language=de` is ungated entirely.

packages/cli/src/commands/stories/find/references.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  if (node === null || typeof node !== "object") {
+    return;
+  }
+
+  const obj = node as Record<string, unknown>;
+
+  // Multilink field: { fieldtype: "multilink", linktype: "story", id: "<uuid>" }
+  if (
+    obj.fieldtype === "multilink" &&
```
maoberlehner, line 90: Claude:

> Requiring `fieldtype === "multilink"` on top of `linktype === "story"` is stricter than the reference model, so a link written without `fieldtype` is invisible to the scan.
>
> Two stories written via MAPI with the same dangling target uuid, one with `fieldtype: "multilink"` and one without: only the first is reported, the second shows zero references and the run reads clean. MAPI stored both verbatim, nothing normalized `fieldtype` in.
>
> There is no fallback either — `isStoryRelationField` (`../content-fields.ts:43-48`) only matches `option`/`options` with `source: internal_stories`, so multilink relies entirely on value-shape matching.
>
> Severity caveat worth stating: the editor always writes `fieldtype`, and its content normalizer replaces the whole value with an empty multilink when `fieldtype !== "multilink"`. So this only affects migrations and direct MAPI writes, and those values are latently lossy in the editor anyway.

adr/0015-cli-stories-find-command.md
```diff
@@ -0,0 +1,140 @@
+# ADR-0015: CLI `stories find` Command
```
maoberlehner, line 1: Claude:

> Number collision: `adr/` already holds `0015-block-tag-identity-by-name.md` and `0015-pnpm-supply-chain-hardening.md`, so this makes three. `0016` is unique and fine.
>
> Renumbering means touching `find/README.md:103` and `:1208`, this file's line 139, `adr/0016-cli-jsonl-pipe-contract.md:7` and `:38`, and the PR body.

packages/cli/src/commands/stories/find/command.ts
```diff
@@ -0,0 +1,143 @@
...
+  .option("--starts-with <path>", "scope to story subtree")
+  .option("--container-block <name>", "filter by container block type (server-side)")
+  .option("--includes-block <name>", "block presence at any depth (server-side, comma-separated)")
+  .option(
+    "-q, --query <query>",
+    "filter by root-level content attributes (server-side, MAPI filter_query)",
+  )
+  .option(
+    "--where <jsonpath>",
+    "client-side JSONPath (RFC 9535) filter (repeatable)",
```
maoberlehner, line 45: Claude:

> This needs an example. `--where` has no syntax hint anywhere a CLI user can see, and the natural spellings fail silently.
>
> Blackbox pass took six attempts. `$.content.category == "technology"` is a parse error, then four spellings in a row returned 0 results with exit 0 and no message: `$[?@.content.category == "technology"]`, `$.content[?@.category == "technology"]`, `$[?(@.content.category == "technology")]`, `$.content.category[?@ == "technology"]`. Only `$..[?@.category == "technology"]` worked.
>
> The cause — RFC 9535 `$[?…]` filters the root's member values, not the root node — is explained well in the README, but the README is invisible mid-task. One example in this string, e.g. `--where '$..[?(@.component == "hero")]'`, removes the whole detour. Silently-zero is the worst failure mode for a filter language this subtle.

packages/cli/src/commands/stories/find/command.ts
```diff
@@ -0,0 +1,143 @@
...
+  .command("find [text]")
+  .description("Find stories matching filters. Outputs JSONL to stdout (one story JSON per line).")
+  .option("-s, --space <space>", "space ID")
+  .addOption(
+    new Option("--entry-type <type>", "filter by entry type")
+      .choices(["all", "story", "folder"])
+      .default("all"),
+  )
+  .option("--starts-with <path>", "scope to story subtree")
+  .option("--container-block <name>", "filter by container block type (server-side)")
```
maoberlehner, line 37: Claude:

> Not discoverable as "content type". `--component blog` and `--content-type blog` both fail with `unknown option`, no suggestion and no pointer, and "container block type" does not read as "the story's content type".
>
> The README says it plainly ("matches a story's content type (root block)"). Suggest something like `"content type / root block, e.g. 'product'"`.

packages/cli/src/commands/stories/find/command.ts
```diff
@@ -0,0 +1,143 @@
...
+      "draft",
+    ]),
+  )
+  .option("--sort <fields>", "order results server-side, e.g. 'updated_at:desc' (comma-separated)")
+  .option("--limit <n>", "stop after this many results")
+  .option(
+    "--references <uuids>",
+    "find stories referencing these story UUIDs, comma-separated (server-side)",
+  )
+  .option("--check-references", "detect broken references and stale cached_url (client-side)")
```
maoberlehner, line 67: Claude:

> Two things missing here, both of which cost a first-time user real time.
>
> It never says the findings land in `_ref_issues`. Blackbox: `Results: 5 stories with reference issues` followed by lines that looked like ordinary stories, so I concluded the mode produced nothing useful and only found the payload by running `jq keys`. Append "…, adding a `_ref_issues` array to each emitted story".
>
> And in a draft-heavy space the mode reports everything: on a freshly seeded space every story containing a link was an issue, all `type: "unpublished"` — 5 of 10, pure noise. The intended answer is the README's `--where "$._ref_issues[?(@.type == 'broken')]"`, but that is the one thing you cannot discover from `--help`, and it leans on the `--where` syntax that is already the biggest stumbling block. A value form (`broken`, `stale_url`, `unpublished`) or an example in the help would fix it.

packages/cli/src/commands/stories/find/command.ts
```diff
@@ -0,0 +1,143 @@
...
+  .description("Find stories matching filters. Outputs JSONL to stdout (one story JSON per line).")
+  .option("-s, --space <space>", "space ID")
+  .addOption(
+    new Option("--entry-type <type>", "filter by entry type")
+      .choices(["all", "story", "folder"])
+      .default("all"),
+  )
+  .option("--starts-with <path>", "scope to story subtree")
+  .option("--container-block <name>", "filter by container block type (server-side)")
+  .option("--includes-block <name>", "block presence at any depth (server-side, comma-separated)")
```
maoberlehner, line 38: Claude:

> The AND semantics are absent from the help text, which is where the ambiguity bites. This is AND — `--includes-block hero,cta` returns `home, faq` while `cta` alone returns `home, contact, faq` — while `--tag` and `--workflow-stage` on the adjacent lines are OR.
>
> The README is emphatic that this is worth stating out loud; the help string says nothing. I only avoided getting burned because I tested one name at a time first.

packages/cli/src/commands/stories/find/command.ts
```diff
@@ -0,0 +1,143 @@
...
+  .option(
+    "--where <jsonpath>",
+    "client-side JSONPath (RFC 9535) filter (repeatable)",
+    collectValues,
+    [],
+  )
+  .option("--tag <names>", "stories carrying any of these tags, comma-separated (server-side)")
+  .option(
+    "--workflow-stage <ids>",
+    "stories at any of these workflow stage IDs, comma-separated (server-side)",
```
maoberlehner, line 52: Claude:

> Accepts non-numeric input and silently returns nothing: `--workflow-stage abc` gives `Results: 0 stories found`, exit 0.
>
> Inconsistent with the neighbours — `--limit abc` and `--references not-a-uuid` both produce specific, good usage errors. Since this flag takes opaque numeric IDs, "0 results" is indistinguishable from "wrong ID", which is the same reasoning the code already applies to `--references`.

packages/cli/src/commands/stories/find/command.ts
```diff
@@ -0,0 +1,143 @@
...
+  )
+  .addOption(
+    new Option("--publish-status <status>", "filter by publish status").choices([
+      "published",
+      "changed",
+      "draft",
+    ]),
+  )
+  .option("--sort <fields>", "order results server-side, e.g. 'updated_at:desc' (comma-separated)")
```
maoberlehner, line 61: Claude:

> A bad `--sort` column stacks two errors, names neither the column nor the valid ones, and then prints a result line after failing:
>
> `▲ error Failed to pull stories`, then `▲ error Not sortable by this column`, then `ℹ Results: 0 stories found`, exit 1.
>
> The README has the grammar and the `content.` prefix rule; the help has only `updated_at:desc`. The "Results: 0" line printed after the error is the part that could mislead someone reading logs.

packages/cli/src/commands/stories/find/command.ts
```diff
@@ -0,0 +1,143 @@
...
+import { runCheckReferences } from "./check-references";
+import type { FindOptions } from "./types";
+
+function collectValues(value: string, previous: string[]): string[] {
+  return previous.concat([value]);
+}
+
+const findCmd = storiesCommand
+  .command("find [text]")
+  .description("Find stories matching filters. Outputs JSONL to stdout (one story JSON per line).")
```
maoberlehner, line 29: Claude:

> Nothing in the output says why a story matched free-text search. `stories find Tech` returned 4 stories, 3 of which have no "Tech" in `name`. All four were correct — the matches were in content — but confirming that meant grepping the seed fixtures.
>
> Search is the headline use case and the one where a user most needs to trust the result. "Find stories matching filters" plus a bare positional `[text]` gives no hint that content is searched, not just names.

packages/cli/src/commands/stories/find/command.ts
```diff
@@ -0,0 +1,143 @@
...
+    "--skip-content",
+    "skip the per-story content fetch and emit list metadata only (no content-dependent filters)",
+  )
+  .option(
+    "--capi-filter",
+    "evaluate --where against bulk CAPI content and fetch only the matches (requires --where)",
+  )
+  .option(
+    "--capi-params <params>",
+    "extra CAPI query params for --capi-filter, e.g. '{version: published, language: de}'",
```
maoberlehner, line 78: Claude:

> A typo'd CAPI param degrades to no pruning with nothing to explain why. `--capi-params 'zzz=1'` gives `0 pruned by the CAPI filter`, all stories fetched from MAPI, exit 0, correct final answer.
>
> The exit-0 fallback is right and `1 batch(es) failed` does appear in the per-phase line, but the CDN's rejection reason is never surfaced or tied back to this flag, so the only symptom of a typo is that the optimization did nothing.

packages/cli/src/lib/pipe/output.ts
```diff
@@ -0,0 +1,294 @@
...
+   * The same output as a `Writable`, for use as the terminal stage of a
+   * `stream.pipeline()`.
+   *
+   * This is the form to prefer: holding the stream callback back until stdout
+   * drains is what pushes a slow reader's pace all the way up the pipeline, so a
+   * run costs the memory of what is in flight rather than of the whole result
+   * set.
+   */
+  readonly sink: Writable;
+  /** Marks the end of the output and releases the pipe watcher. */
```
maoberlehner, line 33: Claude:

> `close()` only calls `stopWatching()` (see line 195) — no sentinel, no `end()`, no drain wait. Nothing marks or flushes the output, so the first half of this sentence is not true.

packages/cli/src/lib/pipe/output.ts
```diff
@@ -0,0 +1,294 @@
...
+ * that can act on the first line without waiting for the last — against a
+ * cosmetic outcome this process cannot even reliably detect: whether the command
+ * downstream prints to the same terminal is a property of that command, not of
+ * anything visible from here. Streaming is also what lets `signal` fire early
+ * enough to save work, and what keeps memory flat on a result set of any size.
+ *
+ * What *is* detectable is stdout being a terminal, which means the data itself
+ * is landing on the same screen as the progress bars. That collision is certain
+ * rather than guessed, so the bars are dropped for the run; the summary and
+ * warnings on stderr stay. `2>/dev/null` or `--no-ui-enabled` silences the rest.
```
maoberlehner, line 92: Claude:

> 15 lines of design argument, and ADR-0016 in this PR makes the case at length. Worth keeping only what is not recoverable from the code or the ADR: lines go out as produced, and stdout being a TTY drops the progress bars for the run.

packages/cli/src/lib/pipe/input.ts
```diff
@@ -0,0 +1,128 @@
...
+ * `< /dev/null`, for a file redirect and for a closed descriptor alike — so under
+ * CI, cron or `docker` without `-i`, a command that guessed would silently
+ * operate on zero records and exit 0. And a command that consumes stdin
+ * unasked breaks as the child of a loop sharing that descriptor: in
+ * `cat list.txt | while read x; do cmd; done` it would eat the rest of
+ * `list.txt` and the loop would run once. This is why `ssh` has `-n`.
+ *
+ * `-` is the established convention: `git apply -`, `kubectl apply -f -`,
+ * `docker build -`, `tar -f -`.
+ */
```
maoberlehner, line 21: Claude:

> Same as the streaming docblock in `output.ts` — this is ADR-0016's argument restated. The one fact a future reader cannot recover is the detection problem: `process.stdin.isTTY` is `null` for a pipe, for `< /dev/null`, for a file redirect and for a closed descriptor alike, so detection cannot tell piped input from none. The `ssh -n` aside and the four-tool convention list can go.

packages/cli/src/commands/stories/find/streams.ts
```diff
@@ -0,0 +1,260 @@
...
+ * story-level expression (`$[?($.updated_at > …)]`) decides here too rather than
+ * falling through to a fetch.
+ *
+ * `attachContent` turns the stage into a bulk content *source* rather than a
+ * filter: the content it already holds rides along on the story instead of being
+ * discarded, so a consumer that would otherwise fetch each story individually can
+ * read it here. Off by default, because `find` re-fetches every match from MAPI
+ * and emits that, and CAPI draft content carries editor metadata (`_editable`)
+ * the Management API does not.
+ */
```
maoberlehner, line 115: Claude:

> 26-line docblock, and the settled-here/merge half is restated three more times in this PR: the param docs at lines 130-139, `filterStoriesStream` at lines 51-53, and `pipeline.ts:226-227`.
>
> Worth keeping "prunes only, anything undecided falls through" and the `attachContent` paragraph, and letting the param docs carry the rest. (The sentence at line 101 also needs a correctness fix — separate comment.)

packages/cli/src/commands/stories/find/pipeline.ts
```diff
@@ -0,0 +1,252 @@
...
+              // Whether a failed batch costs the run anything depends on what is
+              // downstream of it. With the per-story MAPI fetch still to come,
+              // its stories pass through undecided and are settled exactly as
+              // they would have been without the flag: the result set is
+              // complete, so the run is a success and must exit 0. A script
+              // doing `find > out.jsonl || exit` would otherwise throw away a
+              // correct answer. Without that fetch (`--skip-content`, or the
+              // reference scan, where CAPI *is* the content source) the stories
+              // in the batch really are decided on less, so it is a real
+              // failure. The summary reports the count either way.
```
maoberlehner, line 194: Claude:

> Ten-line essay inside a callback. One sentence carries it: a failed batch only costs the run when nothing downstream can settle its stories, and with the MAPI fetch still to come the result set is complete, so it is logged rather than raised.

packages/cli/src/commands/stories/query-params.ts
```diff
@@ -0,0 +1,51 @@
...
+ * The scope flags every story-listing command shares.
+ *
+ * `find`, `pull` and `migrations run` narrow the same list endpoint with the same
+ * two flags, `validate` with the first of them, and each used to normalize them
+ * itself — down to the same comment about the leading slash, copied between
+ * files, while `migrations run` did not normalize at all and quietly matched
+ * nothing for a `--starts-with=/en/blog/`. Sibling subcommands must not import
+ * from each other, so the shared half lives here in the parent command
+ * directory, and each command layers its own flags on top.
+ */
```
maoberlehner, line 16: Claude:

> Written from the refactor's perspective — "each used to normalize them itself … copied between files" is changelog, not something a future reader needs.
>
> The two facts worth keeping: which commands share these flags, and that sibling subcommands must not import from each other, so the shared half lives in the parent command directory. The leading-slash rationale is already stated correctly at lines 37-38.

packages/cli/src/commands/stories/find/check-references.ts
```diff
@@ -0,0 +1,269 @@
...
+  capi?: CapiFilter;
+}): Promise<void> {
+  // Said before the scan rather than after it: the gap changes what a clean
+  // report means, and it is cheap to act on while nothing has been read yet.
+  //
+  // CDN content omits every `<field>__i18n__<lang>` key, so a reference held in
+  // a field-level translation is not in the document this scan sees — including
+  // whole nested blocks under a translated `bloks` field. `extractReferences`
+  // matches those fields by their base name when reading MAPI content, and that
+  // handling has nothing to work with here.
```
maoberlehner: Claude:

> The `ui.warn` directly below already says the `__i18n__` omission at length, including the nested-blocks case. Two things here are not in the warn and are worth keeping: why it is said before the scan rather than after, and that `extractReferences` matches translated fields by base name when reading MAPI content. The middle paragraph can go.

packages/cli/src/commands/stories/find/run.ts
```diff
@@ -0,0 +1,172 @@
...
+    const list = tracker.counts("list");
+    const capiFilter = tracker.counts("capiFilter");
+    const content = tracker.counts("content");
+    const filtered = tracker.counts("process");
+    ui.br();
+    ui.info(resultsHeadline({ tracker, filters, skipContent, capi: capi !== undefined }));
+
+    // A deliberate stop, so it reports as one. The counts below describe a
+    // partial scan and would otherwise read as "this is all there was", and
+    // anything less explicit than "not an error" reads as one next to them.
```
maoberlehner, line 86: Claude:

> Paraphrases the `ui.ok` string right below it — the string already says "not an error" and that the counts describe a partial scan. Same at lines 96-99 and 108-111, though those two can keep the clause that is not in the string: undecided stories are dropped on metadata alone once there is no MAPI fetch left.
>
> The warning text at line 100 also needs a factual fix — separate comment.

packages/cli/src/commands/stories/find/actions.ts
```diff
@@ -0,0 +1,304 @@
...
+  return value;
+}
+
+export function buildQueryParams(
+  text: string | undefined,
+  options: FindOptions,
+): StoriesQueryParams {
+  const params: StoriesQueryParams = {};
+
+  // Text search
```
maoberlehner, line 119: Claude:

> Section label restating the assignment below. Same at lines 139, 155, 168 and 181. The neighbouring comments that carry a reason — 124-127, 144-146, 160-166, 173-176 — are worth keeping; these five are not.

packages/cli/src/commands/stories/find/references.ts
```diff
@@ -0,0 +1,211 @@
...
+        targetUuid: attrs.uuid,
+        refType: "richtext",
+        fieldPath: path,
+        cachedUrl: typeof attrs.href === "string" ? attrs.href : undefined,
+      });
+      return;
+    }
+  }
+
+  // Track current component name for relation field lookup
```
maoberlehner, line 118: Claude:

> Restates the line below it. Same at line 152 ("Recurse into object values").

packages/cli/src/commands/stories/find/filters.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Filters a story by publish status (client-side).
+ *
+ * Server-side narrowing (is_published) is applied at the query level.
+ * This filter handles the client-side distinction:
+ * - `published`: is_published=true + unpublished_changes === false
+ * - `changed`: is_published=true + unpublished_changes === true
+ * - `draft`: fully server-side, no client filter needed
+ */
```
maoberlehner, line 14: Claude:

> The bullets restate the switch body verbatim, `return true; // fully server-side filtered` at line 22 restates the third bullet, and the docblock at lines 26-28 restates the function name.
>
> One line covers the whole thing: `draft` is decided entirely server-side, and the other two share `is_published` and are told apart from `unpublished_changes`.

packages/cli/src/lib/pipe/output.ts
```diff
@@ -0,0 +1,294 @@
...
+  const controller = new AbortController();
+  const stopWatching = onStdoutClosed(() => {
+    controller.abort(new DownstreamClosedError());
+  });
+
+  let written = 0;
+  /**
+   * Called once a line is out, which is what makes the stop safe: the run ends
+   * holding exactly the number of results it was asked for, never one short.
+   */
```
maoberlehner, line 140: Claude:

> Same argument as the `limit` param doc at lines 104-111, forty lines up in the same file.

packages/cli/src/lib/pipe/phases.ts
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Renders a phase mark as an elapsed-since-start reading.
+ *
+ * Written as `done @12.3s` rather than a bare duration because the phases
+ * overlap on purpose — content fetches begin while later pages are still
+ * listing. A bare duration would read as three figures that add up to the total,
+ * and listing would look "slow" whenever backpressure held the pager back
+ * waiting on the stage below, which is the pipeline working, not stalling.
+ */
```
maoberlehner, line 67: Claude:

> Nine lines of UX rationale on a one-line formatter. The first sentence plus "phases overlap, so a bare duration would read as figures that sum to the total" is the whole content.

packages/cli/src/commands/stories/find/index.test.ts
```diff
@@ -0,0 +1,238 @@
...
+  // Regression: aborting the pipeline tears every in-flight stage down at once,
+  // and each one reported that teardown through its own error callback. The run
+  // printed "▲ error The operation was aborted", counted a failed listing page,
+  // and exited 1 — for what is a complete, deliberate use of the command.
+  //
+  // One test rather than four, because "the reader hung up" is module state that
+  // survives for the life of the process: a second run in the same file starts
+  // with stdout already closed and aborts before any stage has work in flight,
+  // so it exercises none of this. All four consequences of the one run are
+  // asserted together instead.
```
maoberlehner, line 221: Claude:

> The first half is a transcript of the old bug rather than the invariant being guarded — better to state what must hold: a reader closing the pipe is a deliberate stop, so the run exits 0 and reports no failed stage.
>
> The second half is worth keeping as is. That closed stdout is module state surviving for the life of the process, and therefore why this is one test and not four, is not recoverable from the code.

packages/cli/src/commands/stories/find/streams.test.ts
```diff
@@ -0,0 +1,390 @@
...
+    );
+
+    expect(received.map((story) => story.id)).toEqual([2]);
+  });
+
+  // Regression: the stage ended with `Promise.all(...).finally(() => callback())`,
+  // which ran the callback and then dropped the rejection. A batch that died on
+  // the way to `push` therefore took its stories with it, the stage still ended
+  // cleanly, and the run reported a short result set as a complete one while an
+  // unhandled rejection escaped the process.
```
maoberlehner, line 239: Claude:

> Narrates the old implementation instead of the invariant. What is guarded: a batch that rejects must fail the stage, otherwise a short result set ends cleanly and reads as complete.
>
> Same pattern at `../filter-query.test.ts:36` and `:78`, `actions.test.ts:219`, `references.test.ts:234`, and `../../../lib/pipe/output.test.ts:72`, `:127`, `:211`.

packages/cli/src/commands/stories/find/capi.ts
maoberlehner: Claude:

> Answering your question here: keep it `false`, and if the other behavior is ever wanted, make it an explicit flag rather than a side effect of `--capi-params resolve_relations`.
>
> Inlining breaks the invariant the stage rests on. `--where` would run against a document MAPI does not have, so the flag stops being an accelerator and becomes a different question. `isAlreadyMatched` makes that concrete: a CAPI match is final, so a `@.quote.content.role` expression would emit stories whose emitted `content` has a bare UUID where the predicate matched an object. We have a live example of that exact failure shape from `--capi-params version=published`: a `$[?($.content.title == 'PUBLISHED_MATCH')]` predicate emitting `{"title":"draft_other"}`. A user cannot write one expression that means the same thing with and without the flag.
>
> `resolve_relations` also caps how much it resolves per request, so past the cap some values come back as objects and some as UUIDs inside the same batch, and the filter prunes on the difference with nothing in the run to show it. A pruning stage must not be able to be wrong invisibly.
>
> And cost: 24 in-flight batches times 25 stories with relations inlined puts the memory high-water mark at the mercy of fan-out rather than story size.
>
> Your two-pass example is the honest trade and the right one for now — `find` answers "which stories", joining two answers is the pipe's job, which is what ADR-0016 is for. If the single-statement form turns out to be real demand, the shape worth having is a flag that inlines relations *and* forces those stories through as undecided (`onUnresolved`), so the CDN read stays an accelerator and the MAPI pass stays authoritative.
>
> Worth adding to the README regardless: the `rels` sidecar is dropped today, so a relation's content is not merely awkward to filter on, it is unreachable in a single pass. That is the actual limitation and the reason the two-pass example exists.

maoberlehner: I don't think we should enable it. It makes sense in the case you mention, but I can also imagine cases where it would not make sense and would come as a surprise. E.g., it would find a word in the bio of an author's story when you suspect you should only search for it in blog articles.

I think in the context of this command, not resolving is the right call. It's not about what's rendered but actual _stories_, so I'd expect to find the actual stories that contain a keyword, not also stories referencing stories containing it.

packages/cli/src/commands/stories/find/command.ts
maoberlehner: Our naming is confusing... I agree with Claude here.

Another complicating layer for the block/component naming thing is that we need to consider what LLMs have in their training data (and humans, too), so maybe we should consider aliases.

maoberlehner, line 37: Is `(server-side)` here and in other help texts added on purpose, or is this Claude injecting context from the conversation?

maoberlehner (changes requested):
Fyi: as Opus 5 tends to go wild on comments, I ran a dedicated comment reviewer on some of the more extensive ones.

packages/cli/src/commands/stories/find/README.md
```diff
@@ -0,0 +1,1077 @@
+---
```
alexjoverm, line 1: BLOCK - just a comment to prevent merging till this is moved to docs platform

packages/cli/src/commands/stories/find/capi.ts
alexjoverm: Perfect, let's leave it like that then!

PR #773, opened 2026-08-20
fix(capi-client): treat HTTP error results as failures in the cache strategies

maoberlehner:
Both cache strategies that react to a failed request only ever saw *thrown* errors. With `throwOnError` disabled — the default — an HTTP error is not a rejection but a resolved `ApiResponse` carrying `error`, so the most common failure of all was invisible to both of them.

Four user-visible consequences, one root cause, one commit each. The last two came out of review.

## 1. `network-first` did not fall back on an HTTP error

It is documented as "try the network, fall back to cached data if available." The fallback lived entirely in a `catch`:

```ts
try {
  return await loadNetwork();   // a 503 resolves here, it does not throw
} catch (error) {
  if (cachedResult !== undefined) return cachedResult;
  throw error;
}
```

So the origin answering 5xx surfaced straight to the caller while a usable cached entry sat unused. Only a transport-level failure — DNS, connection reset, `fetch` rejecting — ever triggered the fallback. That inverts the intent: the case `network-first` exists to survive is a struggling origin, and that is the case it did not handle.

| scenario | before | after |
|---|---|---|
| transport error, warm cache | serves cached | serves cached |
| HTTP 503, warm cache | **surfaces the 503** | serves cached |

## 2. `swr` silently swallowed a failed background revalidation

`onRevalidationError` fired only from the `catch` of the background promise, so a revalidation that came back 5xx was dropped without a word. The cached entry went on being served with no signal that it had stopped being refreshed — the one thing the callback exists to tell you.

| scenario | before | after |
|---|---|---|
| revalidation throws | `onRevalidationError` | `onRevalidationError` |
| revalidation returns 503 | **silence** | `onRevalidationError` |

Reporting happens off the background promise, so swr stays non-blocking.

## 3. …and the same rule did not hold for rejections

Fixing (1) on the resolved path left the `catch` untouched, still falling back for anything thrown. With `throwOnError: true` the same 404 rejects instead of resolving, so whether a definitive answer surfaced came down to a client-wide flag:

| `throwOnError` | 404 with a warm cache |
|---|---|
| `false` | the 404 surfaces |
| `true` | **the cached copy is served, the deletion masked** |

A rejection is now classified the same way a resolved failure is. Reproduced first, then fixed.

## 4. `swr` could report one failure twice, then crash the process

The revalidation chain was `.then(report).catch(report)`. A callback that threw while reporting a resolved failure was immediately re-entered by the `.catch` — called a second time, with its own failure in place of the one it was reporting. If that second call threw too, the chain ended rejected with nothing awaiting it: an unhandled rejection, fatal on current Node defaults.

The handlers are now paired in a single `.then(onFulfilled, onRejected)`, so the rejection handler sees only what `loadNetwork` rejected with and cannot be reached from the fulfillment path at all. Both are total: `getFailure` and `onRevalidationError` are caller-supplied and run in a chain nobody awaits, so anything they throw is logged and dropped rather than left to escape.

## How

The strategies live in `utils/cache.ts` and are generic over `TData`, so they cannot inspect an `ApiResponse` themselves. `StrategyContext` gained two optional reporters, supplied by the client and shared by the fixes — one per way a request can fail:

```ts
getFailure?: (value: TData) => { transient: boolean; error: unknown } | undefined;
getThrownFailure?: (error: unknown) => { transient: boolean; error: unknown };
```

Keeping both in the client leaves the strategies unaware of `ClientError`. Optional is load-bearing: `cache.strategy` accepts a caller-supplied handler, and a context without the reporters behaves exactly as before — every resolved value a success, every rejection transient, which is what a transport-level failure is.

The two strategies read them differently, on purpose:

- **`network-first` falls back only on `transient`.** Serving a cached copy for a definitive answer would be wrong.
- **`swr` reports every failure.** A revalidation that 404s is worth surfacing; the caller decides what to act on.

| status | `network-first` | `swr` |
|---|---|---|
| 5xx, 429, 408, 413 | serve cached | report |
| 404 | return error — the story really is gone | report |
| 401, 403 | return error — the token really is bad | report |
| 400, 422 | return error — the request is wrong | report |

Same verdict whether the failure resolved or was thrown.

Every 5xx counts as transient: a server that fails to answer says nothing about whether the resource still exists, so a cached copy is the best available answer. The three 4xx exceptions are the ones the origin uses to ask for the same request again — a timeout, a payload it wants re-sent, and a rate limit.

`cache-first` is untouched, and covered by a test so a future change has to be deliberate: it never reaches the network while a valid entry exists, so an error there means the entry was already gone and there is nothing to fall back to.

## Verification

```
api-client:  196 passed  (149 on main, +47 here)
js-client:   194 passed | 1 skipped
lint, types: clean
```

Coverage was added at both levels, because `network-first` and swr's error path had **no integration coverage at all** — the only existing tests were three unit tests against the strategy functions.

- unit: transient error result falls back, definitive does not, a context without the reporter is unchanged, `isTransientStatus` over 13 statuses, swr reports a resolved transient failure, a resolved definitive failure, and stays quiet on success
- unit, from review: a throwing `onRevalidationError` is called once — on the resolved path and on the rejected one — and a throwing `getFailure` cannot end the revalidation as an unhandled rejection
- integration through `createApiClient` and msw: `500 / 502 / 503 / 429 / 408` each serve the cached response; `400 / 401 / 403 / 404 / 422` each surface the error; a transient error with an empty cache still surfaces; `cache-first` does not fall back; a swr revalidation returning 503 reaches `onRevalidationError` without blocking the read
- integration, from review: the same ten statuses again under `throwOnError: true`, where they reject rather than resolve, plus a transport-level failure with no status at all

Every new test was confirmed to fail with its fix reverted and pass with it in place. All four commits are independently green.

### DIFF
--- a/packages/capi-client/src/utils/cache.ts
+++ b/packages/capi-client/src/utils/cache.ts
@@ -17,6 +17,18 @@ interface StrategyContext<TData> {
    * custom strategy handlers and callers that do not supply it keep working.
    */
   getFailure?: (value: TData) => StrategyFailure | undefined;
+  /**
+   * Describes the failure a rejection from `loadNetwork` represents.
+   *
+   * The mirror of {@link StrategyContext.getFailure} for the other way a request can
+   * fail: with `throwOnError` enabled an HTTP error rejects rather than resolving, so a
+   * definitive answer reaches a strategy as a thrown value and has to be told apart from
+   * a transport-level failure the same way.
+   *
+   * Optional: a context without it behaves as if every rejection were transient, which is
+   * what a transport-level failure is.
+   */
+  getThrownFailure?: (error: unknown) => StrategyFailure;
 }
 
 export interface StrategyFailure {
@@ -99,9 +111,12 @@ export const createMemoryCacheProvider = (
 };
 
 /**
- * Statuses that mean "try again later" rather than "here is your answer". Mirrors the set
- * the client retries, so one of these only reaches a strategy once the retries are spent
- * and the failure is a real outage.
+ * Statuses that mean "try again later" rather than "here is your answer".
+ *
+ * Every 5xx counts: the server failing to answer says nothing about whether the resource
+ * exists, so a cached copy is still the best available answer. The 4xx exceptions are the
+ * three the origin uses to ask for the same request again — a timeout, a payload it wants
+ * re-sent, and a rate limit.
  */
 const TRANSIENT_STATUS_CODES = new Set([408, 413, 429]);
 
@@ -120,7 +135,12 @@ export const createCacheFirstStrategy = (): CacheStrategyHandler => {
 };
 
 export const createNetworkFirstStrategy = (): CacheStrategyHandler => {
-  return async <TData>({ cachedResult, loadNetwork, getFailure }: StrategyContext<TData>) => {
+  return async <TData>({
+    cachedResult,
+    loadNetwork,
+    getFailure,
+    getThrownFailure,
+  }: StrategyContext<TData>) => {
     try {
       const result = await loadNetwork();
 
@@ -135,8 +155,12 @@ export const createNetworkFirstStrategy = (): CacheStrategyHandler => {
 
       return result;
     } catch (error) {
-      // network-first: try network, fall back to cached data if available.
-      if (cachedResult !== undefined) {
+      // The rule the resolved path applies has to hold here too: with `throwOnError`
+      // enabled the same 404 rejects instead of resolving, and serving a cached copy
+      // would mask the deletion just as thoroughly. Without a reporter every rejection
+      // counts as transient — a context that cannot classify only ever sees the
+      // transport-level failures, which are.
+      if (cachedResult !== undefined && getThrownFailure?.(error).transient !== false) {
         return cachedResult;
       }
 
--- a/packages/capi-client/src/utils/cache.ts
+++ b/packages/capi-client/src/utils/cache.ts
@@ -4,6 +4,26 @@ interface StrategyContext<TData> {
   key: string;
   cachedResult: TData | undefined;
   loadNetwork: () => Promise<TData>;
+  /**
+   * Describes the failure a value `loadNetwork` resolved with represents, or `undefined`
+   * when it succeeded.
+   *
+   * A failed request does not always reject: with `throwOnError` disabled — the default —
+   * an HTTP error arrives as a resolved response carrying `error`. A strategy that reacts
+   * to failure has to recognise those, and `network-first` also has to tell a temporary
+   * failure apart from a definitive one.
+   *
+   * Optional: a context without it behaves as if every resolved value succeeded, so
+   * custom strategy handlers and callers that do not supply it keep working.
+   */
+  getFailure?: (value: TData) => StrategyFailure | undefined;
+}
+
+export interface StrategyFailure {
+  /** `true` for a failure worth retrying — an outage rather than an answer. */
+  transient: boolean;
+  /** The failure itself, for reporting. */
+  error: unknown;
 }
 
 export type CacheStrategyHandler = <TData>(context: StrategyContext<TData>) => Promise<TData>;
@@ -78,6 +98,17 @@ export const createMemoryCacheProvider = (
   };
 };
 
+/**
+ * Statuses that mean "try again later" rather than "here is your answer". Mirrors the set
+ * the client retries, so one of these only reaches a strategy once the retries are spent
+ * and the failure is a real outage.
+ */
+const TRANSIENT_STATUS_CODES = new Set([408, 413, 429]);
+
+/** Returns `true` for a status that reflects a temporary failure rather than an answer. */
+export const isTransientStatus = (status: number): boolean =>
+  status >= 500 || TRANSIENT_STATUS_CODES.has(status);
+
 export const createCacheFirstStrategy = (): CacheStrategyHandler => {
   return async <TData>({ cachedResult, loadNetwork }: StrategyContext<TData>) => {
     if (cachedResult !== undefined) {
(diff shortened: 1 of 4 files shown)


alexjoverm (changes requested):
### Major: `throwOnError: true` still falls back to the cache on a definitive error

**File:** `packages/capi-client/src/utils/cache.ts:137-144`

The PR's central rule is "fall back only on a transient failure; a 404 or 401 is an answer, not an outage." That rule is implemented on the *resolved* path only. The `catch` path is untouched and still falls back for **any** thrown error:

```ts
} catch (error) {
  // network-first: try network, fall back to cached data if available.
  if (cachedResult !== undefined) {
    return cachedResult;
  }
  throw error;
}
```

With `throwOnError: true` an HTTP error does not resolve — `client.gen.ts:95-97` (`if (opts.throwOnError) throw finalError`) rejects with a `ClientError`. So the exact case the PR argues must never be masked *is* masked, purely depending on a client flag:

| client | 404 with a warm cache |
| --- | --- |
| `throwOnError: false` | error surfaces (this PR) |
| `throwOnError: true` | **cached copy returned, deletion masked** |

`ClientError` already carries what is needed (`error.response.status`, `src/error.ts:17`).

**Suggestion:**

```ts
} catch (error) {
  // A thrown ClientError is an HTTP answer (throwOnError: true); anything else is a
  // transport-level failure and is always transient.
  const isDefinitive = error instanceof ClientError && !isTransientStatus(error.response.status);
  if (cachedResult !== undefined && !isDefinitive) {
    return cachedResult;
  }
  throw error;
}
```

`cache.ts` importing `ClientError` would couple the util to the client, so the cleaner shape is to widen the reporter to cover thrown values too (`getFailure` for resolved values + an equivalent for rejections), or move the classification behind a single injected `classifyFailure`. Either way this needs a test: `throwOnError: true`, warm cache, 404 → rejects.

### Minor: the transient set does not match the comment

**File:** `packages/capi-client/src/utils/cache.ts:101-110`

The comment says the set "mirrors the set the client retries", but ky retries `408, 413, 429, 500, 502, 503, 504` while `isTransientStatus` accepts *every* 5xx. `501 Not Implemented` and `505 HTTP Version Not Supported` are definitive answers, are never retried, and now silently serve stale content. Either narrow the set to the retried statuses or drop the "mirrors" claim and justify `>= 500` on its own terms (defensible — most 5xx are outages — but then the comment is wrong).

alexjoverm (approved):

packages/capi-client/src/utils/cache.ts
```diff
@@ -116,13 +182,22 @@ export const createSwrStrategy = (options: SwrStrategyOptions = {}): CacheStrate
...
+            // only from `catch` would drop it silently and leave the entry to go stale
+            // with no signal at all. Every failure reaches the callback, transient or
+            // not: the caller decides what is worth acting on.
+            const failure = getFailure?.(value);
+            if (failure !== undefined) {
+              onRevalidationError(failure.error);
+            }
+          })
           .catch((error: unknown) => {
             onRevalidationError(error);
```
dipankarmaikap: **Bug: `.catch()` intercepts exceptions thrown by `onRevalidationError` in `.then()`**

The chain is `.then(fn).catch(handler)`. If `fn` calls `onRevalidationError` and that callback throws, `.catch(handler)` catches the exception and calls `onRevalidationError` a second time — with the error thrown by the first invocation, not with the original network error. If the second call also throws, the resulting rejected promise is unhandled (nobody awaits the revalidation promise stored in the map).

Confirmed by probe: a throwing `onRevalidationError` was called **2 times** (expected 1), and Vitest captured an unhandled rejection at `cache.ts:200`.

Suggested fix — two-argument `.then()` keeps the paths isolated:

```ts
loadNetwork()
  .then(
    (value) => {
      const failure = getFailure?.(value);
      if (failure !== undefined) onRevalidationError(failure.error);
    },
    (error: unknown) => {
      // rejection-only path — a throwing onRevalidationError in the
      // fulfillment handler above cannot re-enter here
      onRevalidationError(error);
    },
  )
  .finally(() => {
    revalidations.delete(key);
  });
```

dipankarmaikap (commented):
## AI Review — Findings

Reviewed the three cache strategy changes across `cache.ts` and `client.ts`. Ran targeted probes against the live code to distinguish actual bugs from test gaps.

---

### 🐛 Bug confirmed: SWR `onRevalidationError` double-invocation + unhandled rejection

The new `.then()` handler calls `onRevalidationError`. If that callback throws, the `.catch()` below it catches the exception and calls `onRevalidationError` a second time with the wrong error. If that second call also throws, `.finally()` still clears the key but the promise chain ends as a rejected promise that nobody awaits — unhandled rejection in Node.

Reproduced with a probe test: expected 1 invocation, got 2. Vitest also captured the unhandled rejection directly.

**Fix:** use the two-argument form of `.then()` so the rejection handler is isolated from the fulfillment handler:

```ts
.then(
  (value) => {
    const failure = getFailure?.(value);
    if (failure !== undefined) onRevalidationError(failure.error);
  },
  (error: unknown) => {
    onRevalidationError(error); // only sees loadNetwork() rejections, never .then() throws
  },
)
```

See inline comment for the exact location.

---

### ✅ Not a bug: `network-first` with no `getFailure` + failure-shaped resolved value

When `getFailure` is omitted, `getFailure?.(result)?.transient === true` evaluates to `undefined === true` → `false`. The network result is returned as-is. Probe passed. Code is correct — test coverage was just absent for this specific combination.

---

### ✅ Not a bug: SWR fires `onRevalidationError` on every request during a persistent outage

`.finally()` always clears the in-flight key, so each subsequent SWR request starts a fresh revalidation cycle. During a persistent server failure, `onRevalidationError` fires on every request. Probe confirmed this is the actual behaviour. It is intentional — the cache keeps serving stale data while notifying the caller continuously — but it is undocumented.

packages/capi-client/src/utils/cache.ts
maoberlehner: thanks! Fixed now!

PR #774, opened 2026-08-20
chore(react): enable rc pre-release channel

dipankarmaikap:
## Summary

Enables the `rc` pre-release channel for `@storyblok/react`.

## Changes

- **`tools/release.mjs`** — adds `"rc"` to `RELEASE_BRANCHES` so `pnpm release` can be run from an `rc` branch
- **`packages/react/package.json`** — adds `{ "name": "rc", "prerelease": true }` to `release.branches` so the package is eligible for versioning and publishing from the `rc` branch

## How to use after merge

```bash
git checkout -b rc
# make changes with conventional commits
pnpm release
# GitHub Actions → Publish → Run workflow → select branch: rc
```

Users install the RC with:
```bash
npm install @storyblok/react@rc
```

Versions will follow the pattern `7.4.0-rc.0`, `7.4.0-rc.1`, etc.

Fixes DX-554

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -75,7 +75,7 @@ pnpm release
 
 This script will:
 
-- Check that you're on a release branch (`main`, `alpha`, `beta`, or `next`)
+- Check that you're on a release branch (`main`, `alpha`, `beta`, `next`, or `rc`)
 - Check for uncommitted changes
 - Fetch the latest changes from remote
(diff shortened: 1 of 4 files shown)


maoberlehner (changes requested):
I think we need to add `rc` info also to `nx.json` and  `RELEASING.md`

maoberlehner (approved):
LGTM

PR #775, opened 2026-08-20
fix(cli): remove Eleventy template support

dipankarmaikap:
## What

Removes Eleventy as a supported starter template from the `storyblok create` command.

## Why

The `blueprint-core-eleventy` GitHub repository has been archived and Eleventy is no longer supported as a CLI starter template.

## Changes

- **`constants.ts`**: Removed `ELEVENTY` from the `TemplateTechnology` union and from the `templates` record (affects the offline fallback list)
- **`actions.ts`**: Filter out archived repos from the live GitHub blueprint fetch — the GitHub search API includes archived repos by default, so `blueprint-core-eleventy` was still appearing in the interactive picker
- **`actions.test.ts`**: Added a test asserting archived repos are excluded from the template list
- **`README.md`**: Removed all Eleventy references from the create command docs

## Testing

```
pnpm nx test storyblok --run   # 1286 passed
pnpm nx lint:fix storyblok     # 0 errors
```

### DIFF
--- a/packages/cli/src/commands/create/README.md
+++ b/packages/cli/src/commands/create/README.md
@@ -20,23 +20,22 @@ The command will:
 
 ## Options
 
-| Option                      | Description                                                                   | Default               |
-| --------------------------- | ----------------------------------------------------------------------------- | --------------------- |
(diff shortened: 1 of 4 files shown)


maoberlehner (approved):
LGTM

PR #776, opened 2026-08-21
fix(nuxt): first bug fixes after audit

JakubAndrzejewskiSB:
## packages/nuxt/src/runtime/plugin.ts
Removed the JSON.parse(JSON.stringify(...)) workaround around useRuntimeConfig().public.storyblok. This pattern bypassed Nuxt's SSR read-only config restriction, broke Vue reactivity, stripped non-serializable values, and threw a SyntaxError if storyblok was undefined. Now reads the config directly.

## packages/nuxt/src/module.ts
Added a fail-fast check in setup(): if enableServerClient: true and accessToken is missing, the module now throws immediately at setup time instead of only failing lazily on the first server request. Also added a clarifying comment on the else branch documenting that exposing the token in runtimeConfig.public when enableServerClient: false is intentional (needed client-side for the CDN API plugin).

## packages/nuxt/src/runtime/composables/useAsyncStoryblok.ts
Fixed a bug where passing bridge: false didn't actually disable bridge registration (the option type didn't even allow false). bridge now accepts StoryblokBridgeConfigV2 | false, and bridge setup is skipped entirely when explicitly disabled.

## packages/nuxt/src/runtime/server/index.ts
serverStoryblokClient now throws an H3 createError() (500, with a statusMessage/message) instead of a plain JS Error when accessToken is missing, so Nitro handles it as a proper HTTP error response instead of an uncaught exception.


maoberlehner:
fyi in monoblok, the conventional commit format is used in the release process, so it is important that we use e.g. `fix(nuxt)` in the PR title.

JakubAndrzejewskiSB:
> fyi in monoblok, the conventional commit format is used in the release process, so it is important that we use e.g. `fix(nuxt)` in the PR title.

Thanks Markus! I have updated the title :)

packages/nuxt/src/module.ts
```diff
@@ -48,6 +48,12 @@ export default defineNuxtModule<AllModuleOptions>({
     addImportsDir(resolver.resolve("./runtime/composables"));
 
     if (options.enableServerClient) {
+      if (!options.accessToken) {
```
maoberlehner: What happens if `NUXT_STORYBLOK_ACCESS_TOKEN` is supplied at runtime? Does it still work?

packages/nuxt/src/runtime/server/index.ts
```diff
@@ -9,9 +9,11 @@ export const serverStoryblokClient = (event: H3Event) => {
   const { apiOptions = {} } = config.public.storyblok;
 
   if (!accessToken) {
-    throw new Error(
-      `Storyblok access token is not configured. Make sure to set storyblok.accessToken in your nuxt.config.ts and enable storyblok.enableServerClient = true to use the server-side client.`,
-    );
+    throw createError({
```
maoberlehner: If I understand this correctly, using `createError` here means that now users will see an error page in the browser `Storyblok access token is not configured.`, and the server log will show nothing. IMO, that's a regression from the current behavior of  `Error 500, Server Error`. Regular users should not see anything about `accessToken` and an error like this should be logged to the server log so it can be detected.

I think adding `fatal: true` and a `cause` should do the trick:

```ts
throw createError({
  statusCode: 500,
  statusMessage: "Internal Server Error",
  fatal: true,
  cause: new Error("Storyblok access token is not configured. Make sure to set storyblok.accessToken in your nuxt.config.ts and enable storyblok.enableServerClient = true to use the server-side client."),
});
```

maoberlehner (changes requested):
Thanks @JakubAndrzejewskiSB for taking the time to improve our Nuxt client! Two findings to double-check for you.

packages/nuxt/src/module.ts
JakubAndrzejewskiSB: Good catch! Added NUXT_STORYBLOK_ACCESS_TOKEN to if statement there.

Also moved the whole if statement higher to avoid situation where someone will use the module and will not pass the env variable.

packages/nuxt/src/runtime/server/index.ts
JakubAndrzejewskiSB: As the same check was in the module.ts (higher in the sdk hierarchy) I decided to remove this completely. IF the token is missing, the core module.ts file will discover it and throw error earlier

packages/nuxt/package.json
```diff
@@ -31,7 +31,7 @@
   },
   "scripts": {
     "prepack": "nuxt-module-build build",
-    "build": "nuxt-module-build prepare && nuxt-module-build build",
+    "build": "NUXT_STORYBLOK_ACCESS_TOKEN=123 nuxt-module-build prepare && nuxt-module-build build",
```
maoberlehner: Was this added here on purpose?

JakubAndrzejewskiSB: This was added because the if statement that throws error if the token is missing is also being triggered during the build process in CI. I could also add it in the ci scripts if you prefer that way. This assures that the building and running the module will fail if the access token is not provided.

I am also open for suggestions if that is the right approach in your opinion :)

maoberlehner: Hmm, I don't think this is the right approach. Typically, you wouldn't provide the access token to the CI/build environment, but only during runtime (except for `generate` mode). Is there a reason why we need it earlier / already at build time?

JakubAndrzejewskiSB: Updated the code based on our discussion. Let me know if now it is ok :)

maoberlehner (approved):
LGTM!

PR #779, opened 2026-08-21
fix(js-client): keep fetchOptions on a retried request

viict:
`cacheResponse` retries a request after a 429 response. It does not give `fetchOptions` to the retry. The retried request loses the fetch configuration that the caller supplied. The caller cannot see this, because the retry is internal.

## Change

Give `fetchOptions` to the recursive `cacheResponse` call.

## Test

A new test sends `fetchOptions`, makes the first attempt fail with a 429, then makes the second attempt succeed. It shows that the client applies the same options on the two attempts.

- With the change: 67 tests pass, 1 test is skipped, no type errors.
- Without the change: the new test fails.

### DIFF
--- a/packages/js-client/src/index.test.ts
+++ b/packages/js-client/src/index.test.ts
@@ -278,6 +278,51 @@ describe("storyblokClient", () => {
       expect(mockGet).toHaveBeenCalledTimes(2);
     });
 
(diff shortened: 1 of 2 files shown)


maoberlehner (approved):
Great contribution! Fun fact: fixed this in another, larger branch, but it still needs some time, and your tests are better!

PR #782, opened 2026-08-24
fix(live-preview): rework bridge loading, add destroy, and improve Angular DX

dipankarmaikap:
## Summary

Reworks `@storyblok/live-preview` bridge loading and editor event handling, scoped to the live-preview and angular packages only.

### `@storyblok/live-preview`

- **Remove singleton pattern** — `loadStoryblokBridge` now returns a fresh instance on each call. ES module cache deduplicates the underlying import, so no extra network requests.
- **Server-side guard** — throws a clear error when called outside a browser (`isBrowser` check replaces the old `canUseStoryblokBridge` util).
- **Backward-compat shims** — exposes `window.StoryblokBridge` and `window.storyblokRegisterEvent` via `Object.defineProperty` get+set descriptors so legacy consumers and other bridge loaders (`@storyblok/js`, legacy CDN bundle) can co-exist without a `TypeError`. Accessing either global emits a deprecation warning. These globals will be removed in a future major version.
- **Export `isBrowser`** — added to the package public index.
- **Delete `canUseStoryblokBridge.ts`** — logic absorbed inline.
- **Bump `@storyblok/preview-bridge` to `^2.3.0`** — picks up the new `destroy()` method.
- **Fix concurrent subscriptions** — `onStoryblokEditorEvent` forces `initOnlyOnce: false` so every call gets its own fully-initialised bridge with its own `window` message listener. The bridge default (`true`) silently skips `addMessageListener()` when a `.storyblok__hint` element is already in the DOM, making any second concurrent subscription permanently deaf to editor events.
- **Idempotent cleanup** — the returned cleanup function guards with `!active` so calling it twice does not invoke `bridge.destroy()` a second time.
- **Updated tests** — setter acceptance, read-after-set has no deprecation warning, skip-if-defined, stable identity, `initOnlyOnce` override, idempotent cleanup, reload silenced after cleanup, null/undefined/unknown-action/missing-story event guards.

### `@storyblok/angular`

- **`LivePreviewService.connect()`** — new preferred API for component use. Accepts a `DestroyRef` and wires bridge cleanup automatically, removing all boilerplate from the call site:

  ```ts
  // Before
  private cleanupLivePreview: (() => void) | undefined;
  async ngOnInit(): Promise<void> {
    this.cleanupLivePreview = await this.livePreview.listen(...);
  }
  ngOnDestroy(): void { this.cleanupLivePreview?.(); }

  // After
  private readonly destroyRef = inject(DestroyRef);
  ngOnInit(): void {
    this.livePreview.connect(
      (story) => this.story.set(story),
      this.destroyRef,
      this.bridgeConfig,
    );
  }
  ```

  `connect()` registers `DestroyRef.onDestroy` synchronously before the async bridge load, so if the component is destroyed while the bridge is still loading, the bridge is torn down as soon as the promise resolves — no leaks.

- **`connect()` rejection handled** — a `.catch()` surfaces errors (e.g. `LivePreviewNotEnabledError` in dev mode) via `console.error` instead of leaving an unhandled promise rejection.
- **JSDoc example fixed** — `inject(DestroyRef)` is only valid in an injection context; the previous example called it inside `ngOnInit()` which throws `NG0203` at runtime. Updated to show `DestroyRef` injected as a class field.
- **`listen()` is kept as-is** — valid for manual control, RxJS composition, or non-component contexts. JSDoc links to `connect()` as the preferred alternative.
- **Playground components updated** — `CatchAllComponent` and `HomeComponent` now use `connect()` with no manual cleanup boilerplate.
- **Tests** — fully covers `listen()` happy path, `connect()` API including race condition and rejection, production-mode no-op, and the NG0911 SSR race where `DestroyRef.onDestroy()` throws before the bridge resolves.

## Docs

https://github.com/storyblok/storyblok-docs-platform/pull/642

Fixes DX-515

### DIFF
--- a/packages/angular/src/lib/livepreview/livepreview.service.ts
+++ b/packages/angular/src/lib/livepreview/livepreview.service.ts
@@ -1,4 +1,4 @@
-import { InjectionToken, inject, NgZone, Injectable } from "@angular/core";
+import { InjectionToken, inject, NgZone, Injectable, DestroyRef } from "@angular/core";
 import { Story } from "@storyblok/api-client";
 import { type BridgeParams, onStoryblokEditorEvent } from "@storyblok/live-preview";
 
@@ -56,6 +56,17 @@ export class LivePreviewService {
 
   private readonly baseConfig = inject(LIVE_PREVIEW_CONFIG, { optional: true }) ?? {};
 
+  /**
+   * Subscribes to Storyblok Visual Editor live preview updates.
+   *
+   * Returns a cleanup function that destroys the bridge when called.
+   * For automatic cleanup tied to a component or service lifetime, prefer
+   * {@link connect} which accepts a `DestroyRef` and handles teardown for you.
+   *
+   * @param callback Called with the updated story on every `input` event.
+   * @param options Optional bridge configuration; merged over the base config.
+   * @returns A promise that resolves to a cleanup function.
+   */
   async listen(callback: LivePreviewCallback, options?: BridgeParams): Promise<() => void> {
     if (!this.enabledFlag) {
       if (typeof ngDevMode === "undefined" || ngDevMode) {
@@ -74,5 +85,50 @@ export class LivePreviewService {
         callback(story as Story);
       });
     }, mergedConfig);
+  }
+
+  /**
+   * Subscribes to Storyblok Visual Editor live preview updates and
+   * automatically destroys the bridge when the provided `DestroyRef` fires.
+   *
+   * This is the preferred API for component use. It eliminates the need for
+   * a manual cleanup field and an `ngOnDestroy` implementation, and correctly
+   * handles the case where the component is destroyed while the bridge is
+   * still loading.
+   *
+   * @example
+   * ```ts
+   * ngOnInit(): void {
+   *   this.livePreview.connect(
+   *     (story) => this.story.set(story),
+   *     inject(DestroyRef),
+   *     this.bridgeConfig,
+   *   );
+   * }
+   * ```
+   *
+   * @param callback Called with the updated story on every `input` event.
+   * @param destroyRef The `DestroyRef` of the calling component or service.
+   * @param options Optional bridge configuration; merged over the base config.
+   */
+  connect(callback: LivePreviewCallback, destroyRef: DestroyRef, options?: BridgeParams): void {
+    let cleanup: (() => void) | undefined;
+    let destroyed = false;
+
+    // Register teardown synchronously — before the async bridge load — so
+    // that if the component is destroyed while the bridge is still loading,
+    // `destroyed` is set to true and the bridge is torn down as soon as the
+    // promise resolves.
+    destroyRef.onDestroy(() => {
+      destroyed = true;
+      cleanup?.();
+    });
+
+    this.listen(callback, options).then((fn) => {
+      cleanup = fn;
+      if (destroyed) {
+        fn();
+      }
+    });
   }
 }
--- a/packages/angular/src/lib/livepreview/livepreview.service.ts
+++ b/packages/angular/src/lib/livepreview/livepreview.service.ts
@@ -98,10 +98,12 @@ export class LivePreviewService {
    *
    * @example
    * ```ts
+   * private readonly destroyRef = inject(DestroyRef);
+   *
    * ngOnInit(): void {
    *   this.livePreview.connect(
    *     (story) => this.story.set(story),
-   *     inject(DestroyRef),
+   *     this.destroyRef,
    *     this.bridgeConfig,
    *   );
    * }
@@ -133,11 +135,15 @@ export class LivePreviewService {
       destroyed = true;
     }
 
-    this.listen(callback, options).then((fn) => {
-      cleanup = fn;
-      if (destroyed) {
-        fn();
-      }
-    });
+    this.listen(callback, options)
+      .then((fn) => {
+        cleanup = fn;
+        if (destroyed) {
+          fn();
+        }
+      })
+      .catch((err: unknown) => {
+        console.error("[Storyblok] connect() failed to subscribe to live preview updates:", err);
+      });
   }
 }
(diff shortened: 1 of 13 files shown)


packages/angular/src/lib/livepreview/livepreview.service.ts
```diff
@@ -75,4 +86,49 @@ export class LivePreviewService {
...
+    // Register teardown synchronously — before the async bridge load — so
+    // that if the component is destroyed while the bridge is still loading,
+    // `destroyed` is set to true and the bridge is torn down as soon as the
+    // promise resolves.
+    destroyRef.onDestroy(() => {
+      destroyed = true;
+      cleanup?.();
+    });
+
+    this.listen(callback, options).then((fn) => {
```
maoberlehner: nitpick: If this throws, we never handle it.

packages/angular/src/lib/livepreview/livepreview.service.ts
```diff
@@ -75,4 +86,49 @@ export class LivePreviewService {
...
+   * still loading.
+   *
+   * @example
+   * ```ts
+   * ngOnInit(): void {
+   *   this.livePreview.connect(
+   *     (story) => this.story.set(story),
+   *     inject(DestroyRef),
+   *     this.bridgeConfig,
+   *   );
```
maoberlehner, line 108: Claude tells me this: `The documented example calls inject(DestroyRef) inside ngOnInit(), which is outside Angular's injection context.`

My Angular knowledge is basically non-existent so I can't vouch for it but might be an easy fix.

packages/live-preview/src/loadStoryblokBridge.ts
```diff
@@ -1,41 +1,69 @@
...
-): boolean {
-  return JSON.stringify(config1) === JSON.stringify(config2);
+import { isBrowser } from "./utils/isBrowser";
+import { isInEditor } from "./utils/isInEditor";
+
+declare global {
+  interface Window {
+    storyblokRegisterEvent: (cb: () => void) => void;
+    StoryblokBridge: new (options?: BridgeParams) => StoryblokBridge;
+  }
```
maoberlehner, line 13: Should we make them the same as `packages/js/src/types/index.ts`?

maoberlehner:
Let Claude do some testing and it found the following:

| #   | What was done                                                                                                            | Expected                       | Actually happened                                                                                                                                                                                                |
| --- | ------------------------------------------------------------------------------------------------------------------------ | ------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1   | Subscribed twice on one page (`onStoryblokEditorEvent(cbA)`, then `cbB`), then sent one `input` message from the editor  | Both callbacks fire once       | `cbA` fired once, **`cbB` never fired**. No error thrown. The 2nd bridge's `init()` short-circuits on the `.storyblok__hint` element the 1st one left in the DOM, so it never attaches a `message` listener      |
| 2   | With both subscriptions live, called `cleanupA()`, then sent another `input`                                             | `cbB` keeps receiving updates  | **`cbB` fired 0 times**, and `.storyblok__hint` + `#storyblok-bridge-stylesheet` were **removed from the DOM**. `destroy()` dropped the shared refcount to 0 and tore down the overlay the whole page depends on |
| 3   | Started both subscriptions in one `Promise.all([...])`, then sent `input`                                                | Both fire                      | `cbA` once, **`cbB` never** — same defect, reached the way a framework would actually reach it                                                                                                                   |
| 4   | Subscribed, called cleanup, subscribed again (component remount), sent `input`                                           | New callback fires             | Fired once ✅ — the defect only appears when two subscriptions **overlap in time**                                                                                                                               |
| 5   | Non-strict script assigned `window.StoryblokBridge = Fake` after `loadStoryblokBridge()`                                 | Assignment takes effect        | No throw, but the **assignment was silently discarded** — the property is a getter with no setter, so the write vanishes                                                                                         |
| 6   | Same assignment from strict-mode code (`'use strict'`)                                                                   | Assignment takes effect        | **`TypeError: Cannot set property StoryblokBridge of #<Window> which has only a getter`**                                                                                                                        |
| 7   | Called live-preview's `loadStoryblokBridge()`, then the real `@storyblok/js` public `loadStoryblokBridge()`              | Both packages coexist          | **`TypeError: Cannot set property storyblokRegisterEvent of #<Window> which has only a getter`** — `@storyblok/js` throws and its bridge never initializes                                                       |
| 8   | Called live-preview's `loadStoryblokBridge()`, then loaded the legacy CDN bundle `storyblok-v2-latest.js` via `<script>` | Bundle installs its own global | **Uncaught `TypeError`** in the bundle; `window.StoryblokBridge` stayed the live-preview getter                                                                                                                  |
| 9   | Reverse order: `@storyblok/js` first, then live-preview                                                                  | Both coexist                   | Worked ✅ — so the breakage in 7 and 8 is **load-order dependent** and will present as intermittent                                                                                                              |
| 10  | Read `window.storyblokRegisterEvent` twice and compared                                                                  | Same function reference        | **`false`** — the getter mints a new closure per read, so identity/dedupe checks against it fail                                                                                                                 |

Rows 1–3 are the same root cause, 5–8 the same root cause. Row 4 and row 9 are the cases that pass, which is why neither defect is obvious in casual testing.

**Root cause A (rows 1–3, and the collateral in 2):** the PR gives every subscription its own bridge instance, but the bridge is not built to be instantiated twice. Its `init()` reads `config.initOnlyOnce && document.querySelectorAll('.storyblok__hint').length > 0 || isInIframe && (…attach…)` — `initOnlyOnce` defaults to `true`, and the first instance leaves a `.storyblok__hint` element in the DOM, so every later instance takes the left branch and never attaches its `message` listener. It stays silently inert. The teardown side is the same assumption inverted: the bridge's DOM (hint, overlay, stylesheet) is shared and refcounted module-wide, so the first `destroy()` removes it out from under whoever else is still subscribed.

**Root cause B (rows 5–8, 10):** `loadStoryblokBridge()` publishes the two legacy globals with `Object.defineProperty(window, name, { get, configurable: true })` — a getter and no setter. That makes them read-only, but both `@storyblok/js` and the legacy CDN bundle install those same globals by plain assignment. Whoever writes second loses: a `TypeError` from strict/ESM code, a silent no-op from sloppy code. The unstable identity in row 10 falls out of the same choice — the getter builds a fresh closure on each read instead of returning one stored function.

maoberlehner (changes requested):

maoberlehner (approved):
LGTM!

PR #783, opened 2026-08-25
feat(nuxt): add new E2E/Unit Tests Setup

JakubAndrzejewskiSB:
This PR removes flaky E2E tests in Cypress in favor of Nuxt native Vitest setup for both E2E and Unit Tests.

## Unit Tests
image

## E2E Tests
image


packages/nuxt/package.json
maoberlehner: Can we bump `@nuxt/test-utils` to `v4`? (would fix a peer dependency warning).

packages/nuxt/test/fixtures/server-client/nuxt.config.ts
```diff
@@ -0,0 +1,21 @@
...
+    },
+  },
+
+  storyblok: {
+    accessToken: "fixture-secret-token",
+    enableServerClient: true,
+  },
+
+  compatibilityDate: "2025-01-13",
+} as any);
```
maoberlehner: why do we need `as any` here?

maoberlehner (changes requested):
nitpick: oxlint.config.ts and .gitignore still have `cypress` and `playground-e2e` entries.

packages/nuxt/package.json
JakubAndrzejewskiSB: Bumped

packages/nuxt/test/fixtures/server-client/nuxt.config.ts
JakubAndrzejewskiSB: Good Catch! Fixed

JakubAndrzejewskiSB:
> nitpick: oxlint.config.ts and .gitignore still have `cypress` and `playground-e2e` entries.

Very good catch that I missed. Fixed now! :)

JakubAndrzejewskiSB:
@maoberlehner would you be also able to check the pnpm lock? There were some conflicts that I needed to resolve and I am not sure if everything was rebased correctly.

maoberlehner (approved):
Recreated the lockfile from main. LGTM!

PR #786, opened 2026-08-26
feat(react)!: redesign SDK with RSC-first architecture

dipankarmaikap:
## Overview

Redesigns `@storyblok/react` with a cleaner, framework-agnostic API. The old implicit plugin system (`storyblokInit` / `apiPlugin`) is replaced by two explicit, composable functions. Live editing is now handled by dedicated components instead of a monolithic hook.

### `createApiClient(config)`
Creates a typed API client for the Storyblok Content Delivery API. Replaces `storyblokInit` + `apiPlugin`.

```ts
import { createApiClient } from "@storyblok/react";

export const apiClient = createApiClient({
  accessToken: "YOUR_ACCESS_TOKEN",
  region: "eu", // "eu" | "us" | "ap" | "ca" | "cn"
});
```

### `defineStoryblokComponents(config)`
Registers block components and returns a pre-wired `StoryblokComponent` and `StoryblokRichText`. Replaces the `components` option in `storyblokInit`.

```ts
import { defineStoryblokComponents } from "@storyblok/react";

export const { StoryblokComponent, StoryblokRichText } = defineStoryblokComponents({
  components: { page: Page, teaser: Teaser },
  fallback: FallbackComponent,           // shown for unregistered block types
  suspenseFallback: <GlobalSkeleton />,  // shown while lazy components load
});
```

Supports `React.lazy` and Next.js `dynamic` out of the box — lazy components are auto-detected and wrapped in `Suspense`. Per-component fallbacks are also supported.

### `@storyblok/react/client` (new entry point)
Replaces the `/rsc` and `/ssr` sub-exports. All client-side live editing primitives live here.

#### `StoryblokPreview`
Client component for live editing in SPAs and Pages Router apps. Uses a render-prop so the latest story is always passed to children.

```tsx
import { StoryblokPreview } from "@storyblok/react/client";

<StoryblokPreview story={story} bridgeOptions={{ resolveRelations: ["article.author"] }}>
  {(live) => <StoryblokComponent block={live.content} />}
</StoryblokPreview>
```

#### `StoryblokPreviewRsc`
Client component for live editing in RSC / App Router apps. Calls a Server Action on every editor event to produce fresh server-rendered output. Supports Suspense streaming — the page never goes blank during updates.

```tsx
import { StoryblokPreviewRsc } from "@storyblok/react/client";

<StoryblokPreviewRsc renderContent={renderStoryServerAction}>
  <StoryContent story={story} />
</StoryblokPreviewRsc>
```

#### `useStoryblokState(story, options)`
Lower-level hook that returns the live story. Still available for custom integrations.

#### `useStoryblokEditorEvent(callback, options)`
Lowest-level hook — subscribes to raw Visual Editor events with optional debouncing.

### New types
- `StoryblokComponentProps<T>` — typed props helper for block components
- `StoryblokBlockData` — base shape for all block objects
- `StoryblokReactRichTextComponentMap` — map of richtext element overrides
- `StoryblokReactRichTextProps<T>` — props for custom richtext node/mark components
- `Story` — re-exported from `@storyblok/api-client`

### `createRichTextRenderer(options)`
Returns a render function for richtext documents as React elements. Use when you need an imperative render call rather than the `<StoryblokRichText>` component.

## Removed / Breaking Changes

| Removed | Replacement |
|---|---|
| `storyblokInit` | `createApiClient` + `defineStoryblokComponents` |
| `apiPlugin` | `createApiClient` |
| `getStoryblokApi()` | `apiClient` exported directly from your `lib/storyblok.ts` |
| `useStoryblok(slug, apiOptions, bridgeOptions)` | `apiClient.stories.get()` + `StoryblokPreview` |
| `@storyblok/react/rsc` | `@storyblok/react/client` |
| `@storyblok/react/ssr` | `@storyblok/react/client` |
| `StoryblokStory` | `StoryblokPreviewRsc` |
| `StoryblokServerComponent` | `StoryblokComponent` from your `lib/storyblok.ts` |
| `StoryblokServerStory` | `StoryblokPreview` or `StoryblokPreviewRsc` |
| `StoryblokLiveEditing` | `StoryblokPreviewRsc` |
| `setComponents` | Call `defineStoryblokComponents` per route module |
| `registerStoryblokBridge` | Handled internally by `StoryblokPreview` / `StoryblokPreviewRsc` |
| `useStoryblokRichText` | `createRichTextRenderer` |
| `convertAttributesInElement` | No longer needed — renderer outputs React elements directly |
| `SbReactRichTextComponentMap` | `StoryblokReactRichTextComponentMap` |
| `SbReactRichTextProps` | `StoryblokReactRichTextProps` |
| `blok` prop convention | `block` prop convention |
| `enableFallbackComponent` / `customFallbackComponent` in `storyblokInit` | `fallback` option in `defineStoryblokComponents` |

## Playgrounds updated

All playgrounds have been migrated to the new API:
- `playground/react` — Vite SPA with `StoryblokPreview`
- `playground/next-latest` — App Router with `StoryblokPreviewRsc` + Server Actions + Suspense streaming
- `playground/next-13-app-router` — App Router with `StoryblokPreview` (simpler pattern)
- `playground/next-13-pages-router` — Pages Router with `apiClient` in `getStaticProps` + `StoryblokPreview`
- `playground/next-latest-static` — Static export (no live editing)

## README updated

Full README rewrite to reflect the new API, covering initialization, all framework patterns (React SPA, Next.js App Router, Next.js Pages Router), richtext, code splitting, and troubleshooting.

Fixes DX-555


packages/react/cypress/e2e/index.cy.ts
```diff
@@ -1,7 +1,9 @@
 describe("@storyblok/react/rsc", () => {
```
alexjoverm: idea (non-blocking): I was thinking that, as we're revamping the React SDK from scratch, could be a good chance to replace Cypress with Playwright. It's something we have on the table for all SDKs, as it'll simplify and modernize the e2e test suite 😄 

_(doesn't need to be necessarily part of the scope of this PR, so feel free to apply it if you're motivated, otherwise we can do it afterwards)_

maoberlehner:
QA findings:

## 1 components wrapped in `memo()` or `forwardRef()`

... do not work with `defineStoryblokComponents()`

```
export const MemoTeaser = memo(TeaserImpl);

export const ForwardRefTeaser = forwardRef<HTMLDivElement, TeaserProps>(({ block }, ref) => (
  <div ref={ref} {...storyblokEditable(block)} data-test="forwardref-teaser">
    <h3>{block.headline}</h3>
  </div>
));
```

## 2. React <19 + `@storyblok/react/client`

In case anything is imported from `@storyblok/react/client`, the barrel pulls `StoryblokPreviewRsc`, which does `import { use } from "react"` → HTTP 500, `SyntaxError: The requested module 'react' does not provide an export named 'use'`. The declared peer range is `^17 || ^18 || ^19`.

```tsx
import { StoryblokPreview } from "@storyblok/react/client"; // enough to break the route
```

React 18.3.1 / Next 13 (Pages Router) — this is the README's own Pages Router example.

## 3. Old story renders forever

In case the `story` prop changes while `StoryblokPreview` / `useStoryblokState` stays mounted (SWR cache hit, `keepPreviousData`, same component across routes) and there is no `key`, the old story renders forever.

```tsx
<StoryblokPreview story={story}>{(live) => <StoryblokComponent block={live.content} />}</StoryblokPreview>
```

This is the README's "React SPA" example, which omits the `key`.

## 4. Error handling

In case the `renderContent` action rejects (thrown error, network blip, deploy skew), the rejection is read by `use()` with no boundary and Next replaces the whole document with "Application error: a server-side exception has occurred". Preview is dead until reload.

```tsx
async function renderContent(story) { throw new Error("boom"); }
<StoryblokPreviewRsc renderContent={renderContent}>{content}</StoryblokPreviewRsc>
```

React 19 / Next 16 (App Router).

## 5. `StoryblokRichText` + `blok`

In case a richtext field contains a `blok` node and it is rendered with the root `StoryblokRichText`, the block renders as nothing, with no warning.

```tsx
import { StoryblokRichText } from "@storyblok/react";   // no blok support
<StoryblokRichText document={story.content.richtext} />  // nested blocks vanish
```

Any React version. This is the README's primary richtext example.

## 6. `data` passed to `StoryblokRichText`

In case `data` is passed to the root `StoryblokRichText`, it typechecks and custom node components receive `context.data === undefined`. The `defineStoryblokComponents` variant forwards it correctly.
```tsx
import { StoryblokRichText } from "@storyblok/react";
<StoryblokRichText document={doc} data={{ locale: "de-AT" }} components={{ paragraph: P }} />
```

**What `data` is for.** `createRichTextRenderer` threads a render context to every custom node/mark component as a `context` prop. `data` is the free-form slot in it — the way you get app state (locale, the surrounding story, a link resolver) into a custom renderer without prop-drilling through the document tree.

```ts
// src/richtext/renderer.tsx:50
export interface StoryblokReactRichTextRenderContext {
  optimizeImage?: boolean | StoryblokRichTextImageOptions;
  components?: StoryblokReactRichTextComponentMap;
  data?: unknown;
}
```

**Why it typechecks.** Both richtext components take `StoryblokReactRichTextComponentProps`, which extends that context interface (`renderer.tsx:57`). So `data` is a valid prop on both — the type says nothing about whether it's used.

**The two components.** There are two things called `StoryblokRichText`:

```tsx
import { StoryblokRichText } from "@storyblok/react";                 // (a) root export
const { StoryblokRichText } = defineStoryblokComponents({ … });        // (b) registry-bound
```

(b) forwards `data`, (a) drops it on the floor:

```tsx
// (a) src/richtext/StoryblokRichText.tsx:22 — `data` is not destructured
export function StoryblokRichText({ document, optimizeImage, components }: Props) {
  const render = createRichTextRenderer({ optimizeImage, components });   // no data
  return render(document);
}

// (b) src/richtext/create-storyblok-richtext.tsx:12 — `data` is destructured and passed
return function StoryblokRichText({ document, optimizeImage, components, data }: Props) {
  const render = createRichTextRenderer({ optimizeImage, components: { blok: DefaultBlock, ...components }, data });
  return render(document);
};
```

**What a consumer hits.** A custom node component that reads `context.data`:

```tsx
function Paragraph({ children, context }: { children?: ReactNode; context?: { data?: unknown } }) {
  return <p lang={(context?.data as { locale?: string })?.locale}>{children}</p>;
}

<StoryblokRichText                      // root export
  document={doc}
  data={{ locale: "de-AT" }}
  components={{ paragraph: Paragraph }}
/>
```

## 7. Non-destructured `StoryblokRichText`

In case `StoryblokRichText` is read off the registry object during render instead of destructured once, the getter mints a new component type per access and the whole richtext subtree unmounts and remounts every render (state loss, image refetch).

```tsx
const sb = defineStoryblokComponents({ … });
function Page() { return <sb.StoryblokRichText document={doc} />; }  // remounts each render
```

## 8.`playground/next-13-pages-router` preview wrapper

In case you follow the README's Pages Router link to `playground/next-13-pages-router`, the page renders `StoryblokComponent` bare with no preview wrapper, so there is no Pages Router coverage of the bridge path at all. Adding the wrapper hits 2.

```tsx
export default function Home({ story }) { return <StoryblokComponent block={story.content} />; }
```

React 18 / Next 13.

## 9. Mutating refs while rendering

`useStoryblokEditorEvent` assigns to `callbackRef.current` / `debounceRef.current` during render rather than in an effect, which breaks React's rule against mutating refs while rendering. Observable consequence: a render that has not committed — one suspended inside `startTransition`, or one React discards — has already installed its callback, so an editor event arriving in that window runs the closure from that render instead of the committed one.

It bites only a consumer calling the hook directly with a callback that closes over state, and it will trip `react-hooks` lint and future React versions.

```tsx
// reproduced skew, not a stale result:
startTransition(() => setVersion(v => v + 1));  // render suspends, not yet committed
// editor event fires → callback sees version 1 while the committed UI still shows 0
```
React 19 (concurrent rendering).

**Proposed fix** — `packages/react/src/client/use-storyblok-editor-event.ts:35`

```diff
-import { useEffect, useRef } from "react";
+import { useEffect, useRef } from "react";
@@
   const callbackRef = useRef(callback);
-  callbackRef.current = callback;
-
   const debounceRef = useRef(debounceMs);
-  debounceRef.current = debounceMs;
+
+  // Synced after commit, never during render: a render that is discarded or
+  // suspended must not install its callback for the live subscription.
+  useEffect(() => {
+    callbackRef.current = callback;
+    debounceRef.current = debounceMs;
+  });
 
   useEffect(() => {
```

The effect deliberately has no dependency array — it must re-sync on every commit. `useEffect` rather than `useLayoutEffect` because the component is SSR'd in Next and bridge events arrive asynchronously via `postMessage`, so there is no paint-time handler to beat. Mount order is unaffected: this effect is declared before the subscription effect, so the ref holds the first callback before `onStoryblokEditorEvent` is ever called.

## 10. Story tree exists twice

In case a second or later editor event arrives, React keeps the mounted `LiveContent` (`display:none`) and renders `committedContent` again as the fallback, so the story tree exists twice — duplicate `data-blok-uid`, duplicate ids, still-mounted embeds — for the duration of every server-action round trip. Not visually duplicated; the stale copy has no layout box.
```tsx
<StoryblokPreviewRsc renderContent={renderContent}>{content}</StoryblokPreviewRsc>
// edit twice → 2× [data-blok-uid="571214631-41a3856c…"] until the action resolves
```
React 19 / Next 16 (App Router).

Here's the mechanism and the harm — and I got one detail wrong before.

**What `StoryblokPreviewRsc` renders after the first edit**

```tsx
<Suspense fallback={<>{committedContent ?? children}</>}>
  <LiveContent promise={livePromise} onCommit={setCommittedContent} />
</Suspense>
```

- **First edit.** `LiveContent` suspends on its very first render, so the boundary has never revealed its children. React shows the fallback only. One copy of the tree in the DOM.
- **Second and later edits.** `LiveContent` is already mounted and revealed. A new promise makes it suspend *again*. React does not unmount already-revealed children — it hides them (to preserve their state and DOM in case the promise resolves quickly) and renders the fallback next to them. The fallback is `committedContent`, i.e. the tree from the previous edit — the same tree that is sitting there hidden. Two copies, for the full duration of every server-action round trip.

**Why that matters — the bridge picks the wrong one**

`storyblokEditable` writes `data-blok-uid` into the markup, and the bridge resolves a block with `document.querySelector` (`utils/dom.js`: `getBlokInDocument = t => document.querySelector('[data-blok-uid="' + t + '"]')`) — first match in document order. The hidden copy comes first, and it has no layout box:

```
initial            matches: 1   bridge picks: {top: 24, height: 1177, width: 768}
edit-2 in flight   matches: 2   bridge picks: {top:  0, height:    0, width:   0}   ← hidden copy
edit-2 settled     matches: 1   bridge picks: {top: 24, height: 1177, width: 768}
```

The bridge polls `calculateElementPosition` every 300 ms and writes that rect straight onto the overlay, so click-to-edit collapses to a 0×0 box at the top-left corner for the length of each round trip. Same first-match problem for `enterComponent`, `highlightComponent` and `hoverComponent`. On top of that, side-effectful embeds (iframes, video, analytics) exist twice and client component state is duplicated.

Correction to what I said earlier: the stale copy's own computed `display` is still `block` — it's inside a container React hides, which is why it measures 0×0. My "the stale copy has no layout box" conclusion holds; the `display:none` attribution was on the wrong element.

**Proposed fix** — mark the promise swap as a transition, `packages/react/src/client/StoryblokPreviewRsc.tsx:129`

```diff
-import { type ReactNode, Suspense, use, useEffect, useState } from "react";
+import { type ReactNode, startTransition, Suspense, use, useEffect, useState } from "react";
@@
   useStoryblokEditorEvent(
     (updatedStory) => {
-      setLivePromise(renderContent(updatedStory));
+      const promise = renderContent(updatedStory);
+      startTransition(() => setLivePromise(promise));
     },
     { debounceMs, bridgeOptions },
   );
```

React does not show a fallback when a transition re-suspends already-revealed content; it keeps the current tree on screen and commits the new one when it is ready. I applied this in the worktree, rebuilt, and re-ran both probes:

```
initial            roots: 1  dup: 0  body: "Test RichtText …"
edit-1 in flight   roots: 1  dup: 0  body: "Test RichtText …"
edit-1 settled     roots: 1  dup: 0  body: "EDIT-1 …"
edit-2 in flight   roots: 1  dup: 0  body: "EDIT-1 …"
edit-2 settled     roots: 1  dup: 0  body: "EDIT-2 …"
edit-3 settled     roots: 1  dup: 0  body: "EDIT-3 …"
```

Never more than one copy, live updates still land, and the bridge lookup keeps resolving the real element. I reverted the source afterwards — only the playgrounds are left dirty.

Two notes on the fix. The comment at line 126 ("no useTransition, no awaiting here") argues against this, but its concern is the initial streaming path, which is untouched: `renderContent` is still called eagerly outside the transition, and the `if (!livePromise)` early return still hands `children` straight to the RSC channel. And once fallbacks are no longer used for re-suspension, `committedContent` and the `onCommit` plumbing become dead weight — worth removing in the same pass, though I did not test that reduction.

packages/react/src/__tests__/client/define-storyblok-components.test.tsx
```diff
@@ -0,0 +1,250 @@
+import React from "react";
```
alexjoverm: idea (non-blocking): similar as my comment above - as we're revamping the React SDK, and we're already using `vitest` for unit/integration testing... what about moving the files alongside the src code (instead of in a separated __tests__ folder)? In that way, it's more modern and aligned with other packages.

FYI - the `__test__` folder comes from the times of Jest

packages/react/src/define-storyblok-components.tsx
```diff
@@ -0,0 +1,233 @@
...
+      if (config.fallback) {
+        const FallbackComponent = config.fallback;
+        return <FallbackComponent block={block} {...rest} />;
+      }
+      console.warn(`[Storyblok] No component registered for "${block.component}".`);
+      return null;
+    }
+
+    const { component: Component, fallback, suspense } = normalizeEntry(entry);
+    const needsSuspense = suspense ?? isLazyComponent(Component);
```
maoberlehner: `next/dynamic` isn't one function. The compiler aliases the specifier per layer, and neither of the
two variants it resolves to is a lazy object, so `isLazyComponent` misses both and the entry's
`fallback` never renders, where `React.lazy` in the same slot does. Checked against the real modules
in the playground (Next 16.1.6):

| resolves to | returns | `$$typeof` | `isLazyComponent` |
| --- | --- | --- | --- |
| App Router → `api/app-dynamic` | plain function | — | `false` |
| Pages Router → `api/dynamic` | object | `Symbol(react.forward_ref)` | `false` |

```tsx
defineStoryblokComponents({
  components: { hero: { component: dynamic(() => import("./Hero")), fallback: <Skeleton /> } },
});  // <Skeleton /> never renders
```

`suspense: true` only rescues App Router. The Pages variant never suspends at all; it renders its own
`loading` (default `null`) internally, so our `<Suspense>` is inert and the block is blank until the
chunk lands. And with the README's `dynamic(() => import(...))` (default `ssr: true`, no `loading`),
Next adds no boundary of its own, so on App Router the suspension bubbles to whatever ancestor
happens to exist. README says `dynamic` gives "the same effect" as `React.lazy`; for fallbacks it
doesn't, on either router.

packages/react/src/define-storyblok-components.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+/**
+ * Normalize a component entry to extract component and config.
+ */
+function normalizeEntry(entry: StoryblokComponentEntry): {
+  component: BlockComponentType;
+  fallback?: SuspenseFallback;
+  suspense?: boolean;
+} {
+  if (typeof entry === "function" || isLazyComponent(entry) || isWrappedComponent(entry)) {
```
maoberlehner: Can we invert this with `"component" in entry`?

```ts
const normalizeEntry = (entry: StoryblokComponentEntry) =>
  typeof entry === "object" && entry !== null && "component" in entry
    ? entry
    : { component: entry as BlockComponentType };
```

packages/react/src/define-storyblok-components.tsx
```diff
@@ -0,0 +1,233 @@
...
+ * ecosystem and has been stable across React 16–19.
+ */
+function isLazyComponent(component: unknown): boolean {
+  if (typeof component !== "object" || component === null) {
+    return false;
+  }
+  const typedComponent = component as { $$typeof?: symbol };
+  return (
+    typeof typedComponent.$$typeof === "symbol" &&
+    typedComponent.$$typeof.toString() === "Symbol(react.lazy)"
```
maoberlehner: Can we drop `.toString()` and the `typeof === "symbol"` here and compare the same registered symbol directly? Instead: `$$typeof === Symbol.for("react.lazy")`

packages/react/src/define-storyblok-components.tsx
```diff
@@ -0,0 +1,233 @@
...
+ */
+function isWrappedComponent(component: unknown): boolean {
+  if (typeof component !== "object" || component === null) {
+    return false;
+  }
+  const typedComponent = component as { $$typeof?: symbol };
+  if (typeof typedComponent.$$typeof !== "symbol") {
+    return false;
+  }
+  const str = typedComponent.$$typeof.toString();
```
maoberlehner: Same as above.

packages/react/src/define-storyblok-components.tsx
```diff
@@ -0,0 +1,233 @@
...
+ * Registration type: accepts any component whose block prop is a subtype of StoryblokBlockData.
+ * Using `any` intentionally avoids contravariance errors for components with specific block shapes.
+ */
+type AnyBlockComponent = ComponentType<{ block: any }>;
+
+/**
+ * A value accepted as a Suspense fallback: either a rendered node or a component with no
+ * required props (passed as `WeatherWidgetSkeleton` or `<WeatherWidgetSkeleton />`).
+ */
+export type SuspenseFallback = ReactNode | ComponentType<object>;
```
maoberlehner: Is the `ComponentType` half of this union worth it? `<Suspense>` only takes a `ReactNode`, so
requiring `fallback={<Skeleton />}` would drop the union, `resolveSuspenseFallback`, and both call
sites (162 and 210). Related: `DefaultSuspenseFallback` is a component that returns `null`, and
`fallback={null}` is legal, so 83-86 and 162-167 collapse to
`const defaultSuspenseFallback = config.suspenseFallback ?? null`.

packages/react/src/define-storyblok-components.test.tsx
```diff
@@ -0,0 +1,305 @@
...
+    });
+
+    it("passes extra props through to the registered component", () => {
+      function WithExtra({ _block, extra }: { _block: StoryblokBlockData; extra?: string }) {
+        return <div data-testid="extra">{extra}</div>;
+      }
+      const { StoryblokComponent } = defineStoryblokComponents({
+        components: { widget: WithExtra },
+      });
+      const block = { component: "widget", _uid: "w1" } as StoryblokBlockData;
```
maoberlehner: nitpick: Can we avoid casting here with regular `const block: StoryblokBlockData =`? If not, why not? If it is about saving on boilerplate, not declaring mandatory fields if not needed for testing, I'd prefer a `makeBlockData({ component: "widget", _uid: "w1" })` helper creating a block with defaults.

maoberlehner: (same for all type casts here)

packages/react/src/types.ts
```diff
@@ -1,45 +1,20 @@
...
-
-export interface SbReactComponentsMap {
-  [key: string]: React.ElementType;
+/** Base shape of every Storyblok block. All registered block components receive this as their `block` prop. */
+export interface StoryblokBlockData {
+  _uid: string;
+  component: string;
+  _editable?: string;
+  [key: string]: unknown;
 }
```
maoberlehner: Can we use the `real` block type (generated from the OpenAPI spec) here (and for other types too)?

Our goal is to have one source of truth for common Storyblok types. SDKs should only declare what is special in the context of the SDK.

packages/react/src/define-storyblok-components.tsx
```diff
@@ -0,0 +1,233 @@
...
+   *
+   * @example
+   * ```tsx
+   * <StoryblokComponent block={story.content} />
+   * <StoryblokComponent block={story.content.body} />
+   * ```
+   */
+  StoryblokComponent: ComponentType<
+    { block: StoryblokBlockData | StoryblokBlockData[] } & Record<string, unknown>
+  >;
```
maoberlehner: `& Record<string, unknown>` makes every misspelled prop legal and turns off autocomplete on this
component: `<StoryblokComponent block={block} titlee="typo" />` compiles clean, and the typo is
forwarded to every nested block at runtime. A generic `TExtraProps = {}` would pass extra props
through without opting out of checking.

nitpick: I'm not quite sure if I like the choice of allowing `{ block: StoryblokBlockData | StoryblokBlockData[] }` as `block` very clearly suggests a singular block. So although useful, it is not an obvious API.

packages/react/src/define-storyblok-components.tsx
```diff
@@ -0,0 +1,233 @@
...
+    return <Component block={block} {...rest} />;
+  }
+
+  StoryblokComponent.displayName = "StoryblokComponent";
+
+  // Pre-compute once so every access returns the same function reference.
+  // A getter would call createStoryblokRichText() on each access, producing a
+  // new component type per render and causing React to unmount + remount the
+  // entire richtext subtree on every render.
+  const StoryblokRichText = createStoryblokRichText(StoryblokComponent as BlockComponentType);
```
maoberlehner: I think `as BlockComponentType` is not necessary here.

packages/react/src/define-storyblok-components.tsx
```diff
@@ -0,0 +1,233 @@
...
+    if (!entry) {
+      if (config.fallback) {
+        const FallbackComponent = config.fallback;
+        return <FallbackComponent block={block} {...rest} />;
+      }
+      console.warn(`[Storyblok] No component registered for "${block.component}".`);
+      return null;
+    }
+
+    const { component: Component, fallback, suspense } = normalizeEntry(entry);
```
maoberlehner: All of this is derivable at factory time from a map that never changes, but it runs per block per
render: the lookup, `normalizeEntry`, the symbol stringify (a fresh allocation per block per render),
and a fresh `<F />` for the fallback. Building a
`Map<string, { Component, needsSuspense, fallbackNode }>` once above `StoryblokComponent` makes this
a `Map.get` plus a branch. Roughly 30 lines, no API change, and it matters most on the live preview
path where everything re-renders per keystroke.

packages/react/src/types.ts
```diff
@@ -1,45 +1,20 @@
...
-  enableFallbackComponent?: boolean;
-  customFallbackComponent?: React.ElementType;
+/**
+ * Helper type for typed block component props.
+ *
+ * @example
+ * type PageProps = StoryblokComponentProps<{ body: StoryblokBlockData[] }>;
+ * export default function Page({ block }: PageProps) { ... }
+ */
+export interface StoryblokComponentProps<
```
maoberlehner, line 28: I find this type confusing. The name reads as "props of `<StoryblokComponent>`". So I'd expect this is the type of the props of `<StoryblokComponent>` but that's not what this is.

---

It's generally pretty confusing (not your code but general nomenclature around all of this):

`Component` can mean: 1) a React `Component`, 2) a Storyblok `Component` (which we want to name `Block`)

So if we don't want to use `Component` for Storyblok *components*, then in the case of `StoryblokComponent` it must mean this is the Storyblok *React* component. But suffixing a React component with `Component` is not a common pattern.

Long story short: I think `StoryblokComponent` is the wrong name. `StoryblokBlock` would be the name if we follow our new naming pattern (which would make sense for a complete rewrite of the SDK).

But personally I'd go with `<StoryblokBlocks blocks=[]>` so it is clear multiple are supported, and for rendering one, one simply adds one in the array.

```tsx
<StoryblokBlocks blocks=[]>
// Maybe two for convenience
<StoryblokBlock block={}>
```

---

What this would leave us for `StoryblokComponentProps` is that I think if we use `<StoryblokBlock(s)>` a possible name for props for React components that render a Storyblok Block is: `StoryblokBlockComponentProps`.

packages/react/src/richtext/fixtures/HeadingWithRichText.tsx
```diff
@@ -3,20 +3,22 @@
  * This tests the infinite loop prevention - without it, this would cause:
  * HeadingWithRichText -> StoryblokRichText -> HeadingWithRichText -> ...
  */
-import type { SbReactRichTextProps } from "@storyblok/react";
-import { StoryblokRichText } from "@storyblok/react";
+import type { StoryblokReactRichTextProps } from "@storyblok/react";
```
maoberlehner: Do we use the package name instead of a path here on purpose (look at all such instances)?

packages/react/src/index.ts
```diff
@@ -1,58 +1,50 @@
...
+  groupLinkNodes,
+  hasContent,
+  isSelfClosing,
+  normalizeNodes,
+  processAttrs,
+  renderRichText,
+  resolveTag,
+  splitTableRows,
+  styleToString,
+} from "@storyblok/richtext";
```
maoberlehner: Not quite sure if we really want to do this. This tightly couples the React SDK's API surface to that of the richtext package.

packages/react/src/index.ts
```diff
@@ -1,58 +1,50 @@
...
+
+export { type ContentApiClientConfig, createApiClient, type Story } from "@storyblok/api-client";
+export { storyblokEditable } from "@storyblok/live-preview";
```
maoberlehner: It's more something like a feeling than a general statement, but re-exporting from other packages does not feel quite right to me. Especially with the `api-client` we force a dependency onto our users that they might not want to use at all.

There are a couple of ways around this. E.g., we could mark `api-client`, `live-preview` and `richtext` as `peerDependencies`.

But what I'd prefer is to either keep them out or provide useful abstractions around them that fit the SDK paradigm. (cc @alexjoverm what do you think on that matter?)

For richtext, we already do that, but for live-preview and the capi we just re-export as far as I can see.

For `storyblokEditable` we could do something like:

```tsx
// 🚫
const Feature = ({ block }: FeatureProps) => (
  <div data-test="feature" {...storyblokEditable(block)}>
    <div>{block.name}</div>
    <p>{block.description}</p>
  </div>
);

// ✅
const Feature = ({ block, editable }: FeatureProps) => (
  <div data-test="feature" {...editable}>
    <div>{block.name}</div>
    <p>{block.description}</p>
  </div>
);
```

maoberlehner (changes requested):

packages/react/src/client/storyblok-preview.tsx
```diff
@@ -0,0 +1,41 @@
+"use client";
```
maoberlehner, line 1: different casing than all the other files: I'd highly favor kebab-case always for file names as macos is not case sensitive when it comes to file names: `storyblok-preview.tsx`.

(also true at least for `storyblok-preview-rsc.tsx` too.

packages/react/src/client/use-storyblok-editor-event.ts
```diff
@@ -0,0 +1,84 @@
...
+        if (!mounted) return;
+
+        const ms = debounceRef.current;
+
+        if (ms !== undefined) {
+          if (debounceTimer !== null) clearTimeout(debounceTimer);
+          debounceTimer = setTimeout(() => {
+            debounceTimer = null;
+            if (!mounted) return;
+            callbackRef.current(updatedStory as Story);
```
maoberlehner: I think this can be simplified, but please double check:

```diff
--- a/src/client/use-storyblok-editor-event.ts
+++ b/src/client/use-storyblok-editor-event.ts
@@ -46,7 +46,7 @@
   useEffect(() => {
     let mounted = true;
     let unsubscribe: (() => void) | undefined;
-    let debounceTimer: ReturnType<typeof setTimeout> | null = null;
+    let debounceTimer: ReturnType<typeof setTimeout> | undefined;
 
     const setup = async () => {
       const fn = await onStoryblokEditorEvent((updatedStory) => {
@@ -55,12 +55,10 @@
         const ms = debounceRef.current;
 
         if (ms !== undefined) {
-          if (debounceTimer !== null) clearTimeout(debounceTimer);
-          debounceTimer = setTimeout(() => {
-            debounceTimer = null;
-            if (!mounted) return;
-            callbackRef.current(updatedStory as Story);
-          }, ms);
+          clearTimeout(debounceTimer);
+          debounceTimer = setTimeout(() => callbackRef.current(updatedStory as Story), ms);
         } else {
           callbackRef.current(updatedStory as Story);
         }
@@ -77,7 +75,7 @@
 
     return () => {
       mounted = false;
-      if (debounceTimer !== null) clearTimeout(debounceTimer);
+      clearTimeout(debounceTimer);
       unsubscribe?.();
     };
   }, []);
```

packages/react/src/client/use-storyblok-editor-event.ts
```diff
@@ -0,0 +1,84 @@
...
+    debounceRef.current = debounceMs;
+  });
+
+  useEffect(() => {
+    let mounted = true;
+    let unsubscribe: (() => void) | undefined;
+    let debounceTimer: ReturnType<typeof setTimeout> | null = null;
+
+    const setup = async () => {
+      const fn = await onStoryblokEditorEvent((updatedStory) => {
```
maoberlehner, line 51: Every call to this hook constructs its own bridge, its own `window.message` listener and its own
overlay DOM. `onStoryblokEditorEvent` force-sets `initOnlyOnce: false` precisely because a second
subscriber would otherwise go deaf, which is a workaround for a one-subscriber design. A preview
banner plus a `useStoryblokState` page means each editor event is deserialized and dispatched twice.
A refcounted broker in `@storyblok/live-preview` (one bridge per options key, fan out to N callbacks,
destroy at zero refs) fixes it for every SDK, and gives `reconcileStory` a single place to run.


```diff
--- a/packages/live-preview/src/onStoryblokEditorEvent.ts
+++ b/packages/live-preview/src/onStoryblokEditorEvent.ts
@@
+type Subscriber = (story: LivePreviewStory) => void;
+
+type BridgeEntry = {
+  bridge: Promise<StoryblokBridge>;
+  subscribers: Set<Subscriber>;
+};
+
+const entries = new Map<string, BridgeEntry>();
+
+/**
+ * Options that cannot be compared structurally (callbacks) get their own
+ * bridge, since two such subscriptions can't be proven equivalent.
+ */
+function optionsKey(options: BridgeParams | undefined): string | undefined {
+  try {
+    return JSON.stringify(options ?? {}, (_, value) =>
+      typeof value === "function" ? UNSHAREABLE : value,
+    );
+  } catch {
+    return undefined;
+  }
+}
+
+function acquire(key: string, bridgeOptions: BridgeParams | undefined): BridgeEntry {
+  const existing = entries.get(key);
+  if (existing) return existing;
+
+  const subscribers = new Set<Subscriber>();
+  // Cache the promise, not the resolved instance: two subscribers mounting in
+  // the same tick would otherwise both miss the cache and construct a bridge.
+  const bridge = loadStoryblokBridge({ ...bridgeOptions, initOnlyOnce: false }).then((instance) => {
+    instance.on(["input", "change", "published"], (event) => {
+      if (!event) return;
+
+      if (event.action === "input" && event.story) {
+        const story = event.story as LivePreviewStory;
+        for (const subscriber of [...subscribers]) subscriber(story);
+        return;
+      }
+
+      if (event.action === "change" || event.action === "published") {
+        window.location.reload();
+      }
+    });
+    return instance;
+  });
+
+  const entry = { bridge, subscribers };
+  entries.set(key, entry);
+  return entry;
+}
+
 export async function onStoryblokEditorEvent<TStory extends Story = Story>(
   callback: (story: LivePreviewStory<TStory>) => void,
   bridgeOptions?: BridgeParams,
 ): Promise<() => void> {
   if (!isBrowser() || !isInEditor(new URL(window.location.href))) {
     return () => {};
   }
 
-  let active = true;
-  // Force initOnlyOnce: false so that every call gets its own fully-initialised
-  // bridge instance with its own window `message` listener. The bridge defaults
-  // to initOnlyOnce: true, which silently skips addMessageListener() when a
-  // .storyblok__hint element is already in the DOM (left by a prior instance),
-  // making any second concurrent subscription permanently deaf to editor events.
-  const bridge = await loadStoryblokBridge({ ...bridgeOptions, initOnlyOnce: false });
-
-  bridge.on(["input", "change", "published"], (event) => {
-    if (!active || !event) {
-      return;
-    }
-
-    if (event.action === "input" && event.story) {
-      callback(event.story as LivePreviewStory<TStory>);
-      return;
-    }
-
-    if (event.action === "change" || event.action === "published") {
-      window.location.reload();
-    }
-  });
+  const key = optionsKey(bridgeOptions) ?? `unshareable:${++unshareableCount}`;
+  const entry = acquire(key, bridgeOptions);
+  const subscriber = callback as Subscriber;
+  entry.subscribers.add(subscriber);
+  await entry.bridge;
 
   return () => {
-    if (!active) return;
-    active = false;
-    bridge.destroy();
+    if (!entry.subscribers.delete(subscriber)) return;
+    if (entry.subscribers.size > 0) return;
+
+    entries.delete(key);
+    entry.bridge.then((instance) => {
+      // A resubscribe during the await would have created a fresh entry, so
+      // this instance is unreachable either way and safe to destroy.
+      instance.destroy();
+    });
   };
 }
```

packages/react/src/client/use-storyblok-editor-event.ts
maoberlehner: > callbackRef.current(updatedStory as Story);

As far as I remember, this is a lie, as the bridge does not always respond with a (full) story. So I think `as Story` is not something we can / should do here.

packages/react/src/client/StoryblokPreview.tsx
```diff
@@ -0,0 +1,41 @@
...
+ *   {(live) => <StoryblokComponent block={live.content} />}
+ * </StoryblokPreview>
+ * ```
+ */
+export function StoryblokPreview({ story, children, ...options }: StoryblokPreviewProps) {
+  const current = useStoryblokState(story, options);
+  return <>{children(current)}</>;
+}
+
+export default StoryblokPreview;
```
maoberlehner: Do we need this default export here in addition to the named one?

packages/react/package.json
```diff
@@ -32,24 +33,14 @@
         "default": "./dist/index.cjs"
       }
     },
-    "./ssr": {
+    "./client": {
```
maoberlehner, line 36: I looked at other libraries and started to wonder: do we actually need `/client` entry point? What does it buy us? Especially, `/client` is a little odd considering that React is primarily a client-side library. I guess it is about `use client` but that's not obvious, and also not quite clear to me why it earns a separate entry point?

Other libraries tend to have a special package/entry for `next` in particular. E.g., `@sanity/next-loader` or `@prismicio/next`. I find this more intuitive than having a separate `(use)client` entry point.

packages/react/src/client/storyblok-preview-rsc.tsx
```diff
@@ -0,0 +1,191 @@
...
+ *   const story = await fetchStory(params.slug);
+ *   return (
+ *     <StoryblokPreviewRsc renderContent={renderStory} >
+ *       <MyStoryContent story={story} />
+ *     </StoryblokPreviewRsc>
+ *   );
+ * }
+ * ```
+ */
+export function StoryblokPreviewRsc({
```
maoberlehner: Not a big fan of the `Rsc` suffix. Bad reason why: it looks weird. Better reason why: the Storyblok React **Server** Component is a `use client` component? From the naming, I'd expect it to be a `use server` component. Personally, I get that both `use server` and `use client` are part of the Next RSC implementation, but it's still not very intuitive naming, IMO.

Another thing that I think is not great is that there are two rendering concepts in one component, and users are expected to make sure that both render the same thing.

```tsx
<StoryblokPreviewRsc renderContent={renderStory} >
  <MyStoryContent story={story} />
 </StoryblokPreviewRsc>
```

`<MyStoryContent story={story} />` must render the same as `renderStory`. That's asking for trouble, and users will wonder why they need to specify the same thing twice.

Variants I'd prefer:

```tsx
// a) just change the entries but keep the same name for all:
import { StoryblokPreview } from "@storyblok/react" // Plain React variant
import { StoryblokPreview } from "@storyblok/react/next" //Next-specific component or...
// if TanStack and React router also need this version:
import { StoryblokPreview } from "@storyblok/react/server"
```

Or one component to rule them all:

```tsx
type Common = { story: Story; debounceMs?: number; bridgeOptions?: BridgeParams };
type ClientProps = Common & { children: (story: Story) => ReactNode; renderContent?: never };
type ServerProps = Common & { renderContent: (story: Story) => Promise<ReactNode>; children?: never };
export type StoryblokPreviewProps = ClientProps | ServerProps;

/**
 * A truthiness or `in` check on `renderContent` cannot narrow the union: TS only
 * discriminates on unit-typed properties, and a function type is not one.
 */
function isServerMode(props: StoryblokPreviewProps): props is ServerProps {
  return typeof props.renderContent === "function";
}

/**
 * Kept out of StoryblokPreview so the exported component stays sync. An async
 * component reached from a client bundle throws "Only Server Components can be
 * async"; this one is only ever rendered in server mode.
 */
async function ServerRenderedPreview({ renderContent, story, ...rest }: ServerProps) {
  // Direct server call, not a round trip. Awaiting only resolves the outer
  // function, so async components in the returned tree still stream.
  return (
    <StoryblokPreviewClient renderContent={renderContent} story={story} {...rest}>
      {await renderContent(story)}
    </StoryblokPreviewClient>
  );
}

export function StoryblokPreview(props: StoryblokPreviewProps): ReactNode {
  if (isServerMode(props)) return <ServerRenderedPreview {...props} />;

  const { children, story, debounceMs, bridgeOptions } = props;
  return (
    <StoryblokPreviewClient story={story} debounceMs={debounceMs} bridgeOptions={bridgeOptions}>
      {children}
    </StoryblokPreviewClient>
  );
}
```

The `"use client"` half subscribes once and branches in the callback, so a page using both modes
opens one bridge instead of two:

```tsx
export function StoryblokPreviewClient(props: StoryblokPreviewClientProps): ReactNode {
  const { story, bridgeOptions } = props;
  const renderContent = isServerMode(props) ? props.renderContent : undefined;
  // Server mode debounces by default so a keystroke does not trigger a Server
  // Action per character. Client mode re-renders locally, so it does not.
  const debounceMs = props.debounceMs ?? (renderContent ? DEFAULT_SERVER_DEBOUNCE_MS : undefined);
  const [liveStory, setLiveStory] = useState<Story>(story);
  const [livePromise, setLivePromise] = useState<Promise<ReactNode> | null>(null);

  // Sync the prop into state when the story identity changes, so cross-route
  // navigation on a reused component instance does not render stale content.
  useEffect(() => {
    setLiveStory(story);
  }, [story.id]);

  useStoryblokEditorEvent(
    (updatedStory) => {
      if (!renderContent) {
        setLiveStory(updatedStory);
        return;
      }
      startTransition(() => setLivePromise(renderContent(updatedStory)));
    },
    { debounceMs, bridgeOptions },
  );

  if (!isServerMode(props)) return <>{props.children(liveStory)}</>;
  if (!livePromise) return <>{props.children}</>;

  return (
    <LiveContentBoundary promise={livePromise} fallback={<>{props.children}</>}>
      <Suspense fallback={<>{props.children}</>}>
        <LiveContent promise={livePromise} />
      </Suspense>
    </LiveContentBoundary>
  );
}
```

Both call sites then name their content exactly once:

```tsx
<StoryblokPreview story={story}>{(live) => <StoryContent story={live} />}</StoryblokPreview>
<StoryblokPreview story={story} renderContent={renderContent} />
```

and the playground loses its duplicate:

```diff
-  const content = <StoryContent story={story} />;
-
   return (
     <>
       <PreviewBanner />
-      <StoryblokPreviewRsc renderContent={renderContent}>{content}</StoryblokPreviewRsc>
+      <StoryblokPreview story={story} renderContent={renderContent} />
     </>
   );
```

packages/react/src/client/storyblok-preview-rsc.tsx
```diff
@@ -0,0 +1,191 @@
+"use client";
+import type { BridgeParams } from "@storyblok/live-preview";
+import type { Story } from "@storyblok/api-client";
+// `React` namespace imported separately so `React.use` is a runtime property
+// lookup rather than a static named import. This means the module loads on
+// React <19 without a SyntaxError, even though StoryblokPreviewRsc itself
+// requires React 19 to function.
```
maoberlehner: The comment claims the namespace import makes `React.use` "a runtime property lookup rather than a
static named import", so the module loads on React <19. Webpack disagrees, and
`playground/next-13-pages-router` is red on the branch right now:

```
Attempted import error: 'use' is not exported from 'react' (imported as 'React$1').
Import trace: dist/client/StoryblokPreviewRsc.mjs → dist/client.mjs
> Build failed because of webpack errors
```

Reproduced against unmodified branch `dist`. Turbopack and the App Router builds do not care, which
is why nothing else caught it. One line fixes it and the build goes green:

```ts
const reactUse = Reflect.get(React, "use") as (<T>(p: Promise<T>) => T) | undefined;
```

`React$1["use"]` with a string literal is not enough, webpack folds that too. This matters more if the
two preview components merge: today the failing module is one a pages-router app has no reason to
import, but merged it becomes the only client component. I confirmed that by deleting
`StoryblokPreviewRsc` and rebuilding, which moves the same error onto the merged file.

packages/react/src/richtext/storyblok-rich-text.tsx
```diff
@@ -0,0 +1,37 @@
...
+  data,
+}: StoryblokReactRichTextComponentProps): ReactNode {
+  // Memoize the renderer so it is not recreated on every render when options
+  // are stable references (e.g. module-level component maps).
+  const render = useMemo(
+    () => createRichTextRenderer({ optimizeImage, components, data }),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [optimizeImage, components, data],
+  );
+  return render(document);
```
maoberlehner: The new `useMemo` caches the renderer, but the expensive half is `render(document)` on the next line,
which still runs every time. It calls `normalizeNodes(doc, true)`, whose `addKeys` recursively
shallow-copies every node and every mark to attach `_key`. I measured it on this commit with a stable
document and a hoisted `components` object, the best case for the memo: three parent renders still do
3× the node work, 60 renders for a 20-node doc. `document` needs to be in the deps, or the memo
should wrap the rendered output rather than the renderer.

Two things alongside that. The deps `[optimizeImage, components, data]` are all inline literals at
the README's call site, so in the documented usage the memo never hits; worth keying on entries or
documenting that they must be hoisted. And `create-storyblok-richtext.tsx` didn't get this fix, which
is the copy `defineStoryblokComponents` returns and therefore the one almost everyone renders. Same
60-for-20 there.

packages/react/src/richtext/storyblok-rich-text.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * @example
+ * ```tsx
+ * <StoryblokRichText
+ *   document={story.content.richtext}
+ *   optimizeImage={{ width: 800 }}
+ *   components={{ heading: MyHeading }}
+ * />
+ * ```
+ */
+export function StoryblokRichText({
```
maoberlehner: This is `create-storyblok-richtext.tsx` minus the `blok` injection. Making that factory's parameter
optional and doing `export const StoryblokRichText = createStoryblokRichText()` in `richtext/index.ts`
deletes this file and guarantees the two can't drift again, which they already have twice: first the
`data` prop, and now `151c5002c` adding the `useMemo` here and not there, so the memoized copy is the one hardly anyone uses. `richtext.test.tsx:20` never touches this export, it goes through
`defineStoryblokComponents`, so they're already interchangeable in practice. `create-default-block.tsx`
is 19 lines with one call site and could be inlined in the same pass: four files to two.

packages/react/src/richtext/renderer.tsx
```diff
@@ -22,9 +22,8 @@ import {
 import React, { type ComponentType, type ReactNode } from "react";
```
maoberlehner: Only file using the `import React, { … }` form despite `jsx: "react-jsx"`. Named imports
(`createElement`, `Fragment`, `type Key`) match the rest of the package and tree-shake better.

packages/react/package.json
```diff
@@ -18,6 +18,7 @@
     "dist"
   ],
   "type": "module",
+  "sideEffects": false,
```
maoberlehner, line 21: Nice addition! Yet the flag landed here, but it doesn't do anything on its own. I re-measured against this commit: an
entry importing nothing but `storyblokEditable` still bundles to 37,270 bytes minified (37,314 before
the flag), still including the api-client's request layer, `getValidRequestBody`, `isHTTPError`,
`isTimeoutError`, `timeout`, `delay`. The retained code is in the dependencies, so bundlers still
can't prove *those* modules are inert. With `sideEffects: false` also on `live-preview` and
`capi-client` the same bundle is 401 bytes, 98.9% smaller (`richtext` already has it). Worth landing
all three in this PR; `unbundle: true` already gives us the per-module output that makes it pay off.

packages/react/package.json
```diff
@@ -108,27 +94,9 @@
     "vitest": "catalog:"
   },
   "peerDependencies": {
-    "next": "^13 || ^14 || ^15 || ^16",
     "react": "^17 || ^18 || ^19",
```
maoberlehner: The `^17` here doesn't hold up, and it's worth settling as a policy rather than patching one import.
The package reaches past 17 in four places, two of which fail before anything renders:

| Used | Floor | Where |
| --- | --- | --- |
| `react/jsx-runtime` (`jsx`, `Fragment`) | 17, but 18 to resolve under Node ESM | every module emitting JSX |
| `Suspense`, `useState`, `useEffect`, `useRef` | 16.6 / 16.8 | throughout |
| `startTransition` (static named import) | **18** | `StoryblokPreviewRsc.tsx:9` |
| `React.use` (namespace read, now guarded) | **19** | `StoryblokPreviewRsc.tsx:56`, guard at `:148` |

React 17 fails to load under Node ESM twice over: it has no `exports` map, so `react/jsx-runtime`
doesn't resolve for any JSX module we ship, and `startTransition` isn't one of its exports, so
`client.mjs` is rejected at link time with `Named export 'startTransition' not found`. Bundlers hide
both (extensionless resolution works, interop turns the missing export into `undefined`, and it's
never called on the SPA path), which is why the playgrounds are green. The new `React.use` guard is
the right move (option 3 below), but on 17 it never runs, since the module can't load to reach it.

Four options, roughly in order of value:

1. **Drop 17.** It's inherited from v7, not chosen for this rewrite. The jsx-runtime resolution
   failure can't be fixed without a classic-runtime build, the RSC path needs 19, and Next 13+
   already requires 18+, so the range promises a version where half the package can't run and nothing
   is tested. `^18 || ^19` is the honest line.
2. **Whatever the floor is, never import a version-gated export by name.**
   `import { x } from "react"` is a link-time dependency; `React.x` is a runtime read. At module
   scope, `const startTransition = React.startTransition ?? ((fn) => fn())` is a correct degradation,
   since a renderer without transitions is eager anyway. `use` already does this three lines above.
   Same treatment for anything post-floor: `useId`, `useSyncExternalStore`, `useTransition`,
   `useDeferredValue` (18), `use`, `useOptimistic`, `useActionState` (19).
3. **For APIs with no fallback, gate loudly.** Already done for `use` in this commit. npm peers are
   per-package, not per-subpath, so a runtime guard plus a README line is the ceiling for per-export
   ranges; worth adding the README half too.
4. **Test the floor.** Install each supported major and, for each, import both entries under Node ESM
   *and* through a bundler, then render the SPA path. The two resolution paths disagree, the bundler
   hiding what ESM rejects, so testing only the bundler path (which is all the playgrounds do) hides
   exactly this class of bug.

packages/react/tsconfig.json
```diff
@@ -10,7 +10,7 @@
       "@/*": ["./src/*"]
     },
     "resolveJsonModule": true,
-    "types": ["cypress", "node"],
+    "types": ["node"],
     "allowJs": true,
     "strict": false,
```
maoberlehner: This is the only package in the repo with `strict: false`; all 17 others are strict. I flipped it and
ran `tsc --noEmit`: zero errors. One-line change, no migration, and it's the difference between the
next contributor's `any` being caught and not.

packages/react/src/client/storyblok-preview-server-runtime.tsx
```diff
@@ -0,0 +1,191 @@
...
+  // Note: this guard cannot be covered by unit tests because the React module
+  // namespace is sealed in the test environment. It is verified manually.
+  if (typeof React.use !== "function") {
+    throw new Error(
+      "[Storyblok] StoryblokPreviewRsc requires React 19 (React.use is not available). " +
+        "Use StoryblokPreview for React 17/18.",
+    );
+  }
+
+  const [livePromise, setLivePromise] = useState<Promise<ReactNode> | null>(null);
```
maoberlehner, line 174: A superseded action runs to completion on the server: I fired two editor events 5 ms apart and both
settled, the first one's output discarded. Every async Server Component in the subtree re-runs its
fetch per batch, which is why the playground's `WeatherWidget` needs `unstable_cache` plus `cache`
purely to hide a 10 s fetch from this loop.

Aborting isn't available: an `AbortSignal` can't be serialized to a Server Action, and even if the
client drops the connection Node keeps running the handler. What is available is refusing to start
the next one until the current settles, which bounds concurrent server work to one action instead of
one per debounce window:

```diff
   const [livePromise, setLivePromise] = useState<Promise<ReactNode> | null>(null);
+  const inFlight = useRef(false);
+  const queued = useRef<Story | null>(null);
+
+  // One action at a time. A story arriving mid-flight replaces whatever was
+  // queued, so the server renders the first and the latest, never the ones in
+  // between.
+  function run(story: Story) {
+    inFlight.current = true;
+    const promise = renderContent(story).finally(() => {
+      inFlight.current = false;
+      const next = queued.current;
+      queued.current = null;
+      if (next) run(next);
+    });
+    startTransition(() => setLivePromise(promise));
+  }

   useStoryblokEditorEvent(
     (updatedStory) => {
-      const promise = renderContent(updatedStory);
-      startTransition(() => setLivePromise(promise));
+      if (inFlight.current) {
+        queued.current = updatedStory;
+        return;
+      }
+      run(updatedStory);
     },
     { debounceMs, bridgeOptions },
   );
```

`.finally` passes the value through, so `React.use` and the error boundary see the same resolution and
the same rejection they do today, and the promise identity the boundary compares still changes per
event.

The trade is latency on the last keystroke of a burst: it waits out the in-flight action before its
own starts, so worst case the final render is one action-duration later than today. That's the right
side to err on when an action can re-run a 10 s fetch, but it's a behavior choice, not a fix, so it's
worth deciding rather than defaulting into.

(I also checked whether narrowing the payload to `story.content` would help. On the playground's real
story the metadata is 0.5 KB of 3.37 KB, and that share only shrinks as content grows, so that one's
cosmetic.)

maoberlehner (changes requested):
My review became quite extensive. Some of it is very opinionated. Feel free to disagree!

Should we add React Router and TanStack Start playgrounds? Although they don't have the numbers of Next.js, they're not too far away from Nuxt numbers, and for Nuxt we have a complete SDK.

Even more importantly, they would guide our architecture in the sense of if we need to make something Next-specific or merely RSC-specific.

packages/react/src/index.ts
dipankarmaikap: Right, Removing.

packages/react/cypress/e2e/index.cy.ts
dipankarmaikap: Done

packages/react/src/__tests__/client/define-storyblok-components.test.tsx
dipankarmaikap: Done

packages/react/src/define-storyblok-components.tsx
dipankarmaikap: Nice catch, `React.lazy` the way

packages/react/package.json
```diff
@@ -32,14 +33,14 @@
         "default": "./dist/index.cjs"
       }
     },
-    "./ssr": {
+    "./client": {
```
maoberlehner: @dipankarmaikap after you told me it is needed because of "use client" I double-checked, and I don't think it is the case that it is needed for that. It only seems to be true for `unbundle: false`, but we want `unbundle: true` anyway.

packages/react/src/rsc/storyblok-preview.tsx
```diff
@@ -0,0 +1,76 @@
...
+export async function StoryblokPreview({
+  story,
+  renderContent,
+  debounceMs,
+  bridgeOptions,
+}: StoryblokPreviewProps): Promise<ReactNode> {
+  // The bridge only ever sends the LivePreviewStory shape (id, and
+  // optionally uuid/content) to renderContent on updates. The initially
+  // fetched Story is always a superset of that, so this first call is safe.
+  const content = await renderContent(story as unknown as LivePreviewStory);
```
maoberlehner: I think this cast is not necessary as `Story` should satisfy `LivePreviewStory`.

packages/react/playground/next-latest/src/app/components/StoryContent.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+type StoryContentProps = { story: Story };
+
+export function StoryContent({ story }: StoryContentProps) {
+  return (
+    <main className="container mx-auto">
+      <Nav />
+      {/* story.content resolves to BlockContent<RootBlock> (distributed form);
+          cast to the base BlockContent shape that StoryblokComponent expects. */}
+      <StoryblokComponent block={story.content as BlockContent} />
```
maoberlehner: If I understand this correctly, the need for this is fixed in #806. Reminder to remove this when this PR is rebased onto the latest main.

packages/react/playground/next-latest/src/lib/actions.tsx
```diff
@@ -0,0 +1,11 @@
+"use server";
+
+import type { Story, LivePreviewStory } from "@storyblok/react";
+import type { ReactNode } from "react";
+import { StoryContent } from "@/app/components/StoryContent";
+
+export async function renderContent(story: LivePreviewStory): Promise<ReactNode> {
+  // LivePreviewStory is the minimal bridge payload type; the bridge sends the
+  // full story at runtime so the cast is safe.
+  return <StoryContent story={story as Story} />;
```
maoberlehner: I think we're not allowed to do this cast, as `LivePreviewStory`  isn't guaranteed to be a full story. The comment above is wrong!

1. For regular editing, the live preview story is lacking `default_full_slug`.
2. In version history preview, we only get `{ content, id }`, missing at least  `uuid`, `slug`, `full_slug`, `name`, `alternates`.

I fear `StoryContent` must be able to deal with a `LivePreviewStory`.

packages/react/src/richtext/index.ts
```diff
@@ -0,0 +1,16 @@
...
+  createRichTextRenderer,
+} from "./renderer";
+
+import { createStoryblokRichText } from "./create-storyblok-richtext";
+export { createStoryblokRichText };
+
+// Standalone richtext component with no embedded block support.
+// Consumers who need blok rendering should use the StoryblokRichText
+// returned by defineStoryblokComponents instead.
+export const StoryblokRichText = createStoryblokRichText();
```
maoberlehner, line 16: I think it is confusing to have two `StoryblokRichText`. When should this one be preferred over the one returned by `defineStoryblokComponents`? IMO, I'd prefer that we remove this one.

packages/react/playground/next-latest/src/app/components/storyblok/Page.tsx
```diff
@@ -0,0 +1,20 @@
+import type { BlockContent, StoryblokComponentProps } from "@storyblok/react";
+import { storyblokEditable, StoryblokRichText } from "@storyblok/react";
```
maoberlehner: Shouldn't we use the registry-bound `StoryblokRichText` from `lib/storyblok` here?

packages/react/README.md
```diff
@@ -52,17 +52,19 @@ using Storyblok and Next.js from start to finish.
```
maoberlehner: According to our guidelines, we should keep the docs here as minimal as possible, can you create a docs PR instead and use this opportunity to trim this README to a minimum?

packages/react/README.md
```diff
@@ -190,536 +191,404 @@ Below is the guide and examples on how to use it with different frameworks -
...
-  {
-    resolveRelations: ["Article.author"],
-    resolveLinks: "url",
-    preventClicks: true,
-  },
-);
+```tsx
+<StoryblokPreview
+  story={story}
+  bridgeOptions={{ resolveRelations: ["article.author"] }}
```
maoberlehner: I'm not very happy about the ergonomics of this. We have three places to state one intent here: `inlineRelations: true` on the client, `resolve_relations: "article.author"` on the query, and `resolveRelations: ["article.author"]` here.

Additionally, this `resolveRelations` accepts an array, while CAPI `resolve_relations` takes a comma-separated string.

Not sure about the solution, but at the very least, this needs to be well documented in the docs (also note that the new CAPI client needs explicit `inlineRelations: true`).

Maybe we can document a pattern that reduces duplication and puts as much as possible into one place (e.g., defining `resolve_relations` only once).

packages/react/playground/next-latest/src/app/components/storyblok/Feature.tsx
```diff
@@ -0,0 +1,25 @@
+import type { StoryblokComponentProps } from "@storyblok/react";
+import { storyblokEditable } from "@storyblok/react";
+
+type FeatureProps = StoryblokComponentProps<{
+  name: string;
+  description: string;
+  color: {
```
maoberlehner: `block.color` isn't guaranteed to exist. The field isn't required in the component schema, so a `Feature` where nobody touched the color picker has no color key, and this throws. No error.tsx in this app, so it takes out the whole route rather than the one block.

packages/react/playground/next-13-app-router/app/page.tsx
```diff
@@ -1,23 +1,11 @@
-import { type ISbStoriesParams, type StoryblokClient, StoryblokStory } from "@storyblok/react/rsc";
-import { getStoryblokApi } from "@/lib/storyblok";
+import { apiClient } from "@/lib/storyblok";
+import { StoryContent } from "@/components/StoryContent";
 
 export default async function Home() {
-  const { data } = await fetchData();
+  const result = await apiClient.stories.get("react", { query: { version: "draft" } });
```
maoberlehner: This playground can't be opened in the Visual Editor as it stands. It fetches the story `react` but only serves `/`, and the editor builds the preview URL from the story's real path, so it loads `/react` and gets a 404.

`next-latest` and the Vite one already handle this with a catch-all. Could we do the same here ( `app/[[...slug]]/page.tsx` deriving the slug from params, `react` as the fallback for `/`)?

Same thing in `next-latest-static`.

packages/react/README.md
maoberlehner: There's no UMD or IIFE build — `vite.config.ts` only emits esm and cjs, and there's no `unpkg`/`jsdelivr`/`browser` field, so this URL serves `dist/index.cjs`. In a `<script>` tag that dies on the first `require()` and sets no global. The ESM output needs `type="module"` and an import map for the bare specifiers, so it isn't a drop-in replacement. I'd drop the section rather than try to document a working incantation.

packages/react/README.md
```diff
@@ -128,44 +123,51 @@ Possible values:
...
-  for real-time visual updates.
-- Provides a `storyblokEditable` function to link editable components to the Storyblok Visual
-  Editor.
+- `apiClient` — a typed API client for fetching content from the
+  [Storyblok CDN API](https://www.storyblok.com/docs/api/content-delivery/v2/getting-started/introduction?utm_source=github.com&utm_medium=readme&utm_campaign=storyblok-react).
+- `StoryblokComponent` — renders any registered block component by matching `block.component` to the
+  components map.
+- `storyblokEditable` — attaches the required attributes to link a component to the Storyblok Visual
+  Editor for
+  [real-time visual updates](https://www.storyblok.com/docs/Guides/storyblok-latest-js?utm_source=github.com&utm_medium=readme&utm_campaign=storyblok-react).
```
maoberlehner: Leftover from `storyblokInit`. There's no initialize step now, and `apiClient` isn't part of this package at all — `apiClient` has zero hits in `src/`, and the section right above tells you to build one with `createApiClient` from `@storyblok/api-client`. `defineStoryblokComponents` returns `StoryblokComponent` and `StoryblokRichText`; the latter is missing from this list. `storyblokEditable` is just a named export, not something the call hands back.

packages/react/playground/next-latest/src/lib/storyblok.tsx
```diff
@@ -0,0 +1,28 @@
...
+import Feature from "@/app/components/storyblok/Feature";
+import Grid from "@/app/components/storyblok/Grid";
+import Page from "@/app/components/storyblok/Page";
+import Teaser from "@/app/components/storyblok/Teaser";
+import { WeatherWidget } from "@/app/components/storyblok/WeatherWidget";
+import { WeatherWidgetSkeleton } from "@/app/components/storyblok/WeatherWidgetSkeleton";
+
+export const apiClient = createApiClient({
+  accessToken: "OurklwV5XsDJTIE1NJaD2wtt",
+});
```
maoberlehner, line 13: This token is hardcoded in all five playgrounds with no env fallback, so pointing any of them at a QA space means editing source. Astro already does `process.env.STORYBLOK_ACCESS_TOKEN ?? "<demo>"`; nuxt gets it for free through `runtimeConfig`. Next and Vite don't, so we'd need the explicit fallback here.

Related: react is the only one of the three with no Visual Editor QA harness — no `test/visual-editor/qa.config.ts`, no `test/GUIDE.md`, no scenario, and no dep on the shared tool from #764. The `e2e/index.spec.ts` we do have is two assertions against the demo space.

Not this PR's doing and not a blocker, but with the SDK being redesigned it's a good moment to add it. I have a `has-playground-content` scenario built already (components plus 6 stories under `react/` slugs, mirroring nuxt's).

maoberlehner (changes requested):
Added some more comments.

Additionally: The description has drifted from the code. It documents `createApiClient` as an `@storyblok/react` export (it's `@storyblok/api-client`), a `StoryblokPreviewRsc` component that doesn't exist in the branch, and `/rsc` as removed when it's still declared in `package.json` and is the RSC entry point. Smaller ones: the client example uses a children render-prop instead of `renderContent`, `StoryblokBlockData` is listed under new types but isn't exported, and `Story` comes from the package's own generated types, not re-exported from `@storyblok/api-client`.

packages/react/playground/next-latest-static/src/app/components/Grid.tsx
```diff
@@ -1,11 +1,12 @@
...
+import { storyblokEditable } from "@storyblok/react";
 
-const Grid = ({ blok }: { blok: any }) => {
-  return (
-    <h2 data-cy="teaser" {...storyblokEditable(blok)}>
-      This is a Grid component
-    </h2>
-  );
-};
+type GridProps = StoryblokComponentProps<object>;
```
maoberlehner: StoryblokComponentProps<object> doesn't satisfy the new T extends Record<string, unknown>. object has no index signature. Grid has no extra fields, so the bare StoryblokComponentProps works.

packages/react/src/client/storyblok-preview-server-runtime.tsx
```diff
@@ -0,0 +1,260 @@
...
+      run(updatedStory);
+    },
+    { debounceMs, bridgeOptions },
+  );
+
+  // No editor update yet — return children with zero state involvement.
+  // This is the initial SSR/streaming path: Suspense boundaries inside the
+  // children tree (e.g. WeatherWidget) fire normally because nothing here
+  // interferes with them.
+  if (!livePromise) {
```
maoberlehner: The early return means the first editor event changes the root element type from `Fragment` to `LiveContentBoundary`, so React tears down the whole children tree instead of reconciling it. I put a counter in both `children` and `renderContent` and watched it reset to 0 on event 1 and survive on event 2 — so the very first keystroke wipes client state (`EmojiRandomizer` in the playground, and anything stateful a user has: carousels, open menus, playing video).

Giving the content one stable position fixes it — move the null check into `LiveContent` (`promise ? reactUse!(promise) : children`) and render the boundary + Suspense unconditionally. `LiveContentBoundaryProps.promise` widens to `| null`, `getDerivedStateFromProps` is unaffected. Counter survives event 1 after that, and the suite still passes.

packages/react/src/client/use-storyblok-state.ts
```diff
@@ -0,0 +1,45 @@
...
+ * ```
+ */
+export function useStoryblokState(story: Story, options: UseStoryblokStateOptions = {}): Story {
+  const [current, setCurrent] = useState(story);
+
+  // Sync the prop into state when the story identity changes.
+  // useState only uses the initial value on mount, so cross-route navigation
+  // or SWR keepPreviousData swaps that change `story.id` would otherwise render
+  // stale content forever on a reused component instance.
+  useEffect(() => {
```
maoberlehner, line 40: Keying the sync effect on `story.id` alone means a new story object with the same id never lands in state. That's intentional — there's a test for it — but it also means `router.refresh()`, ISR revalidation, and SWR refetches are dropped for as long as the component stays mounted.

Both playgrounds work around it with `key={story.uuid}`, which suggests the constraint isn't obvious. Either widen the dep to something that moves on content change (`published_at`, a revision) or say in the docs that remounting via `key` is required.

packages/react/scripts/generate.ts
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env -S node --experimental-strip-types --no-warnings=ExperimentalWarning
+/**
+ * Generates the block-data types for `@storyblok/react` from the pinned
+ * OpenAPI overlay spec. Only `BlockContentBase` is requested — the minimal
+ * set needed to back `BlockContent`.
```
maoberlehner, line 5: Says only `BlockContentBase` is requested, but the call below is `include: ["Story", "BlockContent"]`.

PR #788, opened 2026-08-26
feat(nuxt): small nice to have improvements

JakubAndrzejewskiSB:
## useAsyncStoryblok composable — cache key separator
uniqueKey concatenated the stringified API params directly with the URL (`${stableStringify(api)}${url}`), with no separator between them. Different url/api combinations could theoretically produce colliding useAsyncData cache keys. Fixed by inserting an explicit separator: `${url}::${stableStringify(api)}`.

## serverStoryblokClient — missing return type
The function had no explicit return type annotation, so its return type was left to inference. Added : StoryblokClient to the signature for a stable, explicit public contract.

## event.context._storyblokClient — implicit any
The server client was cached on event.context._storyblokClient with no type declared for that field, so it was implicitly any everywhere it was read or written. Added an H3 module augmentation (declare module "h3" { interface H3EventContext { _storyblokClient?: StoryblokClient } }) so the cached client is properly typed.

## package.json — no engines field
Nothing constrained the supported Node.js version for the package. Added "engines": { "node": "^20.19.0 || >=22.12.0" }, matching Nuxt 4's own supported range (the package's peer framework).

## README.md — no usage example
The README only linked out to the external docs site with no inline example. Added a short "Usage" section: install command, minimal nuxt.config.ts, and a minimal page snippet using useAsyncStoryblok.

### DIFF
--- a/packages/nuxt/package.json
+++ b/packages/nuxt/package.json
@@ -74,6 +74,9 @@
       }
     ]
   },
+  "engines": {
+    "node": "^20.19.0 || >=22.12.0"
+  },
   "nx": {
     "implicitDependencies": [
       "!@storyblok/playground-*"
--- a/packages/nuxt/src/runtime/composables/useAsyncStoryblok.test.ts
+++ b/packages/nuxt/src/runtime/composables/useAsyncStoryblok.test.ts
@@ -59,6 +59,7 @@ describe("useAsyncStoryblok", () => {
     mockRuntimeConfig.public.storyblok = { accessToken: "test-token" };
     apiGet.mockClear();
     useStoryblokBridge.mockClear();
+    vi.mocked(useAsyncData).mockClear();
   });
 
   it("throws when the access token is not available client-side (regression)", async () => {
--- a/packages/nuxt/src/runtime/composables/useAsyncStoryblok.test.ts
+++ b/packages/nuxt/src/runtime/composables/useAsyncStoryblok.test.ts
@@ -52,6 +52,7 @@ vi.mock("vue", () => ({
 }));
 
 import { useAsyncStoryblok } from "./useAsyncStoryblok";
+import { useAsyncData } from "#app";
 
 describe("useAsyncStoryblok", () => {
   beforeEach(() => {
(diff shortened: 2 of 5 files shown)


packages/nuxt/src/runtime/composables/useAsyncStoryblok.test.ts
```diff
@@ -98,6 +99,13 @@ describe("useAsyncStoryblok", () => {
     });
   });
 
+  it("separates the url and stringified api params in the cache key (regression)", async () => {
+    await useAsyncStoryblok("home", { api: { version: "draft" } });
+    const key = (vi.mocked(useAsyncData).mock.calls[0]![0] as () => string)();
```
maoberlehner, line 105: I think we don't reset `useAsyncData` between tests, this could become a problem.

packages/nuxt/package.json
```diff
@@ -74,6 +74,9 @@
       }
     ]
   },
+  "engines": {
+    "node": "^20.19.0 || >=22.12.0"
+  },
```
maoberlehner, line 79: Claude tells me:

> engines.node: "^20.19.0 || >=22.12.0" (from Nuxt 4) contradicts peerDependencies.nuxt: "^3.15.0 || ^4.0.0". Nuxt 3 users on Node 18.20.x / 20.9–20.18 / 22.0–22.11 get an engine warning on npm, a hard failure with engine-strict (which this repo's own .npmrc sets). Widen to the union, or drop Nuxt 3 from the peer range. No other packages/* declares engines.

Can we add it in a way that avoids warnings? Is it worth the hassle to add it?

maoberlehner (changes requested):

packages/nuxt/package.json
JakubAndrzejewskiSB: Hmm, this was added purely because it is said in the documentation but not addressed in any way in the code - so someone reading the docs can assume that these versions should be used while the code ignores it completely.

packages/nuxt/src/runtime/composables/useAsyncStoryblok.test.ts
JakubAndrzejewskiSB: Good catch. Fixed!

packages/nuxt/package.json
maoberlehner: I leave it to you, I don't have a strong opinion either way.

maoberlehner (approved):
LGTM!

packages/nuxt/package.json
JakubAndrzejewskiSB: We can always remove it afterwards but at this point I think it is better to have it to be up to date with the docs :)

PR #791, opened 2026-08-28
feat(cli): confirm success at the end of schema init and push

maoberlehner:
Feedback on `schema init`: a successful run ends with one warning and two info lines, and no confirmation that anything worked. It reads like the command half-failed.

`schema push` has the same shape. A run that surfaced stale-entity or migration warnings ends on those warnings, with only a spinner line to say the push succeeded.

## What

Both commands now close with an explicit success block, printed after every advisory line.

**`schema init`**

```
✔ Generated 12 files
  .storyblok/schema/blocks/hero.ts
  ...

 Success

✔ Schema workspace initialized from space 12345 in .storyblok/schema

ℹ Next steps:
  1. Install @storyblok/schema in the project that imports these files
  2. Review the generated files. Your local schema is now the source of truth
  3. Use storyblok schema push for all further schema changes. schema init is a one-time bootstrap
```

The bootstrap caveat and the "review before continuing" nudge are no longer a warning. Nothing actually went wrong, so they become numbered next steps instead.

**`schema push`**

```
⚠️  1 stale entity(s) exist remotely but not in schema. Use --delete to remove.

 Success

✔ Schema pushed to space 12345. 1 created, 0 updated.
```

The confirmation moved below the local-component write step so it is the last thing on screen. The up-to-date case gets the same treatment: `Schema already up to date with space 12345, nothing to push.`

Both use the existing `ui.ok(message, true)` helper, matching `login`, `logout`, `languages`, `types generate`, and `create`.

Fixes DX-563

### DIFF
--- a/packages/cli/src/commands/schema/init/index.test.ts
+++ b/packages/cli/src/commands/schema/init/index.test.ts
@@ -356,13 +356,19 @@ describe("schema init command", () => {
     expect(files.some((f) => f.endsWith("/schema.ts"))).toBe(true);
   });
 
-  it("should warn that init is a bootstrap step", async () => {
+  it("should confirm success and print next steps", async () => {
     preconditions.hasEmptyRemote();
 
     await schemaCommand.parseAsync(["node", "test", "init", "--space", DEFAULT_SPACE]);
 
-    expect(console.warn).toHaveBeenCalledWith(expect.stringContaining("bootstrap step"));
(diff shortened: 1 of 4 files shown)


alexjoverm (approved):
DX niceties! 

Just one thing - theoretically this is a `fix:` (it's not adding any new feat, just correcting some behaviour)

dipankarmaikap (approved):

PR #795, opened 2026-09-01
chore(live-preview,capi-client): add "sideEffects": false

dipankarmaikap:
## Summary

Adds `"sideEffects": false` to `@storyblok/live-preview`, `@storyblok/api-client`, and `@storyblok/management-api-client`. `@storyblok/richtext` already carries this flag.

Also switches `@storyblok/live-preview` to `unbundle: true` (per-module dist output), which the flag needs to pay off; `@storyblok/api-client` and `@storyblok/management-api-client` already build that way.

## Why

Without the flag on dependency packages, bundlers (Webpack, Rollup, esbuild) cannot prove those modules are inert and retain the full request layer — `getValidRequestBody`, `isHTTPError`, `isTimeoutError`, `timeout`, `delay` — even when only an unrelated export such as `storyblokEditable` is imported. The flag must be present on the package that _owns_ the modules for tree-shaking to work across the package boundary.

Measured against an entry importing only `storyblokEditable`:

| | Bundle (minified) |
|---|---|
| Before | 37,270 B |
| After | 401 B |

~98.9% reduction.

### DIFF
--- a/packages/capi-client/package.json
+++ b/packages/capi-client/package.json
@@ -16,6 +16,7 @@
     "directory": "packages/capi-client"
   },
   "type": "module",
+  "sideEffects": false,
   "main": "./dist/index.cjs",
   "module": "./dist/index.mjs",
   "types": "./dist/index.d.cts",
--- a/packages/live-preview/package.json
+++ b/packages/live-preview/package.json
@@ -22,6 +22,7 @@
     "dist"
   ],
   "type": "module",
+  "sideEffects": false,
   "main": "./dist/index.cjs",
   "module": "./dist/index.mjs",
   "types": "./dist/index.d.cts",


maoberlehner (changes requested):
Nice improvement!
Two findings:
- packages/live-preview/vite.config.ts: the PR description says both packages use unbundle: true, but live-preview's pack config lacks it and ships a single dist/index.mjs. The flag alone won't enable per-module dropping there. Either add unbundle: true or fix the description.
- packages/mapi-client/package.json: @storyblok/management-api-client shares the same generated request layer and lacks sideEffects: false. The same retention problem persists there; the fix is the identical one-line flag.

maoberlehner (approved):
Double-checked the potential improvements thanks to tree-shaking, and they're substantial, nice!

Minor: PR description was out of date. It said, "Both packages already use unbundle: true" I fixed it for you.

Same improvement could also added to at least:  @storyblok/react, @storyblok/schema,  @storyblok/vue, @storyblok/nuxt, @storyblok/astro
I leave it up to you to add it here, add a follow up ticket, or skip it completely.

PR #798, opened 2026-09-02
fix(capi-client): support excluding_story_fields

dipankarmaikap:
## Summary

- add `excluding_story_fields` to Content API story requests
- omit the parameter for empty arrays and export its field type
- cover request serialization and related-story forwarding

## Verification

- `pnpm nx lint @storyblok/api-client`
- `pnpm nx run @storyblok/api-client:test:types`
- `pnpm exec vp test run packages/capi-client/src/resources/stories.test.ts packages/capi-client/src/utils/fetch-rel-uuids.test.ts`

### DIFF
--- a/packages/capi-client/src/resources/stories.test.ts
+++ b/packages/capi-client/src/resources/stories.test.ts
@@ -277,6 +277,25 @@ describe("stories.get()", () => {
       "translated_slugs",
     );
   });
+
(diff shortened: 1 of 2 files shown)


dipankarmaikap:
Closing as redundant: the CAPI excluding_story_fields implementation is already present on main, so this branch produces no file diff.

PR #800, opened 2026-09-02
test(cli): type-check test files

dipankarmaikap:
## Summary

- include CLI test files in the package TypeScript project
- align test fixtures and mocks with current API contracts
- retain malformed runtime API-data coverage for null and missing fields

## Verification

- `pnpm --filter storyblok exec vp test run`
- `pnpm nx lint storyblok`
- `pnpm nx run storyblok:test:types`

### DIFF
--- a/packages/cli/src/commands/assets/utils.test.ts
+++ b/packages/cli/src/commands/assets/utils.test.ts
@@ -45,7 +45,13 @@ describe("internalTagNamesFromAssets", () => {
           { id: 2, name: "green" },
         ],
       },
-      { internal_tags_list: [{ id: 3, name: "blue" }, { id: 4, name: "" }, { id: 5 }] },
+      {
+        internal_tags_list: [
+          { id: 3, name: "blue" },
+          { id: 4, name: "" },
+          { id: 5, name: "" },
+        ],
+      },
       {},
     ]);
 
--- a/packages/cli/src/commands/assets/utils.test.ts
+++ b/packages/cli/src/commands/assets/utils.test.ts
@@ -38,7 +38,7 @@ describe("extractAssetSizeFromFilename", () => {
 
 describe("internalTagNamesFromAssets", () => {
   it("collects unique tag names in first-seen order, ignoring blanks", () => {
-    const names = internalTagNamesFromAssets([
+    const assets = [
       {
         internal_tags_list: [
           { id: 1, name: "blue" },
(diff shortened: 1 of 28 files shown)


packages/cli/src/commands/types/generate/actions.test.ts
```diff
@@ -1123,16 +1139,14 @@ describe("component property type annotations", () => {
   });
 
   it("should skip malformed schema entries instead of failing", async () => {
-    // The management API types every schema entry as an object, but malformed spaces
-    // do occur. A single bad entry must not take down the whole command.
-    const componentWithMalformedEntry: Component = {
+    const componentWithMalformedEntry = {
```
maoberlehner, line 1144: I think this is one of the few cases where typecasting would be perfectly appropriate. `const componentWithMalformedEntry = {} as Component` (while keeping the comment explaining why it is done).

packages/cli/src/lib/validation/validation.test.ts
```diff
@@ -8,7 +8,7 @@ import { formatJson } from "./format-json";
 import { formatPretty } from "./format-pretty";
 import { entityToHeader, entityToRef, groupIssuesByEntity } from "./group";
 import { writeValidationReport } from "./report";
-import type { Reporter } from "../reporter/reporter";
+import { Reporter } from "../reporter/reporter";
```
maoberlehner: Not sure if I agree with this change. Why don't we continue to use the fake reporter?

packages/cli/src/commands/components/push/graph-operations/__tests__/test-data.ts
maoberlehner: Seems like this is dead code. So better remove it completely while you're at it.

packages/cli/src/commands/assets/utils.test.ts
```diff
@@ -38,16 +38,19 @@ describe("extractAssetSizeFromFilename", () => {
...
         ],
       },
-      { internal_tags_list: [{ id: 3, name: "blue" }, { id: 4, name: "" }, { id: 5 }] },
+      {
+        internal_tags_list: [{ id: 3, name: "blue" }, { id: 4, name: "" }, { id: 5 }],
+      },
       {},
-    ]);
+    ];
+    const names = Reflect.apply(internalTagNamesFromAssets, undefined, [assets]);
```
maoberlehner: The usage of `Reflect` seems a little hacky to me.

Claude suggests this instead:

```ts
const assets: Parameters<typeof internalTagNamesFromAssets>[0] = [ … ];
const names = internalTagNamesFromAssets(assets);
```

does this work?

maoberlehner: General thought: if types can't be trusted, we might need to make a larger architectural change throughout the whole monoblok to always validate, e.g., with something like `toAsset(asset: unknown): Asset`.

Right now, we pretend methods only take strict types, then make them work with structures that don't match those strict types. That's not ideal.

But more of a general thought that we should keep in mind: I think it would be too big of a change for this PR.

For this PR, I think I'd prefer avoiding `Reflect` and going with type casting instead.

maoberlehner (changes requested):

maoberlehner (approved):
LGTM

PR #801, opened 2026-09-02
fix(openapi): deduplicate generated union types

### LINKED ISSUE #792: [Bug]: New richtext types generated by CLI contain repetitions
### Package

storyblok (CLI)

### Bug Description

New richtext types contain repetitions (the same type is listed more than once as an expected type for a field).

### Steps to Reproduce

Run `storyblok types generate`. I don't think the specific flags matter.

### Reproduction URL

_No response_

### Expected Behavior

I noticed the following two repetitions in the generated types:

1. Repeated `string` under `[key: string]`:
```ts
type BlockContentInputRoot = {
  /**
   * Unique identifier for this component instance (auto-generated when omitted)
   */
  _uid?: string;
  /**
   * Technical name of the component (content type or nestable block)
   */
  component: string;
  /**
   * Storyblok editor markup string for inline editing (present in draft/preview mode)
   */
  _editable?: string;
  [key: string]: null | string | number | boolean | Array<string | AssetFieldValueRoot | BlockContentInputRoot> | AssetFieldValueRoot | MultilinkFieldValueRoot | TableFieldValueRoot | RichTextFieldValueRoot | PluginFieldValueRoot | string | undefined;
};
```

2. Repeated `null` and `string` under `meta_data`:
```ts
type RichTextFieldValueImageNode = {
  type: 'image';
  attrs: {
    /**
     * Storyblok asset ID
     */
    id: number | null;
    /**
     * Image URL
     */
    src: string | null;
    /**
     * Alternative text
     */
    alt: string | null;
    title: string | null;
    source: string | null;
    copyright: string | null;
    /**
     * Asset metadata
     */
    meta_data: {
      alt?: string | null;
      title?: string | null;
      source?: string | null;
      copyright?: string | null;
      [key: string]: string | null | string | null | string | null | string | null | string | null | undefined;
    } | null;
  };
};
```

### Actual Behavior

Generated types should not contain repetitions.

### Environment

```shell
System:
    OS: macOS 26.6.2
    CPU: (11) arm64 Apple M3 Pro
    Memory: 91.50 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.14.0 - /Users/angelikacathor/.asdf/installs/nodejs/22.14.0/bin/node
    npm: 11.2.0 - /Users/angelikacathor/.asdf/plugins/nodejs/shims/npm
    pnpm: 10.21.0 - /Users/angelikacathor/.asdf/installs/nodejs/22.14.0/bin/pnpm
    bun: 1.2.15 - /Users/angelikacathor/.bun/bin/bun
  Browsers:
    Brave Browser: 152.1.94.117
    Chrome: 152.0.7977.65
    Edge: 152.0.4191.53
    Firefox: 154.0.1
    Firefox Developer Edition: 155.0
    Safari: 26.6.2
  npmPackages:
    @storyblok/astro: 10.2.3 => 10.2.3 
    @storyblok/js: 6.3.1 => 6.3.1 
    astro: ^7.1.6 => 7.1.6 
    storyblok: ^4.22.3 => 4.22.3 
    svelte: ^5.56.8 => 5.56.8
```

### Error Logs

```shell

```

### Additional Context

_No response_

dipankarmaikap:
## Summary

This PR fixes duplicate union members in generated OpenAPI TypeScript output and updates the generator to `@hey-api/openapi-ts@0.99.0`.

### Commit 1: `fix(openapi): address codegen review feedback`

- Deduplicates unions in full SDK `types.gen.ts` output.
- Reuses TypeScript printers during AST processing.
- Preserves the first duplicate union member and its documentation.
- Combines reference rewriting and union deduplication into one AST pass.
- Removes unnecessary transformation casts with TypeScript node guards.
- Documents why the generator version is pinned and how the deduplication pass complements upstream behavior.

### Commit 2: `chore(openapi): update hey-api generator`

- Upgrades `@hey-api/openapi-ts` from `0.92.4` to `0.99.0`.
- Regenerates every OpenAPI consumer.
- Adapts to Hey API's optional request/response result contract for transport failures.
- Preserves CAPI `network-first` cache fallback for transport errors.
- Updates MAPI retry types to use `ky` directly.
- Updates richtext and schema consumers for Hey API's normalized generated names.
- Updates richtext generation to recognize the new node/mark names.
- Keeps the stricter asset validation diagnostics introduced by the regenerated Zod schema.
- Adds the required published-version discriminator to a CAPI inline-relations fixture.

## What Changed

Hey API 0.99 removes redundant union members in raw generated output and normalizes several generated names and index signatures. It also makes generated request/response fields optional because transport failures may not produce an HTTP response.

The codegen post-processor remains necessary because reference rewriting can create duplicate union members after aliases are applied.

## Verification

- Full OpenAPI regeneration for all seven consumers.
- `@storyblok/openapi-codegen`: 25 tests passed, typecheck, lint, and build.
- `@storyblok/api-client`: 203 tests passed, typecheck, build.
- `@storyblok/management-api-client`: 139 tests passed, build.
- `@storyblok/richtext`: 311 tests passed, typecheck, build.
- `@storyblok/schema`: 204 tests passed, typecheck, lint, build.
- `@storyblok/live-preview`: 44 tests passed, build.
- CLI: 1,292 tests passed, 1 skipped, 1 todo.
- MAPI integration type tests passed.
- Frozen offline install passed.

Fixes DX-564
Fixes #792

### DIFF
--- a/tools/openapi-codegen/src/transform.test.ts
+++ b/tools/openapi-codegen/src/transform.test.ts
@@ -70,6 +70,27 @@ describe("transformGeneratedFile", () => {
     expect(output).not.toContain(": Content ");
   });
 
+  it("should deduplicate union members after rewriting references", () => {
+    const source =
+      "export type Content = { [key: string]: string | null | string | null | Source | Source };";
+
+    const { output } = run(
+      source,
+      [{ source: "Content", emitAs: "Content" }],
+      new Map([["Source", "AssetSource"]]),
+    );
+
+    expect(squish(output)).toContain("[key: string]: string | null | AssetSource;");
+  });
+
+  it("should preserve distinct literal union members", () => {
+    const source = "export type Heading = { level: 1 | 2 | 3 | 4 | 5 | 6 | null };";
+
+    const { output } = run(source, [{ source: "Heading", emitAs: "Heading" }]);
+
+    expect(squish(output)).toContain("level: 1 | 2 | 3 | 4 | 5 | 6 | null;");
+  });
+
   it("should unwrap a single envelope property", () => {
     const source =
       "export type UpdateStoryRequest = { story: { name: string }; force_update?: boolean };";
--- a/tools/openapi-codegen/src/transform.ts
+++ b/tools/openapi-codegen/src/transform.ts
@@ -111,7 +111,7 @@ export function transformGeneratedFile(
       }
     }
 
-    body = rewriteRefs(body, renameMap);
+    body = deduplicateUnions(rewriteRefs(body, renameMap), sourceText);
 
     const newDecl = withLeadingComments(
       sourceText,
@@ -230,6 +230,35 @@ function rewriteRefs(node: ts.TypeNode, renameMap: ReadonlyMap<string, string>):
           }
         }
         return ts.visitEachChild(n, visit, ctx);
+      };
+      return (n) => ts.visitNode(n, visit) as ts.TypeNode;
+    },
+  ]);
+  const transformed = result.transformed[0] as ts.TypeNode;
+  result.dispose();
+  return transformed;
+}
+
+/** Remove structurally identical members from unions emitted by @hey-api/openapi-ts. */
+function deduplicateUnions(node: ts.TypeNode, sourceText: string): ts.TypeNode {
+  const printer = ts.createPrinter({ removeComments: true });
+  const source = ts.createSourceFile("union.ts", sourceText, ts.ScriptTarget.Latest, false);
+  const result = ts.transform(node, [
+    (ctx) => {
+      const visit: ts.Visitor = (n) => {
+        const visited = ts.visitEachChild(n, visit, ctx);
+        if (!ts.isUnionTypeNode(visited)) {
+          return visited;
+        }
+
+        const members = new Map<string, ts.TypeNode>();
+        for (const member of visited.types) {
+          const key = printer.printNode(ts.EmitHint.Unspecified, member, source);
+          members.set(key, member);
+        }
+        return members.size === visited.types.length
+          ? visited
+          : ctx.factory.createUnionTypeNode([...members.values()]);
       };
       return (n) => ts.visitNode(n, visit) as ts.TypeNode;
     },
--- a/packages/capi-client/src/generated/mapi/types.gen.ts
+++ b/packages/capi-client/src/generated/mapi/types.gen.ts
@@ -170,9 +170,9 @@ export type Space = {
      * Whether the Joyride trial traffic limit has been exceeded.
      */
     joyride_traffic_limit_exceeded?: boolean;
-    [key: string]: unknown | number | string | string | null | number | null | number | null | boolean | number | null | number | null | string | null | string | null | string | null | string | null | string | null | string | null | string | null | {
+    [key: string]: unknown | number | string | null | boolean | {
         [key: string]: boolean | null;
-    } | string | null | Array<SpaceRole> | string | null | string | null | undefined;
+    } | Array<SpaceRole> | undefined;
 };
 
 /**
--- a/packages/capi-client/src/generated/overlay/_internal.gen.ts
+++ b/packages/capi-client/src/generated/overlay/_internal.gen.ts
@@ -866,7 +866,7 @@ export type BlockContentRoot = {
      * Storyblok editor markup string for inline editing (present in draft/preview mode)
      */
     _editable?: string;
-    [key: string]: string | number | boolean | Array<string | AssetFieldValueRoot | BlockContentRoot> | AssetFieldValueRoot | MultilinkFieldValueRoot | TableFieldValueRoot | RichTextFieldValueRoot | PluginFieldValueRoot | string | undefined;
+    [key: string]: string | number | boolean | Array<string | AssetFieldValueRoot | BlockContentRoot> | AssetFieldValueRoot | MultilinkFieldValueRoot | TableFieldValueRoot | RichTextFieldValueRoot | PluginFieldValueRoot | undefined;
 };
 
 /**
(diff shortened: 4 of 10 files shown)


tools/openapi-codegen/src/transform.ts
```diff
@@ -111,7 +111,7 @@ export function transformGeneratedFile(
       }
     }
 
-    body = rewriteRefs(body, renameMap);
+    body = deduplicateUnions(rewriteRefs(body, renameMap), sourceText);
```
maoberlehner: `<sdkDir>/types.gen.ts` (`emitFullSdk`) still has duplicates. Would it be easy to deduplicate there too?

tools/openapi-codegen/src/transform.ts
```diff
@@ -238,3 +238,32 @@ function rewriteRefs(node: ts.TypeNode, renameMap: ReadonlyMap<string, string>):
   result.dispose();
   return transformed;
 }
+
+/** Remove structurally identical members from unions emitted by @hey-api/openapi-ts. */
+function deduplicateUnions(node: ts.TypeNode, sourceText: string): ts.TypeNode {
+  const printer = ts.createPrinter({ removeComments: true });
```
maoberlehner: nitpick: we could create the printer once, but idk if it is an expensive operation, so maybe it's fine to leave it as is.

tools/openapi-codegen/src/transform.test.ts
```diff
@@ -70,6 +70,27 @@ describe("transformGeneratedFile", () => {
     expect(output).not.toContain(": Content ");
   });
 
+  it("should deduplicate union members after rewriting references", () => {
+    const source =
+      "export type Content = { [key: string]: string | null | string | null | Source | Source };";
```
maoberlehner: Is it on purpose that `source` has no `AssetSource`?

tools/openapi-codegen/src/transform.ts
```diff
@@ -238,3 +238,32 @@ function rewriteRefs(node: ts.TypeNode, renameMap: ReadonlyMap<string, string>):
...
+      const visit: ts.Visitor = (n) => {
+        const visited = ts.visitEachChild(n, visit, ctx);
+        if (!ts.isUnionTypeNode(visited)) {
+          return visited;
+        }
+
+        const members = new Map<string, ts.TypeNode>();
+        for (const member of visited.types) {
+          const key = printer.printNode(ts.EmitHint.Unspecified, member, source);
+          members.set(key, member);
```
maoberlehner: Claude:

> Last duplicate wins, JSDoc lost (transform.ts:257). Keys are computed with removeComments, so a later comment-less duplicate replaces an earlier documented member. No current output affected, but x-enum-descriptions would break silently. Use if (!members.has(key)).

tools/openapi-codegen/src/transform.ts
```diff
@@ -238,3 +238,32 @@ function rewriteRefs(node: ts.TypeNode, renameMap: ReadonlyMap<string, string>):
   result.dispose();
   return transformed;
 }
+
+/** Remove structurally identical members from unions emitted by @hey-api/openapi-ts. */
```
maoberlehner, line 298: Claude:

> Dead unknown | ... index signatures remain (transform.ts:242). hey-api 0.93.1 fixed that case upstream; repo pins 0.92.4, latest is 0.99.0. A bump complements rather than replaces this pass since 0.99.0 still emits the nested duplicates from #792. PR should note why no bump and when the workaround can go.

maoberlehner: I think we should check if we can bump to 0.99.0

tools/openapi-codegen/src/transform.ts
```diff
@@ -238,3 +238,32 @@ function rewriteRefs(node: ts.TypeNode, renameMap: ReadonlyMap<string, string>):
...
+function deduplicateUnions(node: ts.TypeNode, sourceText: string): ts.TypeNode {
+  const printer = ts.createPrinter({ removeComments: true });
+  const source = ts.createSourceFile("union.ts", sourceText, ts.ScriptTarget.Latest, false);
+  const result = ts.transform(node, [
```
maoberlehner, line 305: Claude:

> Duplicated transform scaffold (transform.ts:246). ts.transform(body, [renameFactory, dedupFactory]) gives identical output and makes ordering explicit.

tools/openapi-codegen/src/transform.ts
```diff
@@ -238,3 +238,32 @@ function rewriteRefs(node: ts.TypeNode, renameMap: ReadonlyMap<string, string>):
...
+        const members = new Map<string, ts.TypeNode>();
+        for (const member of visited.types) {
+          const key = printer.printNode(ts.EmitHint.Unspecified, member, source);
+          members.set(key, member);
+        }
+        return members.size === visited.types.length
+          ? visited
+          : ctx.factory.createUnionTypeNode([...members.values()]);
+      };
+      return (n) => ts.visitNode(n, visit) as ts.TypeNode;
```
maoberlehner: Claude:

> as casts (transform.ts:263,266). Copied from rewriteRefs. Line 266 cast is removable as-is; line 263 needs ts.isTypeNode plus ?? n.

maoberlehner (changes requested):

packages/mapi-client/src/client.ts
```diff
@@ -199,7 +200,7 @@ const createManagementApiClientBase = <DefaultThrowOnError extends boolean = fal
   );
 
   client.interceptors.error.use(
-    (error: unknown, response: Response) =>
+    (error: unknown, response?: Response) =>
```
maoberlehner, line 203: QA finding:

> This marks the response optional here, but ApiResponse (client.ts:32-36) still declares response: Response; request: Request on every branch. On a transport failure hey-api 0.99 resolves with both undefined, so res.response.status throws a TypeError the types say can't happen — verified against an unresolvable host and a timeout.
>
> capi-client handles this with result.response ?? Response.error() in requestNetwork; worth mirroring that here, or relaxing ApiResponse to response?: Response.

packages/capi-client/src/utils/inline-relations.test.ts
```diff
@@ -42,6 +42,7 @@ const makeStory = (
     translated_slugs: [],
     updated_at: "2024-01-01T00:00:00.000Z",
     uuid,
+    version: "published",
```
maoberlehner, line 45: I just let Claude do some testing, this seems to be masking a real error: the CDN and MAPI never send a version.

> Cause is the 0.92.4 → 0.99.0 bump alone. our cdn-v2.openapi.yaml declares discriminator: {propertyName: version} on Story, while neither DraftStory nor PublishedStory defines that property or lists it in required. OpenAPI 3.1: "the discriminator field MUST be a required field" — so 0.99 is right and the spec is wrong; 0.92 was just lenient.
>
> Same pattern hits App.public in the MAPI types.

I think this should be fixed at the source in the OpenAPI spec.

maoberlehner (changes requested):

dipankarmaikap:
Closing this in favor of #825.

PR #806, opened 2026-09-11
fix: dynamic block content types disagree on null, and stories.get()'s story type diverges from Story

dipankarmaikap:
## Summary

Two related type-only bugs found while testing the React SDK against a real consumer app, both stemming from `story.content`'s dynamic/untyped field-value type.

1. **`schema`/`block-content.yaml`** — the CDN API returns `null` (not an absent key) for any optional field with no value, but the shared overlay spec's `BlockContentRoot.additionalProperties` schema didn't allow `null` alongside the field-type unions. Meanwhile the schema-narrowed, per-block content type (built from the same spec's field definitions) already types every optional field as `T | null`. Net effect: a schema-typed block's content was never assignable to the loose/dynamic content type used for generic block handling — e.g. passing a specific block's content into a generic "render this block" helper failed to type-check.

   Fix: add `null` to the `additionalProperties.oneOf` union in `tools/openapi-codegen/specs/shared/stories/block-content.yaml`, and regenerate every consumer (`api-client`, `live-preview`, `mapi-client`, `schema`).

2. **`capi-client/client.ts`** — `createApiClientBase` has no return-type annotation, so when the DTS bundler emits this package's `.d.ts` it has to independently re-derive and print `stories.get()`'s full structural return type. That re-derivation diverges from how `ContentApiClient["stories"]` references the exact same `ReturnType<typeof createStoriesResource<...>>` expression elsewhere in the same file, producing two differently-shaped printed types for what's meant to be the same "no schema provided" `Story`. In practice, `client.stories.get()`'s inferred `story` type failed to structurally match this package's own plain, exported `Story` type.

   Fix: name the shape (`ApiClientBaseResult`) and annotate `createApiClientBase`'s return type with it, so the bundler echoes the reference instead of re-resolving the conditional types itself.

## Verification

- `tsc --noEmit` clean in `capi-client` (and `schema`, which shares the affected generated types)
- `pnpm nx run-many -t build` clean for `api-client`, `schema`, `mapi-client`, `live-preview`
- `pnpm nx run-many -t test` — all existing tests pass (44 + 203 + 139 + 204)
- `pnpm nx run @storyblok/api-client:lint` clean, `pnpm format` — no changes needed
- Diffed the built `.d.ts` before/after: `createApiClientBase`'s ~400-line hand-expanded/duplicated inline literal for `stories.get()`'s return type collapses to a handful of named type references, and `client.stories.get()`'s `story` type now resolves to the same plain `Story` type used elsewhere.
- Reproduced against a real consumer app (swapped locally-built `dist/` output into the app's `node_modules` for `@storyblok/react`, `@storyblok/api-client`, and `@storyblok/live-preview`): went from 8 `tsc` errors to 0.

## Notes for reviewers

- `packages/react` isn't included here since it doesn't exist on `main` yet (it's on an unmerged feature branch) — this PR only touches packages that already exist on `main`.
- Not a runtime change: the CDN API's actual JSON responses are unaffected; this only fixes the TypeScript type declarations consumers see.


packages/capi-client/src/client.ts
```diff
@@ -438,12 +472,22 @@ export const createApiClientBase = <
...
+  // package's `.d.ts`. Left inferred, the bundler independently re-resolves
+  // `StoryResult`'s conditional for this call site and — for reasons specific
+  // to declaration-emit (no live type-checker, generics not yet substituted)
+  // — can pick a different, more expanded branch than the one a consumer
+  // sees when they reference the `Story`/`ContentApiClient` types directly.
+  // Concretely, that already showed up as `client.stories.get()`'s inferred
+  // `story` type failing to structurally match this package's own plain
+  // `Story` type, even though both are meant to be the same "no schema
+  // provided" shape.
+  const stories: ReturnType<
```
maoberlehner, line 461: Classic Sonnet/Opus 5 _blog article as a comment_. Please make sure the two comments don't contain info that better fits a commit message and instead are terse, useful hints for future readers.

maoberlehner (changes requested):

packages/capi-client/src/client.ts
dipankarmaikap: Fixed it.

maoberlehner (approved):
LGTM

PR #807, opened 2026-09-14
fix(cli,richtext,nuxt): resolve security scanner findings

maoberlehner:
Addresses the security scanner backlog for this repository. The scanner lists 258 findings; most are
not real. Each commit below is one class of issue, and the triage for everything left alone is in "Not
changed" so the next person does not re-derive it.

## Fixed

| Commit | Issue |
| --- | --- |
| `prevent prototype pollution in mergeDeep` | `mergeDeep` recursed into `__proto__` / `constructor`. Config layers are merged from files on disk, so a crafted config file reached `Object.prototype` and changed every object in the process. |
| `keep story file names within the target directory` | A story slug and uuid are remote data and are not constrained to a single path segment, so `stories pull` could write outside `--path`. Separators and `..` are collapsed on both the read and the write side, so stories still round-trip to the same file. |
| `only download assets over http(s)` | `downloadFile` passed a remote-supplied asset filename straight to `fetch`. The guard lives in `downloadFile` itself so every caller is covered, not just the one the scanner flagged. |
| `pin third-party actions to commit SHAs` | A mutable tag lets the action owner change what runs in CI after the fact. |
| `validate PR titles with commitlint instead of a third-party action` | The PR title check was the one third-party action running on `pull_request_target` with `statuses: write`. The title is the squash commit message, so it now goes through the same commitlint config as every commit. That also enforces the scope rule the action never checked: an unscoped title produced a squash commit that Nx attributed to every package. |
| `update dependencies with security advisories` | The advisories that reach a published dependency: `@tiptap/core`, `zod`, `yaml`. Four of roughly 150 dependency findings are in this category; the rest are dev-only. |
| `install Vite+ from the lockfile instead of a remote script` | The CI action piped a remote script into a shell, which then fetched platform tarballs with `curl`; neither hop was verified. `vite-plus` was already a pinned root devDependency, so the action installed a second copy of the same version. Dropping it leaves the lockfile-verified `vp` that package scripts already resolve. |
| `render playground API responses as text` | The js-client playground wrote API responses into `innerHTML`. Playground-only, but it is a pattern people copy out of a playground. The output element is now static in the template and the handlers only set `textContent`. |
| `update lodash past CVE-2026-4800 and CVE-2026-2950` | lodash reaches the published CLI as a production dependency of `json-schema-to-typescript`, which needed a major bump to a patched range. The remaining copies come from dev tooling, pulled forward by a workspace override. |
| `update transitive dependencies on published paths` | `ws` (via the DOM implementation the richtext package depends on at runtime), `defu` (via the CLI's config loader), `@apidevtools/json-schema-ref-parser` (via the CLI's type generator), and `@tiptap/extension-link`, which stayed behind when `@tiptap/core` was raised. |
| `keep remote names and IDs within the target directory` | A shared library ID became a directory segment straight from an API response, with nothing checking it was a number. It is validated on the raw value now; a non-numeric `--library` is rejected instead of silently sending every library's assets to the same directory. |
| `align @nuxt/kit and @nuxt/schema with nuxt` | `nuxt` 4.5 pins both to an exact version; this package still declared `^4.2.2`, and the resulting skew broke `defineNuxtConfig`'s type. |
| `drop committed SvelteKit build output` | A directory holding only generated `.svelte-kit/` output whose source is gone. The ignore rule was anchored to the package root, so it never matched the nested path. Also zeroes a preview token in a test fixture that only needs the field's shape. |
| `prompt for the token in the login debug config` | The debug config passed a literal `1234567890` to `login --token`, which reads as a credential to secret scanners and never worked as a token anyway. It now asks for the real one at launch. |
| `move the schema Astro playground to Astro 7` | Astro 6 pinned `sharp` 0.34.5 and was the last thing holding that branch in the lockfile. Astro 7 carries a patched `sharp`, `@astrojs/node` follows to 11, and the unused `@storyblok/astro` devDependency goes with it. |
| `update development dependencies with security advisories` | Declared minimums raised for astro, `@astrojs/vercel`, `@sveltejs/kit`, svelte, vue, react-router, hono, nuxt (for `@nuxt/devtools`) and next on the current line. |

Each code fix has tests covering the behavior and the traversal/scheme cases.

`pnpm-workspace.yaml` changes by exactly one line (`next16`). `pnpm update` wants to restructure that
file under `catalogMode: prefer` — it hoists everything into the default catalog, drops the comments,
and moves the Next 13 playgrounds onto the current Next. The minimums here were raised by hand and
installed with plain `pnpm install` to avoid that.

## Not changed, on purpose

- **Migration file names.** `migrations generate` builds the file name from a component name with no
  sanitiser. The name always matches an argument the operator typed, so it is not reachable today.
  Sanitizing it changes how `migrations run` matches files to components, with a real trade-off
  (distinct names can map to the same file), so that goes into its own PR. An earlier commit on this
  branch did it; a later one reverts it.
- **The remaining path-traversal findings.** All 84 were traced to their source. Roughly half are
  argv-sourced (`--path`, `--suffix`, `--out-dir`), where writing to the operator's chosen directory
  is the feature. The other half do carry remote data — component, datasource, folder and asset
  names — but each already passes through `sanitizeFilename` or a slug helper that strips path
  separators. The two `schema/init` findings are contained by two different helpers: a kebab-case
  helper for the file name, and a slug helper for the directory segments.
- **Astro dev-toolbar XSS.** False positive. The template has one caller and interpolates two
  things: a module-level SVG literal, and a render of a module-level array of string literals that
  is never exported, reassigned or mutated. The file reads no config, environment or network data,
  and the toolbar only runs under `astro dev`.
- **The leaked-secret findings.** Four distinct Storyblok *preview* tokens across playgrounds --
  read-only and draft-scoped, and necessarily shipped to the browser for a playground to work --
  plus an `invalid-token` placeholder. Moving these to `.env` only relocates the same string into
  a committed `.env.example`; making them disappear means rotating the demo spaces, which is a
  product decision rather than a security one. Each playground family also uses a different public
  env prefix, and repository secrets are not available to fork pull requests, so the plumbing would
  cost more than it returns. The one placeholder that was worth changing is in the table above.
- **Express security headers.** Stock Angular SSR scaffold. `packages/angular` ships only
  `dist/`, so the playground never enters the tarball; its project has no deploy target and no
  deployment config exists in the repo. Adding helmet would pull a production dependency into a
  localhost demo and make the reference app diverge from the scaffold users generate from.
- **Every Next.js finding.** `next` shows up under the react package's lockfile `dependencies`
  only because `autoInstallPeers` is on; it is declared as a peer dependency and never ships. The
  consumer chooses the version.
- **`@hey-api/openapi-ts`.** Low severity, build-time only, and bumping it forces a regeneration of
  every OpenAPI consumer. That belongs in its own PR with its own review.
- **`next13` catalog.** Those playgrounds exist to cover the legacy line; moving them forward would
  defeat what they test.

## Pre-existing failure, not from this branch

`@storyblok/astro:test` fails on `main` as well. `astro dev` daemonizes and exits, so
`start-server-and-test` aborts with "server closed unexpectedly". Verified by reproducing the
identical failure on an unmodified checkout, and the behavior is present in the Astro version this
branch starts from.

Two related things worth a follow-up: stale `astro dev` daemons survive a test run and squat ports
4321+, which makes later runs fail confusingly against a stale server; and `next dev` on Next 16
generates `AGENTS.md` / `CLAUDE.md` in the playground, currently untracked and unignored.

Everything else — build, test and lint across 23 projects — passes.

Fixes DX-581


.github/actions/setup-node/action.yaml
```diff
@@ -12,7 +12,8 @@ runs:
       if: ${{ inputs.cache == 'true' }}
 
     - name: Install pnpm
-      uses: pnpm/action-setup@v4 # no need to specify version, it reads from package.json -> packageManager
+      # Pinned to a commit SHA; the pnpm version itself comes from package.json -> packageManager.
+      uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
```
alexjoverm, line 16: wouldn't this prevent security patches from being pulled? (or if gets pinned to a specific malicious commit, will freeze into it forever till manually changed)

what if we just use `corepack` in the `setup-node` below to enable pnpm, so we can get rid of `pnpm/action-setup`?

```
- uses: actions/setup-node@v6
  with:
    node-version-file: ".nvmrc"
- shell: bash
  run: corepack enable pnpm
```

.github/workflows/commitlint.yaml
```diff
@@ -20,6 +20,6 @@ jobs:
     name: Validate PR title
     runs-on: ubuntu-latest
     steps:
-      - uses: amannn/action-semantic-pull-request@v5.5.3
+      - uses: amannn/action-semantic-pull-request@0723387faaf9b38adef4775cd42cfd5155ed6017 # v5.5.3
```
alexjoverm: @maoberlehner what do you think about removing this dependendy? should be fairly easy check to implement manually

packages/cli/src/commands/migrations/migration-filename.ts
```diff
@@ -0,0 +1,24 @@
+import { sanitizeFilename } from "../../utils/filesystem";
```
alexjoverm: @maoberlehner looks like we're changing more logic than necessary (at least in migrations command), including some behavioral changes. I'd try to keep strictly sec issues within the scope of the PR.

Here's claude take:

```
Collisions are a new failure mode. filenamify is many-to-one: hero:v2 and hero_v2 both produce hero_v2.js. migrationTargetsComponent will now match a migration generated for one against blocks of the other. Sanitizing an identifier trades a traversal that requires operator self-harm for a silent wrong-component migration. Low likelihood — Storyblok component names are unlikely to carry : — but that low likelihood cuts against the fix being needed at all.

The dot ambiguity gets enshrined. getMigrationComponentSegment does .split(".")[0], so a component named my.component yields segment my. Pre-existing (the old targetComponent.split(".")[0] did the same), but it's now in a named, documented helper that reads as authoritative. Worth a comment saying dots in component names aren't supported, or handling suffix separation properly.

relative(process.cwd(), ...) changes output beyond the security fix. Default case is covered by the new test, but --path pointing outside cwd now prints ../../... where it used to print a clean ${path}/migrations/.... Minor UX regression riding along.

The double transform is hard to read. The caller does getMigrationComponentSegment(targetComponent), then migrationTargetsComponent internally does getMigrationComponentSegment(buildMigrationFilename(componentName)). Two different normalization paths meeting in the middle. Correct as far as I can tell, but it'd read better as one normalizeComponentKey() applied symmetrically to both sides.

**My take**

I'd ask for this to be split out of the security PR. It's a behavior change to migration matching — with a real collision trade-off — justified by a finding the author already labels as not currently reachable. Bundled with genuine fixes like the prototype pollution and the story-path traversal, it makes the PR harder sky part cover it hasn't earned on its own.
```

packages/js-client/playground/vanilla/src/main.ts
```diff
@@ -30,27 +30,28 @@ const checkTokens = () => {
   return missingTokens;
 };
 
+// Renders JSON as text, never as markup: the payload is an API response and
+// would otherwise be parsed as HTML.
+const renderJson = (value: unknown, codeClass: string) => {
```
alexjoverm: Great fix to prevent XSS, however, it got a bit ugly the code here :D. As it's mostly using the same `pre > code` template, what do you think about simplifying it to something like this?

```
<div id="result" class="...">
  <pre class="...">
    <code id="result-output"></code>
  </pre>
</div>

const render = (value: unknown, isError: boolean) => {
  const out = document.querySelector<HTMLPreElement>("#result-output")!;
  out.textContent = JSON.stringify(value, null, 2);
  out.classList.toggle("bg-red-100", isError);
  out.classList.toggle("text-red-600", isError);
};

const displayResult = (result: any) => render(result, false);
const handleError = (error: any) => render(error, true);
```

.github/actions/setup-node/action.yaml
maoberlehner: As I understand it, the freeze is the point, and the recommendation is to use SHAs. I suspect Aikido would flag your solution again. Also, corepack will not work anymore out of the box when we upgrade to Node >= 25.

maoberlehner: I suspect Aikido scans SHAs and will notify us if one has a known security issue.

alexjoverm: let's trust Aikido then :)

.github/workflows/commitlint.yaml
maoberlehner: Agree, replaced with `commitlint` on PR title.

packages/js-client/playground/vanilla/src/main.ts
maoberlehner: Yes, a lot cleaner, changed it!

packages/cli/src/commands/migrations/migration-filename.ts
maoberlehner: Agree, moved to https://github.com/storyblok/monoblok/pull/811

alexjoverm (approved):

PR #809, opened 2026-09-14
fix(schema): manage block tags by name so a schema pushes to any space

maoberlehner:
`schema init` emitted the source space's raw block tag ids — a block's `internal_tag_ids` and a field's `component_tag_whitelist` / `component_tag_denylist` — and `schema push` forwarded them verbatim. Tag ids are space-local, so pushing that schema into a second space:

- **fails outright** for a block's own tags, which the Management API validates against the target space:
  ```
  internal_tag_ids: [
    'Invalid internal_tag, there is at least one internal_tag not found in our database'
  ]
  ```
- **fails silently** for a field's tag lists, which the API stores without validating — leaving a restriction that matches nothing in the target space.

## Solution

Block tags are referenced by name, the way folders already are (ADR-0010), and resolved to the target space's ids at push time.

```ts
export const heroBlock = defineBlock({
  name: 'hero',
  tags: ['Marketing'],
  fields: [
    defineField('body', { type: 'bloks', allow: [{ tag: 'Marketing' }] }),
  ],
});
```

- `defineBlock` takes `tags: ['<name>']`; `allow` / `deny` take `{ tag: '<name>' }` entries beside the existing block and folder refs, deriving `restrict_type: 'tags'` instead of asking authors to set it by hand. Mixing dimensions in one list throws, as it already did for blocks and folders.
- `schema push` creates a tag the target space does not have yet, then resolves every name to its id, and reports how many tags it created.
- `schema init` emits names, so the round trip between two spaces is byte-identical.
- Diffing runs in name space and sorts both sides, so neither a differing id nor tag order shows up as a change.
- Raw ids stay legal in both places as a same-space escape hatch, and are what a tag the space cannot name falls back to.
- A field restricted to a tag narrows its content type to the blocks declaring it, and `validateStory` enforces the same restriction.

Rationale and alternatives: `adr/0015-block-tag-identity-by-name.md`.

### Deliberate limitations

- Tags are a dependency of the blocks that name them, not a diffed entity: `--dry-run` shows the block's `tags` change but does not announce the tag it would create, and `--delete` never removes a tag (it would affect blocks outside the schema).
- Renaming a tag creates a new one and leaves the old in place, the same trade-off folders and blocks make.
- A `{ tag }` restriction narrows and validates by what the schema declares. Tags applied to a block in the Storyblok UI are invisible to it.

### Out of scope

A block whose local definition omits tags entirely still has its remote tags cleared on push, through `COMPONENT_DEFAULTS`. That is pre-existing and consistent with the other reset-on-omit fields, but worth a separate look.

## Codegen

`Block.tags` and the tag entry in the field restriction types live in `tools/openapi-codegen/templates/`, so every consumer was regenerated in this PR (`@storyblok/api-client`, `@storyblok/management-api-client`, `@storyblok/live-preview`, `@storyblok/schema`).

## Testing

- Unit and integration coverage per layer: `defineBlock` / `defineField` normalization and guards, schema validators, `map-to-wire`, diffing, `schema init` emission, `schema push` resolution and tag creation, and `validateStory`.
- Full suites green: CLI (1333), `@storyblok/schema` (269 + type tests), and every other package. Lint and typecheck clean (the astro SSR playground's `baseUrl` deprecation fails on `main` too).
- Manual QA against two spaces: `schema init` from A → `schema push` into B creates the tag in B, the block and its field restriction carry B's ids, a re-push is a no-op, and `schema init` from B reproduces A's file exactly.

Fixes DX-579


packages/schema/src/helpers/define-field.ts
```diff
@@ -57,27 +61,41 @@ type NormalizeRestriction<T> = T extends readonly any[]
...
   return refs.map((ref) =>
     isFolderRef(ref)
       ? { folder: ref.path }
-      : typeof ref === "string"
-        ? ref
-        : isRecord(ref)
-          ? ref.name
-          : undefined,
+      : isTagRef(ref)
+        ? { tag: ref.tag }
```
dipankarmaikap: Potential issue: unlike defineBlock, this keeps the field tag name verbatim. A value such as `{ tag: " Marketing " }` does not match a block normalized to `"Marketing"`, and whitespace-only tags are accepted. Please trim and reject blank tag references here (with a regression test).

dipankarmaikap (approved):
LGTM

PR #811, opened 2026-09-15
fix(cli): match migration files to components through the same name mapping

maoberlehner:
`migrations run` matched migration files by the raw component name, which broke for any name whose file-name spelling differs from it: `café` stored as NFD on macOS, or a name containing a regex metacharacter such as `+`.

Such a migration could never be applied:

- `migrations run <name>` reported "No migration files found".
- An untargeted run reported success while silently skipping the affected stories.

## Changes

- The component-name-to-file-name mapping now lives in one module that both generate and run use.
- Matching normalizes to NFC, so a file name written as NFD still compares equal to the name the API returns.
- The interpolated `RegExp` is replaced with a string comparison. That removes a crash on component names containing regex metacharacters and a silent mismatch on `+`.
- Generate now sanitizes the component name into the file name, so a name like `../../etc/passwd` can no longer write outside the migrations directory.
- Generate reports the path it actually wrote instead of reconstructing one, and prints an absolute path when `--path` points outside the current directory.

## Keeping the mapping unambiguous

Sanitizing is many-to-one: `hero:v2`, `hero/v2` and `hero?v2` all collapse to `hero_v2`. A dot is ambiguous in the other direction: `my.component.js` is both "component `my.component`" and "component `my`, suffix `component`". Either way, one migration could mutate another component's blocks. A name that is not already an unambiguous file name now carries a short digest of the original:

| Component name | Migration file           |
| -------------- | ------------------------ |
| `hero`         | `hero.js`                |
| `hero:v2`      | `hero_v2-c07e5b.js`      |
| `hero_v2`      | `hero_v2.js`             |
| `my.component` | `my_component-2e0a80.js` |
| `con`          | `con_-1143da.js`         |

A rewritten name never contains a dot, so the first dot in a file name always starts the suffix. Files are matched against the name they would be generated under rather than parsed back into a component name, so `my.component.js` unambiguously means component `my` with suffix `component`, and `my_component-2e0a80.fix.js` unambiguously means component `my.component` with suffix `fix`.

## Compatibility

Before this PR, `migrations generate` wrote the component name verbatim, so on systems that allow the character a file like `hero:v2.js` exists and works today. Those files keep being applied: a component whose name the sanitizer alters also matches a file named after its raw name. `migrations run` names such files once per run so they can be regenerated under the current rules.

One case does change. A component literally named `my.component` with a `my.component.js` on disk now resolves to component `my`, and needs regenerating. That file already applied to `my` before this PR, so no migration changes which blocks it touches — the targeted run just reports it differently.

`--filter` accepts the component name as well as the file name, so `--filter "hero:v2*"` works without knowing the generated file name. A glob is also matched against the rewritten names, and that rewriting is lossy, so such a filter can select other components that are rewritten the same way — pass the component name as the argument to target exactly one.

Fixes DX-584


packages/cli/src/commands/migrations/migration-filename.ts
```diff
@@ -0,0 +1,24 @@
+import { sanitizeFilename } from "../../utils/filesystem";
```
maoberlehner, line 1: Sanitization is many-to-one: `hero:v2` and `hero_v2` both map to `hero_v2.js`. A migration generated for one would therefore be applied to blocks of the other. This PR does not add collision detection. Should `migrations generate` refuse on a collision, or warn and continue?

PR #812, opened 2026-09-16
fix(schema,cli): flag field types the API rejects and fields the push drops

arif-js:
`schema validate` is the offline gate before `schema push`. It resolves `allow` and `deny` references, catches restriction key conflicts, warns on an unknown `restrict_type` and resolves field plugins. It did not check the one thing the Management API is strictest about: the field's `type`.

So this validated clean:

```ts
defineBlock({
  name: 'valid_name',
  is_nestable: true,
  fields: [defineField('broken', { type: 'not_a_real_field_type' as any })],
});
```

```
✔ 0 errors, 0 warnings across 0 of 1 entity
```

And then failed on push:

```
error Failed to create component valid_name
error schema: The schema is invalid, the field: 'broken' has an invalid type: 'not_a_real_field_type'
```

In CI that means validate passes and the deploy breaks a step later, which is the failure mode validating offline exists to prevent.

## Scope

`defineField` rejects an unknown `type` at compile time, so TypeScript authors are already covered. This matters for schemas authored in plain JavaScript, assembled at runtime, or read back from a space. That is the same audience the existing `allow`, `deny` and `restrict_type` checks serve, so the validator was inconsistent rather than wrong.

Two codes, because the API reports two different failures:

- `missing_field_type` — `type` absent, `null`, or blank. The API rejects these as `type can't be blank`.
- `unknown_field_type` — any other value the API does not accept, including non-strings. `type: 123` fails the push with `has an invalid type: '123'`.

Both are errors rather than warnings. The `restrict_type` check next door is a warning because the API never validates that key, so a space can legitimately hold a value nothing recognises. `type` is the opposite: the API rejects an unrecognised one outright, so there is no stored value this could wrongly fail.

## Implementation

`FieldType` is `Field["type"]`, a type with no value form, so there was nothing to check against at runtime. This adds `FIELD_TYPES` in `src/field-types.ts` as its runtime counterpart, `satisfies readonly FieldType[]`, exported from the package index alongside the sibling constants in `restrictions.ts`.

Drift is the obvious risk in a hand-written list next to generated types, so it is guarded from both sides: `satisfies` rejects an entry that is not a real field type, and `field-types.test-d.ts` fails `test:types` if the generated union gains a member the list is missing. The type test uses `Exclude` plus `toBeNever` rather than `toEqualTypeOf` so the failure names the culprit (`ExpectNever<"commerce">`) instead of printing two truncated unions.

Codegen could emit this list from the OpenAPI spec and delete the whole class of problem, but the templates are static TS with no interpolation, so it means generator work plus regenerating every consumer. Better as its own PR.

## Verification

The list was checked against the API rather than assumed. A single component carrying all 22 types is accepted. Every near-miss probed — `blocks`, `date`, `plugin`, `seo`, `Text`, `TEXT`, `"text "`, `rich_text`, `number_range`, `asset_link`, `reference`, `multi_asset`, `html`, `json`, `color`, `uuid`, `""`, `123`, `true`, `null`, and an absent `type` — comes back 422. The validator now flags exactly that set and nothing else.

Round-tripped a real space through `schema init` and re-validated: `0 errors, 0 warnings across 0 of 5 entities`, so the check does not fire on schemas the editor writes.

## Tests

Five cases in `validate-schema.test.ts`: an unknown string type, a non-string type, an absent and a null type, a blank type, and the message for a nameless field. Each was confirmed to fail without the source change.

`packages/schema`: 290 tests pass, `Type Errors no errors`, lint clean. `storyblok`: 1416 pass.

## The silent drop on push

The same gap has a second half that `schema validate` alone does not close, because `schema push` never runs the validator. A field with no `name` was dropped by the wire mapping outright; a field with no `type` reached the Management API as an empty object, which the API discards on the way in. Both pushed green and left the component missing the field, with nothing reporting it:

```ts
export const hero = {
  name: 'qa_hero',
  fields: [
    { name: 'headline', type: 'text' },
    { name: 'orphan' },
  ],
};
```

```
✔ Schema pushed to space <id>. 1 created, 0 updated.
```

Reading the component back returns a schema holding `headline` only.

The load now refuses those fields before any API call, naming each one:

```
▲ error Schema fields that cannot be pushed: field "orphan" in block "qa_hero" has no "type". Run `storyblok schema validate` for the full report.
```

This lives next to the existing duplicate-identity check in `load-schema.ts`, which fails the same way for the same reason, and covers `schema diff` as well since both share the load. A clean schema still pushes and reads back with every field intact.

Six cases in `load-schema.test.ts`: a nameless field, an absent type, a blank type, a non-object entry, several problems reported in one error, and the pointer to `schema validate`.


maoberlehner:
Thank you for your contribution @arif-js ! Rebased onto latest `main` and added another fix for an adjacent problem.

maoberlehner (approved):

arif-js:
> Thank you for your contribution @arif-js ! Rebased onto latest `main` and added another fix for an adjacent problem.

@maoberlehner Thanks for getting this merged! I appreciate the extra polish with the rebase and the additional fix. Cheers!

PR #814, opened 2026-09-17
chore(astro,nuxt): make the e2e runs reliable locally and in CI

maoberlehner:
Fixes two unrelated e2e failures. The Astro suite has been failing locally
while passing in CI; the Nuxt run intermittently fails `playground:build` in CI
while passing locally.

The Astro failure is not a flake in the usual sense: there are two separate
defects, and the first one hid the second.

## The dev server was never actually running

`astro dev` daemonizes when it detects an agentic environment — it forks the
server, prints `Dev server running at http://localhost:4321 (pid N)`, and the
foreground process exits 0. `start-server-and-test` treats its child exiting as
the server dying and aborts with `server closed unexpectedly` before Cypress
ever starts.

That is why the failure looked environmental: CI runs no agent, so there the
server stays in the foreground and the suite passes. Anyone driving the repo
from a coding agent gets a deterministic failure instead, which is how this got
recorded as a pre-existing failure on `main` in #807.

`ASTRO_DEV_BACKGROUND` suppresses the auto-detection, so the e2e target now
starts the playground through a `playground:test:foreground` script. The name
reads backwards on purpose: any non-empty value tells Astro the background
decision has already been made, so it skips detection and stays in the
foreground. `docs/testing-patterns.md` records this, because the next person to
hit it will be looking at `server closed unexpectedly` with no obvious link to
the dev server forking.

The plain `playground:test` script is untouched, so running the playground by
hand from an agent still gets the background server and `astro dev logs`.

## The dev toolbar loses a race against dependency optimization

With the server actually up, Cypress ran and the first spec failed:

```
Failed to fetch dynamically imported module:
  http://localhost:4321/node_modules/.vite/deps/toolbar-<hash>.js?v=<hash>
```

The dev toolbar is pulled in through a dynamic import. When Vite re-optimizes
dependencies, the module that was already served still points at the previous
optimized file name, so the import rejects and Cypress fails the test on the
unhandled rejection. Clearing the dependency cache between runs does not settle
it — three consecutive clean-cache runs went pass, pass, fail — because it is a
race, not a stale-cache artifact.

In a browser this is invisible: Vite reloads the page. It is only fatal under
Cypress, which fails a test on any unhandled rejection from the application. So
this is a harness artifact rather than a product defect, and the fix is to take
the toolbar out of a suite that never tested it. The test playground gates it on
`STORYBLOK_E2E`, which only the e2e script sets, so running the playground by
hand still gets the toolbar. No spec covers it before or after this change.

## Nuxt: the e2e run deletes the playground's `.nuxt`

The playground e2e suite runs `setup()` with `dev: true`, so `@nuxt/test-utils`
skips the private build directory it uses for non-dev fixtures and works in the
playground's own `.nuxt`. When that directory does not exist yet (every fresh
checkout) it creates it and registers a teardown that removes the whole thing.

nx builds the playground in parallel with the tests, and a `nuxt build` whose
`.nuxt` disappears mid-transform fails every `.vue` file with `Failed to load
tsconfig '.nuxt/tsconfig.json': Tsconfig not found`, because the playground
`tsconfig.json` extends it. It never reproduces locally because `.nuxt` is
already there.

The suite now creates the directory before `setup()`, which leaves it outside
the teardown's ownership. Pointing the fixture at a directory of its own does
not work: the dev server then has no `.nuxt/tsconfig.json` to extend.

## Verification

Five consecutive `test:e2e` runs pass with all five specs, and leave no dev
server behind on ports 4321-4325. `pnpm nx test @storyblok/astro` passes twice
in a row with the cache skipped. Before this, the Astro suite never got past
starting the server.


dipankarmaikap (approved):
LGTM

Issue #802, opened 2026-09-07
Make `rateLimit` in clients more powerful to allow cross instance limits in CI or workers


angelikatyborska:
## Package

storyblok-js-client / @storyblok/api-client / @storyblok/management-api-client
(Fix in storyblok-js-client only if it is perfectly backward-compatible and easily doable!)

## Context

Because our API rate limits are per token and enforced in the clients per instance, the clients can still run into rate limit errors when multiple instances are active in parallel using the same token, as is often the case in CI, worker, or load-balanced environments.

### Possible solution

```typescript
```

Also requested by: [https://storyblok.atlassian.net/jira/polaris/projects/SBD/ideas/view/d5e85a64-6cd4-4bc4-8926-ffbf8af40d1a?selectedIssue=SBD-1540](<https://storyblok.atlassian.net/jira/polaris/projects/SBD/ideas/view/d5e85a64-6cd4-4bc4-8926-ffbf8af40d1a?selectedIssue=SBD-1540>)

## Context provided by the website team

When we deploy our Storyblok website, we run two builds in parallel. We do A/B testing using SSG, which requires two deployments of the website to be deployed at the same time, variant A and variant B. With the current rate-limiting logic in the JS client, we often run into 429 errors that sometimes cannot be recovered from because of the two parallel builds doing identical requests at the same time.

We would love to be able to leverage the current JS client's rate limiting logic that automatically decides the rate limit depending on the request type, but we would like to be able to divide that limit by the number of parallel builds that we run.

Additionally, it would be great if the retry delay was exponential, increasing the chances that the JS client will be able to recover from 429 errors at some point.

### Proposed Solution

1. Allow for the `retriesDelay` configuration option to be a function (or introduce a new configuration option). Pass the number of the retry as the function argument. This would allow users to pass for example: `retriesDelay: (n) => 300 * Math.pow(2, n)`.
2. Allow for the `rateLimit` configuration option to be a function (or introduce a new configuration option). Pass the rate limit that would apply to the request if the user did not provide a configuration option for it as the function argument (based on the request type, last response headers etc.). This would allow users to pass for example `rateLimit: (n) => Math.floor(n/2)` where `2` is the number of builds that they run in parallel.

### Alternatives Considered

I considered using the current configuration options to set `retriesDelay` to a bigger number than the default 300, but that would have negative performance implications on my build process, and I cannot really afford that.

[...]

PR #818, opened 2026-09-21
feat(capi-client,mapi-client): adaptive rate limiting and a pluggable limiter

maoberlehner:
Rate limits are enforced per token; the clients enforce them per instance. When several instances run against one token (two builds of an A/B-tested site, a CI matrix, a worker fleet), each paces itself correctly and the fleet still overruns the quota between them, producing 429s the retry budget cannot always absorb.

This adds both halves of the fix: a default that reacts, and an interface for consumers who need to be proactively correct.

## Adaptive by default

The default limiter backs off multiplicatively when a request is throttled and recovers additively while none is. It is bounded on both sides, never above the rate the client would have used anyway and never below one request per second, so a single instance is unaffected and a fleet degrades instead of failing. `rateLimit.adaptive: false` pins the old behaviour; an object tunes it.

Three guards the interface needs: `minRequestsPerSecond` is clamped to 1, because zero reads as "no limit" to the underlying window; an advertised quota replaces the previous one rather than ratcheting down, so a transient low reading cannot pin the client for the life of the process; and a failing limiter hook cannot turn a served request into an error, which inside the retry loop would mean sending it again.

## Pluggable limiter

`rateLimit.limiter` replaces the in-memory default with any implementation of `acquire` / `recordResponse` / `release`. The context names the quota the request draws from and the rate the client would have used, so a fleet can divide that rate or spend it from one shared Redis bucket. `createDefaultRateLimiter` is exported so a custom limiter can wrap it rather than reimplement it.

Admission sits at the HTTP request boundary, not around the call: retries are issued inside a single call, so gating the call would let one admitted request put `retry.limit + 1` requests on the wire, up to 13 for the Management API, exactly during the storm the limiter exists to damp. The same boundary is why the limiter sees the 429s a retry replaced, which are most of the signal.

## Measured

A simulation of instances sharing one token, with the real limiter under test on a virtual clock. `success` is the share of issued requests that eventually returned data, and `goodput` counts only those: the baseline is not faster, it abandons more work.

| arm | instances | 429s | dropped | success | makespan | goodput req/s |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| baseline | 1 | 0 | 0 | 100.0% | 7.29s | 54.9 |
| aimd | 1 | 0 | 0 | 100.0% | 7.29s | 54.9 |
| baseline | 4 | 2103 | 289 | 81.9% | 26.82s | 48.9 |
| baseline-divided | 4 | 0 | 0 | 100.0% | 33.07s | 48.4 |
| aimd | 4 | 151 | 0 | 100.0% | 35.82s | 44.7 |
| shared-store | 4 | 0 | 0 | 100.0% | 31.12s | 51.4 |
| baseline | 8 | 7385 | 1344 | 58.0% | 37.80s | 49.1 |
| baseline-divided | 8 | 0 | 0 | 100.0% | 66.07s | 48.4 |
| aimd | 8 | 643 | 3 | 99.9% | 73.57s | 43.5 |
| shared-store | 8 | 0 | 0 | 100.0% | 63.07s | 50.7 |
| baseline | 16 | 19096 | 4027 | 37.1% | 48.13s | 49.3 |
| baseline-divided | 16 | 0 | 0 | 100.0% | 133.07s | 48.1 |
| aimd | 16 | 1917 | 77 | 98.8% | 135.83s | 46.5 |
| shared-store | 16 | 0 | 0 | 100.0% | 127.07s | 50.4 |

At 4 instances adaptation drops nothing where the baseline abandons 18% of the work; at 16 it takes success from 37% to 99%. A shared limiter is never throttled at any count. Dividing the rate by hand, the documented workaround, turns a 15s mixed-tier build into a 299s one at 4 instances. ADR 0016 has the full matrix, both workloads, and where adaptation falls short.

## Tuning

A sweep of `decreaseFactor` x `increaseStep` x `recoveryIntervalMs` over both workloads at 4 and 16 instances, worst case across the four runs:

| tuning | worst success | dropped | total time |
| --- | ---: | ---: | ---: |
| d=0.5, +1/1000ms | 96.2% | 249 | 277.6s |
| d=0.5, +2/1000ms | 96.0% | 269 | 265.0s |
| d=0.7, +1/1000ms | 92.6% | 597 | 258.8s |
| d=0.8, +2/500ms | 82.1% | 1637 | 235.1s |
| d=0.8, +10/500ms | 64.1% | 4383 | 181.9s |

Halving on a one-second cooldown is the reliability-optimal corner: every gentler back-off and every faster recovery buys wall-clock for a worse worst case. Those defaults stay.

Recovery is the one the sweep argued against leaving fixed, because it only measured fleets. A step of 1 req/s undoes a halving of the 50/s tier in 25 quiet seconds and of the 6/s tier in one, so the tier a client uses decided what a stray 429 cost it. On one uncontended client meeting a single 429 in a 600-request run:

| recovery | clean | one stray 429 | cost |
| --- | ---: | ---: | ---: |
| adaptive off | 11.2s | 12.0s | +0.8s |
| fixed +1/1000ms | 11.2s | 17.2s | +6.0s |
| scaled to bucket (default) | 11.2s | 15.1s | +3.9s |

Recovery therefore defaults to a twenty-fifth of the bucket's rate, at least 1 req/s: 2/s on the 50/s tier, 1/s on the small ones. Fleet success rates move by less than 0.1 point and the fleet runs finish 4-12% sooner. It does widen the spread between instances at 8 and 16, which is the fairness cost of the larger step.

## Backward compatibility

`rateLimit: false | number | { requestsPerSecond } | { maxConcurrency }` all behave as before. New behaviour is additive, except that adaptation is on by default — it can only lower a client's rate below what it already uses.

## Follow-ups

- Reference docs on the docs site for `adaptive` and `limiter`, with a worked Redis example.
- No way to observe that the client backed off; a callback in the shape of `cache.onRevalidationError` would fit.

Fixes DX-570


angelikatyborska:
Hi Marcus. Thanks for requesting my review, but I don't think I can help out here. I don't have a project that uses the new API clients at the moment so I cannot try out this PR. It would be a bigger task for me to migrate to the new clients. It will happen, but not today 🙂 

Do I understand correctly that this PR adds the possibility of passing `decreaseFactor: 0.5` to the Client config to achieve what I asked for in this issue https://github.com/storyblok/monoblok/issues/802 ? If yes, that sounds perfect to me.

maoberlehner:
> Hi Marcus. Thanks for requesting my review, but I don't think I can help out here. I don't have a project that uses the new API clients at the moment so I cannot try out this PR. It would be a bigger task for me to migrate to the new clients. It will happen, but not today 🙂
> 
> Do I understand correctly that this PR adds the possibility of passing `decreaseFactor: 0.5` to the Client config to achieve what I asked for in this issue #802 ? If yes, that sounds perfect to me.

@angelikatyborska, Not quite. `decreaseFactor` is only a tuning config, and `0.5` is already its default. For your use case, no extra config would be needed at all, because the AIMD (additive increase/multiplicative decrease) algo should handle it just fine.

Yet, for the proactive version your original request described (divide the rate by the number of parallel builds), there's `rateLimit.limiter`: a small interface (`acquire` / `recordResponse` / `release`) that replaces the built-in limiter.

```ts
import { createDefaultRateLimiter, storyblokClient } from "@storyblok/api-client";

const PARALLEL_BUILDS = 2;
const shared = createDefaultRateLimiter();

const client = storyblokClient({
  token: "...",
  rateLimit: {
    limiter: {
      acquire: (context) =>
        shared.acquire({ ...context, limit: Math.max(1, Math.floor(context.limit / PARALLEL_BUILDS)) }),
      recordResponse: shared.recordResponse,
      release: shared.release,
    },
  },
});
```

But as I said, for your use case you'd not need to provide a custom limiter at all!

packages/mapi-client/src/utils/rate-limit.ts
```diff
@@ -30,16 +85,57 @@ export interface ThrottleManager {
...
   return {
-    execute: (fn) => throttle.execute(fn),
+    execute: (fn) => fn(),
+    wrapFetch: (fetchFn) => async (input, init) => {
+      // Admission sits here rather than around the call because the HTTP layer
+      // retries inside one call: with the default `retry.limit` of 12, gating
+      // the call alone would let one admitted request put 13 on the wire.
+      const context: RateLimitContext = {
+        path: pathOf(getRequestUrl(input)),
+        query: {},
```
dipankarmaikap: The MAPI client correctly sends pagination parameters such as `page` and `per_page` on the wire, but the custom limiter currently receives `query: {}`. This means custom limiters cannot observe or correlate requests by pagination parameters, even though `RateLimitContext` documents that query values are available.

maoberlehner: Good catch! Fixed.

dipankarmaikap (approved):
LGTM

Issue #20, opened 2024-10-11
Custom Component Resolver (Vue)


babalugats76:
###  Summary ###

I am using Nuxt 3/Vue and need help migrating from renderRichText to storyblok/richtext, especially when it comes to the **resolution and rendering of custom components** contained within richtext fields.

I realize that this project is relatively new and probably still under heavy development, but, I have reviewed the playgrounds and available materials closely and my use case does not appear to be adequately addressed.  IMHO, the Vue **playground is missing key design patterns** that would seemingly be part of any Vue implementation of the module.

## Previous Solution ###

My previous solution utilized `renderRichText` and the `Vue3RuntimeTemplate` modules (for dynamic resolution).

```javascript
<template>
  <section
    v-editable="blok"
    :class="classes"
  >
    <Vue3RuntimeTemplate :template="resolvedRichText" />
  </section>
</template>

<script setup>
import Vue3RuntimeTemplate from 'vue3-runtime-template';
import { RichTextSchema } from '@storyblok/js';
import cloneDeep from 'clone-deep';

const props = defineProps({ blok: Object });
const mySchema = cloneDeep(RichTextSchema);

const resolvedRichText = computed(() =>
  renderRichText(props.blok.text, {
    schema: mySchema,
    resolver: (component, blok) => `<component is="${component}" :blok='${JSON.stringify(blok)}' />`,
  }),
);

const getProseVariant = (variant) => {
  const variants = {
    gray: 'prose-gray',
    lavender: 'prose-lavender',
    neutral: 'prose-neutral',
    primary: 'prose-primary',
    slate: 'prose-slate',
    stone: 'prose-stone',
    zinc: 'prose-zinc',
    card: 'prose-card',
  };
  return variants[variant] || '';
};

const classes = computed(() => {
  const baseClasses = 'rich no-underline prose-a:no-underline prose dark:prose-invert max-w-none';

  const variantClass = getProseVariant(props.blok?.variant);
  const lgClass = variantClass !== 'prose-card' ? 'lg:prose-lg prose-code:lg:text-xl' : '';
  return `${baseClasses} ${variantClass} ${lgClass}`;
});
</script>

<style>
.rich > div {
  @apply mt-[2em] mb-[2em];
}
</style>
```

## Work-In-Progress ###

[...]

PR #819, opened 2026-09-21
feat(capi-client,mapi-client): discover the rate-limit ceiling from cache hits

maoberlehner:
Stacked on #818 — review that first. Base is `feat/DX-570-adaptive-pluggable-rate-limiting`.

## What this changes

ADR-0016's ceiling is the tier limit, so AIMD can only ratchet down. The Content Delivery API's
per-second tiers (50/s single story or ≤25 items, 15/s for 26–50, 10/s for 51–75, 6/s for 76–100)
apply to requests that reach the origin; a request the CDN answers from its cache never gets there.
A client that re-reads warm URLs was therefore paced by a limit that did not apply to most of its
requests.

Each tier now keeps a rolling window of the last 50 responses whose cache status it could read, and
its ceiling becomes `tierLimit / (1 - hitShare)`, bounded by eight times the tier. Only the requests
that miss the cache reach the origin, so a tier running at that rate still puts at most `tierLimit`
requests per second on the origin. AIMD is otherwise untouched: the rate climbs a step per quiet
second and still halves on a 429, so a raised ceiling is reached over tens of seconds, never in a
burst.

**This reverses ADR-0016's claim that adaptation can only make a client more conservative.** That is
deliberate and is why the manual tests below came before the code. ADR-0017 records the amendment.

## What it buys, measured

A warm re-read workload through the built client, against the live API. The two arms differ only by
`rateLimit: { cacheAware: false }` — the control is the base branch with the feature off.

**25-item listings, the 50/s tier, 60s per arm:**

| | control (`cacheAware: false`) | this PR | |
| --- | ---: | ---: | ---: |
| **stories delivered** | **75,000** | **491,550** | **6.6x** |
| **stories per second** | **1,250** | **8,193** | **6.6x** |
| requests per second | 50.0 | 327.7 | 6.6x |
| peak admitted rate | 50 (pinned) | **400 r/s** | 8x |
| cache hits / misses | 3,000 / 0 | 19,662 / 0 | |
| 429s | 0 | 0 | |

**100-item listings, the 6/s tier, 75s per arm:**

| | control (`cacheAware: false`) | this PR | |
| --- | ---: | ---: | ---: |
| **stories delivered** | **45,000** | **300,000** | **6.7x** |
| **stories per second** | **600** | **4,000** | **6.7x** |
| requests per second | 6.0 | 40.0 | 6.7x |
| peak admitted rate | 6 (pinned) | **48 r/s** | 8x |
| cache hits / misses | 450 / 0 | 3,000 / 0 | |
| 429s | 0 | 0 | |

### How this was measured

The workload calls the public API and the limiter does all the pacing. A QA space seeded with 138
published stories. Client-side cache off and retries off, so every call reaches the wire and a 429 is
counted rather than absorbed. `cv` pinned, so no read redirects through the origin. Each arm warms
its whole working set twice, sleeps, then times; only responses completing inside the measured
window count. Control first, 20s cool-down, whole sequence run twice — spread under 1%. Cache
status and 429s read off the real responses.

### What the numbers do and do not show

Both arms ran entirely on cache hits, so the control's 6 r/s and 50 r/s were pacing against an
origin they never touched. That is the point, and it is also the limit of the claim:

- **The averages are 6.6x, the bound is 8x.** Both arms reached their full ceiling (400 r/s and
  48 r/s); the window fill and the AIMD climb sit inside the measured window and pull the average to
  about 82% of the plateau. A run shorter than half a minute sees much less.
- **A cold first pass gains nothing.** Distinct URLs read once are all misses and the ceiling never
  leaves the tier limit.
- **A browser gains nothing**, since it cannot read the cache status (#7).
- **This says nothing about the cold transition.** The workload never went cold, so the zero 429s
  here are not a claim that the change costs none — that cost is measured separately below.

## Findings

All against the live API over the network. Cache statuses are quoted verbatim.

| # | Scenario | Request | Relevant response headers | What it establishes |
|---|---|---|---|---|
| 1 | Cold published read | `GET /v2/cdn/stories/home?cv=<current>&version=published`, first hit on that URL | `200`; `x-cache: Miss from cloudfront`; no `age`; `cache-control: max-age=0, public, s-maxage=604800, stale-if-error=3600` | The cache status is present on the **cold** request too, so its absence is never ambiguous with a miss |
| 2 | Same URL, 2nd and 3rd read | identical to #1 | `200`; `x-cache: Hit from cloudfront`; `age: 1` then `age: 2` | A hit is distinguishable from a miss by a single header |
| 3 | Published read without `cv` | `GET /v2/cdn/stories/home?version=published` | `301` to the same URL with `cv=<current>`; `cache-control: no-cache`; `x-cache: Miss from cloudfront` | An unpinned or stale `cv` is a redirect that always reaches the origin — it is correctly counted as a miss |
| 4 | Stale `cv` | `GET …&cv=1&version=published` | `301`, same as #3 | Same |
| 5 | Draft read, warm URL | `GET /v2/cdn/stories?version=draft&per_page=100`, repeatedly | `200`; `cache-control: max-age=0, private, must-revalidate`; `x-cache: Hit from cloudfront` on the large majority, `RefreshHit` on the rest | **Draft traffic is cached like published traffic.** The `cache-control` header describes what a client may store, not what the CDN serves. `RefreshHit` is a revalidation against the origin, so only a plain hit is treated as free — see #20 for what that is worth against the tier |
| 6 | Rejected request | `GET …&per_page=9999` | `422`; `x-cache: Error from cloudfront` | A rejected request is never reported as a hit. An unknown query parameter is not a reliable way to force one: `cv` pinning redirects (`301`) before validation runs. Forced misses in the probes below vary `cv` instead |
| 7 | **CORS exposure** | `GET` with `Origin: https://example.com`, published and draft, plus the `OPTIONS` preflight | `access-control-expose-headers: Api-Version, Token, Total, Per-Page` | **`X-Cache` and `Age` are not exposed to cross-origin script.** A browser client cannot read the cache status at all. This is the finding that shaped the design |
| 8 | Rate-limit headers, draft | `GET …&version=draft` | `x-ratelimit: "space-concurrent-requests";r=29`; `x-ratelimit-policy: "space-concurrent-requests";q=30` | Confirms the prior finding: sent only for draft, and the quota describes simultaneous requests, not a rate. Already ignored by the base branch |
| 9 | Rate-limit headers, published | `GET …&version=published` | neither header present | Same |
| 10 | Tier probe, 76–100 band at 10/s | 40 forced-miss `per_page=100` listings paced at 10/s | 23 × `200 Miss`, **17 × `429`** | A 429 arrives where `TIER_LIMITS` predicts for the 6/s tier |
| 11 | Tier probe, 76–100 band at 5/s | 20 forced-miss `per_page=100` listings paced at 5/s | 19 × `200 Miss`, 1 × `429` | Just under the 6/s tier is essentially clean — the tier boundary is where it is documented to be |
| 12 | **429 headers** | the 429s from #10 | `x-cache: Error from cloudfront`; **no** `retry-after`, **no** `x-ratelimit`, **no** `x-ratelimit-policy` | There is nothing to read off a 429 but its status, so back-off has to be blind. A 429 is also never reported as a hit |
| 13 | Tier probe, ≤25 band at 80/s | 320 forced-miss `per_page=1` listings paced at 80/s | 274 × `200 Miss`, **17 × `429`** | Exceeding the 50/s tier with origin-bound traffic is throttled |
| 14 | **Cached traffic at the same 80/s** | 320 reads of one warm `per_page=1` published URL, paced at 80/s | 221 × `200 Hit`, **0 × `429`** | Directly against #13: the same rate is throttled when it reaches the origin and not throttled when the cache serves it. This is the headroom the change goes after |
| 15 | Forced-miss burst, 26–50 band | 60 forced-miss `per_page=25` listings, bursted | 60 × `200 Miss`, 0 × `429` | The ≤25 tier tolerates a burst of 60, consistent with 50/s |
| 16 | Management API, twice | `GET /v1/spaces/<id>/stories?per_page=1` | `200`; `cache-control: max-age=0, private, must-revalidate`; `x-cache: Miss from cloudfront` both times | The Management API never serves a hit, so the mechanism would be inert there even if it were wired up. It is not wired up, which makes that structural |

### What a publish does to a warm client

| # | Scenario | Observation | What it establishes |
|---|---|---|---|
| 17 | **Does a publish purge?** | warm a version-pinned published URL, publish a story mid-watch, keep reading the same URL for 6s: `x-cache: Hit from cloudfront` throughout, `age` rising 3→11, body unchanged | **A publish purges nothing.** The pinned URL keeps serving hits. `s-maxage` is 604800, so that entry stands for days |
| 18 | What a publish does instead | same publish, then one unpinned read: the content version moves (e.g. `…569` → `…875`) | A publish mints a **new** content version. A client that follows it lands on keys nothing has warmed — a key rotation, not a purge. That is the real cold transition, and it is abrupt: every URL goes cold together |
| 19 | Is the cold transition observable before a 429? | the first requests after following the new version: `200` with `x-cache: Miss from cloudfront` | Yes — the miss status arrives on the very first affected response, while a 429 only appears once the rate has already exceeded the tier. The misses lead the 429s, which is what the cold-run clamp keys on |

### Does a cache hit actually spare the origin? (#20)

The premise the whole design rests on is not that the API reports a hit, but that a hit does not
draw on the tier. Only a rate above the tier can show that, so it was measured directly: short
bounded runs against one warm 100-item listing on the 6/s tier, with a forced-miss run as the
positive control that the probe can see throttling at all.

| traffic | rate attempted | outcome |
| --- | ---: | --- |
| forced misses (control) | 16.2/s | **29 of 30 throttled** |
| published, warm | 16.6/s | 60 × `Hit`, **0 × `429`** |
| draft, warm | 18.1/s | 57 × `Hit`, 3 × `RefreshHit`, **0 × `429`** |
| draft, warm | 33.0/s | 58 × `Hit`, 2 × `RefreshHit`, **0 × `429`** |
| published, warm | 32.1/s | 60 × `Hit`, **0 × `429`** |

Draft hits cost the tier nothing, at five times the tier as readily as at three, exactly like published
ones. An earlier reading that draft traffic never produces a plain hit was wrong, and ADR-0017 has
been corrected. The mechanism needs no notion of content version: it reads the cache status, and the
cache status is right for both.

`Age` is not a usable substitute for `X-Cache` in either direction: about a quarter of genuine hits
carry no `Age` at all, and plain hits regularly carry `age: 0`.

## The cold transition, measured

Sustained traffic against one warm URL until the ceiling had visibly climbed, then a real publish
mid-run, then the traffic following the new version onto cold keys — all live, on the 6/s tier, with
the ceiling scaled down so the cold phase could not hammer the origin. The ratio to the tier is what
governs the dynamics, so it is held at the shipped worst case.

| ceiling | warm plateau (all hits, 0 × 429) | 429s in the transition | back inside the tier after |
| ------: | -------------------------------: | ---------------------: | -------------------------: |
| 1x (feature off, control) | 6.0 r/s | **0** | — |
| 4x | 24.2 r/s | 8 | 4.5s |
| 10x | 59.0 r/s | 12 | 2.4s |
| 20x | 118.6 / 120.4 r/s | 27, 24 | 1.7s, 1.6s |

The burst is **not** the window being slow to follow. It is the requests already in flight when the
first miss comes back — roughly the ceiling times one round trip, which no feedback can react its
way out of. I built the obvious reactive fix first (clamp on a run of consecutive misses instead of
waiting for the average) and measured it moving the burst from 27 to 24, inside the run-to-run
spread. The clamp does work — a bucket at 120/s drops to its 6/s limit on a run of misses, where the
average alone would still allow 25/s — it just cannot touch what was already on the wire.

So the ceiling is bounded at **eight times the tier**. The multiple is a judgement on a measured
curve, not a derived constant: the gain flattens while the transition cost keeps rising roughly
linearly. Eight keeps most of the benefit (400/s on the 50/s tier, 48/s on the 6/s one) at a bounded
worst case. Summed across the four tiers a client could in principle drive 648/s, which is the
global worst case.

**Shipped configuration, measured live on the 6/s tier:**

| | |
| --- | --- |
| warm plateau | **47.8 r/s**, 239/239 cache hits, **0 × 429** |
| requests issued between the new version appearing and the rate settling | **40** |
| of those, throttled | **7** |
| overshoot duration | **1.09s** |
| admissions per second across the transition | 48 · 48 · 48 → **34** → **6** · 6 · 6 |

Against the control (feature off), the transition costs 7 extra 429s and is over in about a second.

## What the headroom costs to reach

The window has to fill before the ceiling moves at all (about 9s at the 6/s tier) and AIMD then
takes roughly another 20s to climb to 8x, so a run shorter than half a minute sees little of it. A
cold first pass sees none: 15 distinct published URLs read once returned 15 × `Miss` and 0 × `Hit`.
The gain belongs to workloads that keep going, or that come back to keys something has warmed —
repeated builds, previews, long-running renderers.

## Robustness fixes in this PR

- The cached rate reached the per-second window unguarded. Zero and `NaN` read as "no limit" and
  admitted everything at once; a value below the tier paced it slower than the quota it asked for.
  Every rate the limiter derives from arithmetic now passes through one guard, replacing two
  existing ad-hoc floors.
- The cache status is matched as whole words, so a value joined from more than one cache hop reads
  the same whichever hop is listed first.
- The unparsed-URL fallback in the limiter context passed the raw URL, query included, to a custom
  `RateLimiter`. It now strips the query, so no path can hand a token to third-party limiter code.
  Not reachable through the public client today.
- `CacheAwareConfig` was exported with nothing that accepts one. The Content API client now also
  exports `DefaultRateLimiterOptions` and the stock cache-status detector; the Management API client,
  which has neither, no longer exports the type.

## Follow-ups, not in this PR

- **The docs site's `RateLimitConfig` page is stale and this change makes it misleading.** It still
  describes `rateLimit` as a concurrency cap, lists only `maxConcurrency` and `adaptToServerHeaders`,
  and says the client uses `X-RateLimit-Policy` as a concurrency cap — which the base branch
  deliberately stopped doing. It needs the whole current surface: `requestsPerSecond`, `adaptive`,
  `cacheAware`, `onRateLimitChange`, `limiter`, and the corrected framing.
- **`cache: false` is silently ignored.** `CacheConfig` has no `false` form and the client
  destructures `cache = {}`, so caching stays on: 60 calls produced 1 HTTP request. `ttlMs: 0` does
  not expire entries either. Both are pre-existing on the base branch.


adr/0016-adaptive-and-pluggable-rate-limiting.md
```diff
@@ -1,6 +1,6 @@
 # ADR-0016: Adaptive and Pluggable Rate Limiting in the API Clients
```
maoberlehner: TODO Add DOCS.md with updates for official docs

packages/capi-client/src/utils/limiter.ts
```diff
@@ -191,11 +447,46 @@ export function createDefaultRateLimiter(options: DefaultRateLimiterOptions = {}
...
+      configuredLimit: bucket.configuredLimit,
+      cacheHitShare: hitShareOf(bucket),
+    };
+    const previous = bucket.reportedStatus;
+    if (
+      previous &&
+      previous.requestsPerSecond === status.requestsPerSecond &&
+      previous.ceiling === status.ceiling &&
+      previous.configuredLimit === status.configuredLimit &&
+      previous.cacheHitShare === status.cacheHitShare
```
dipankarmaikap: **claude:** nitpick: `reportStatus`'s dedup check manually compares four named fields of `RateLimitStatus` (`requestsPerSecond`, `ceiling`, `configuredLimit`, `cacheHitShare`) instead of iterating the object. If a future field is added to `RateLimitStatus`, nothing forces this comparison to include it — it would silently fail to dedup on that new field. Low risk today since the interface is stable, but worth a comment noting the fields must be kept in sync, or a more structural equality check if the interface grows.

*This comment was left by an AI agent (Claude) reviewing this PR.*

dipankarmaikap:
**claude:** QA finding — could not independently reproduce ADR-0017's cold-transition burst claim.

I ran real-world traffic (not mocks) against a live QA space to validate this: a real published 106-story listing warmed to the 6/s tier's 8x ceiling (climbed 6→48 r/s over ~29s, 0 429s, matching ADR-0017's own numbers closely). Then I republished a story to rotate the content version and kept reading through the transition.

Confirmed the mechanism it's built on is real: a raw `curl` before/after the republish shows `cv` rotating and the next fresh read on the new `cv` genuinely returning `Miss from cloudfront`. But every time I drove *concurrent* client traffic (dozens of in-flight requests) through that same transition, CloudFront's own request-coalescing absorbed it — the burst came back 100% `Hit`, never `Miss`, across several repeated attempts (including with the artificial client-side pacing gap removed so admission was governed purely by the limiter itself). I could not get a visible burst of misses or 429s out of a republish this way, so I couldn't confirm the ADR's specific "~7 throttled over ~1.1s" measurement independently.

Separately, I did confirm the underlying back-off mechanism itself fires correctly on genuine live 429s: a forced-miss deluge (unique nonce per request, so nothing could ever be cached) produced 46 real 429 responses from the live API — invisible to a naive status check because `ky`'s built-in retry absorbs them before the promise resolves, but visible by instrumenting `fetch` directly. The limiter detected every one and halved the rate live, recovering additively in between.

So: the ceiling-discovery half of this PR is solid and verified live. The cold-transition-burst half is real at the CDN level but I couldn't reproduce it end-to-end through the client under my test conditions — worth the author double-checking the concurrency pattern used for that specific ADR measurement, since a casual reproduction attempt doesn't reliably surface it (possibly CDN request-coalescing dependent on region/POP/timing).

*This comment was left by an AI agent (Claude) reviewing this PR.*

maoberlehner:
> **claude:** QA finding — could not independently reproduce ADR-0017's cold-transition burst claim.
> 
> I ran real-world traffic (not mocks) against a live QA space to validate this: a real published 106-story listing warmed to the 6/s tier's 8x ceiling (climbed 6→48 r/s over ~29s, 0 429s, matching ADR-0017's own numbers closely). Then I republished a story to rotate the content version and kept reading through the transition...

My Claude disagrees with your Claude 🤣

> Re-ran on the PR head. The burst reproduces: a long-running client at the 48 r/s plateau that follows the new cv mid-run took 12 and 22 × 429 over ~1–2s (control: 0). Your concurrent row couldn't show it — 0 misses means it never went cold: either the preceding sequential pass had warmed the new-cv keys (reproduced: 105/106 hits), or the client was still pinned to the old cv, which a publish doesn't purge (reproduced: 106/106 hits at 400 r/s). Following the new version from that pinned state on the 50/s tier threw 60 of 106. We'll update the burst figures in ADR-0017 to a range.

packages/capi-client/src/utils/limiter.ts
maoberlehner: A field missing from the check would suppress reports when only that field changes, not fail to dedup. The object is built directly above the check, so the risk is low. I'll leave it as it is.

PR #825, opened 2026-09-22
fix(openapi): deduplicate generated union types

### LINKED ISSUE #792: [Bug]: New richtext types generated by CLI contain repetitions

dipankarmaikap:
## Summary

Deduplicates unions in generated OpenAPI TypeScript output and updates the generator to `@hey-api/openapi-ts@0.99.0`.

- Deduplicates unions in full SDK `types.gen.ts` output, reusing TypeScript printers during AST processing and preserving the first duplicate union member and its documentation.
- Upgrades `@hey-api/openapi-ts` from `0.92.4` to `0.99.0` and regenerates every OpenAPI consumer.
- Adapts to Hey API's optional request/response result contract for transport failures, while preserving the CAPI `network-first` cache fallback for transport errors.
- **Behavior change:** on a transport failure (no HTTP response — a timeout, an abort, a DNS
  failure), both clients now key off the *effective* `throwOnError` (client default, or the
  per-call override) instead of always producing the same shape:
  - `throwOnError: true` rejects with the **original error unwrapped** (`AbortError`,
    `TypeError`, etc.), so `err.name`/`instanceof` checks keep working — previously this could
    surface as a `ClientError` with an empty `cause`.
  - `throwOnError: false` resolves with `result.error` as a real `ClientError`
    (`status: 0`, `cause` set to the original error), matching the documented `ApiResponse`
    type instead of leaking the raw error where a `ClientError` is declared.
  - A malformed `baseUrl` (or any other failure before a `Request` could be built) no longer
    replaces the original error with a generic "invalid URL" error; the fallback `Request` is
    now built defensively and the original error/message is preserved via `cause`.
- Updates MAPI retry types to use `ky`'s `RetryOptions` directly.
- Updates richtext and schema consumers for Hey API's normalized generated names.
- Updates richtext generation to recognize the new node/mark names.
- Keeps the stricter asset validation diagnostics introduced by the regenerated Zod schema.
- Adds the required published-version discriminator to a CAPI inline-relations fixture.

## Verification

- `@storyblok/openapi-codegen`: 25 tests passed.
- `@storyblok/api-client`: 207 tests passed.
- `@storyblok/management-api-client`: 142 tests passed.
- `@storyblok/richtext`: 311 tests passed.
- `@storyblok/schema`: 290 tests passed.
- `@storyblok/live-preview`: 44 tests passed.
- CLI: 1,416 tests passed, 1 skipped, 1 todo.
- Lint and typecheck clean across all affected packages.

Fixes DX-564
Fixes #792


dipankarmaikap:
@maoberlehner Good catch. Confirmed: the invalid `discriminator: {propertyName: version}` on `Story` was fixed upstream in `storyblok/openapi-wdx@c47d7d106` — it's removed entirely rather than making `version` required, since neither the CDN nor MAPI actually send it.

Pulled the new spec (`spec.lock` → `c47d7d106`), regenerated every consumer, and dropped the `version: "published"` fixture field that was masking this — `Story` is now correctly typed as a plain `DraftStory | PublishedStory` union with no discriminator. All consumer test suites pass.

tools/openapi-codegen/README.md
```diff
@@ -31,6 +31,11 @@ would let a plain `pnpm install` silently change what the next regeneration prod
 `@hey-api/openapi-ts` is pinned to an exact version here. Bump it deliberately, regenerate every
 consumer in the same PR, and review the diff.
 
+Version `0.99.0` includes an upstream fix for redundant index-signature members and requires Node
+`>=22.18.0`. The shared CI setup reads `.nvmrc`, which currently selects Node 24, so the version is
+compatible with the effective CI runtime. This tool's AST pass remains necessary for the nested
+duplicate unions from #792 that `0.99.0` does not remove.
```
maoberlehner: > Version `0.99.0` includes an upstream fix for redundant index-signature members and requires Node
> `>=22.18.0`. The shared CI setup reads `.nvmrc`, which currently selects Node 24, so the version is
> compatible with the effective CI runtime.

I think at least this part of the prose is rather useless. And also the last sentence is not really README worthy IMO.

packages/capi-client/src/client.ts
```diff
@@ -373,13 +379,19 @@ export const createApiClientBase = <
...
       ...options,
       method,
       query,
       security,
       url: path,
     });
+
+    // Keep the public response contract stable when Hey API represents a
+    // transport failure without Response/Request objects.
+    const response = result.response ?? Response.error();
```
maoberlehner: Claude:

The normalization here only covers `client.get()`/`client.request()`. Resource calls go through `asApiResponse`, which is a pure cast, so they hand back whatever `client.gen.ts` produced: on a transport failure that's `{ error, request: undefined, response: undefined }`, while `ApiResponse` declares `response: Response` non-optional. Type-correct consumer code crashes:

```js
const c = createApiClient({ accessToken: 'x', baseUrl: 'http://127.0.0.1:1/v2', retry: { limit: 0 } });
const r = await c.stories.list();
r.response.status // TypeError: Cannot read properties of undefined (reading 'status')
```

On `main` that rejected with `TypeError: fetch failed` instead. MAPI gets this right, since `wrapRequest` normalizes on the path all its resources use, and the new test added here ("should return a synthetic response for transport failures") covers exactly that already-working path. Moving the normalization into `asApiResponse` would fix both this and the SWR problem below; a CAPI-side mirror of the MAPI test would catch it next time.

packages/capi-client/src/client.ts
```diff
@@ -46,7 +46,13 @@ export type HttpRequestOptions = Omit<RequestOptions, "method" | "security" | "u
 const getFailure = (result: ApiResponse): StrategyFailure | undefined =>
   result.error === undefined
     ? undefined
-    : { transient: isTransientStatus(result.response?.status ?? 0), error: result.error };
+    : {
+        // The generated client resolves transport failures with no response. There
+        // was no origin answer to classify, so network-first must treat this as
+        // transient and use its cached value.
+        transient: result.response.status === 0 || isTransientStatus(result.response.status),
```
maoberlehner: Claude:

This dropped the optional chain: `result.response?.status ?? 0` became an unguarded `result.response.status`. Together with the `asApiResponse` gap, it throws inside `reportGuarded`, which swallows it:

```
PR:   console.error "[storyblok/api-client] SWR revalidation reporting threw: TypeError: Cannot read properties of undefined (reading 'status')"
main: onRevalidationError: TypeError: fetch failed
```

So SWR consumers lose the outage signal entirely. `network-first` does still serve its cached value, but only because `createNetworkFirstStrategy`'s catch reclassifies that internal TypeError as transient, which means the behavior the comment above promises is reached by accident rather than by classification. Worth restoring `const status = result.response?.status ?? 0` regardless of where the normalization lands, plus a test asserting `onRevalidationError` receives the `ClientError`.

packages/capi-client/src/client.ts
```diff
@@ -317,7 +323,7 @@ export const createApiClientBase = <
   );
 
   client.interceptors.error.use(
-    (error: unknown, response: Response) =>
+    (error: unknown, response?: Response) =>
```
maoberlehner, line 327: Claude:

The interceptor discards its `error` argument, and `ClientError` never sets `cause`, so the originating error is gone. A pre-aborted signal used to surface as an `AbortError`; now it comes back as `ClientError: API request failed`, status 0, `cause` undefined, and with `throwOnError: false` it doesn't throw at all, it resolves. That breaks `if (err.name === 'AbortError') return;` and makes a user cancel indistinguishable from DNS failure or connection refused: same message, same status, same empty `data`. At minimum attach the original as `cause`. This behavior change isn't mentioned in the PR description.

packages/capi-client/src/generated/capi/types.gen.ts
```diff
@@ -313,8 +313,17 @@ export type DraftStory = {
         path: string;
         name: string | null;
         lang: string;
-        published: string | null;
+        published: boolean | null;
```
maoberlehner: `published` narrows from `string | null` to `boolean | null` here, and this member is part of the public `Story` type of `@storyblok/api-client`. Consumer typecheck against the built `.d.mts`, using `type P = NonNullable<Story["translated_slugs"]>[number]["published"]`, compiles on `main` and fails on this branch with `TS2322: Type 'string | null' is not assignable to type 'boolean | null'`. It comes from the `spec.lock` move bundled into this PR rather than from the dedup work, and the commits are `fix:`/`chore:`, so it would ship as a patch. Needs a release note at least. `@storyblok/management-api-client` isn't affected.

tools/openapi-codegen/src/transform.ts
```diff
@@ -211,30 +245,89 @@ function extractProperty(typeNode: ts.TypeNode, propName: string): ts.TypeNode |
...
+  return output;
+}
+
+/** Remove structurally identical members from unions emitted by @hey-api/openapi-ts. */
+function deduplicateUnions(
+  node: ts.TypeNode,
+  sourceText: string,
+  printer: ts.Printer,
+): ts.TypeNode {
+  const source = ts.createSourceFile("union.ts", sourceText, ts.ScriptTarget.Latest, false);
```
maoberlehner: Claude:

This parses the entire file text once per visited union. Measured on this PR's own `mapi/types.gen.ts` (953 KB, 1432 visited unions): **30.3 s** for one pass against **67 ms** for a single parse, and the pass reported `changed: false`, so that's half a minute producing nothing on that file. It runs once per consumer package per `generate:openapi`.

`source` is only used as the printer's `sourceFile` argument, so hoisting it to a single parse outside the visitor removes nearly all of it.

tools/openapi-codegen/src/transform.ts
```diff
@@ -211,30 +245,89 @@ function extractProperty(typeNode: ts.TypeNode, propName: string): ts.TypeNode |
...
 }
-
-function rewriteRefs(node: ts.TypeNode, renameMap: ReadonlyMap<string, string>): ts.TypeNode {
+function rewriteAndDeduplicate(
+  node: ts.TypeNode,
+  renameMap: ReadonlyMap<string, string>,
+  sourceText: string,
+  printer: ts.Printer,
+): ts.TypeNode {
+  const source = ts.createSourceFile("union.ts", sourceText, ts.ScriptTarget.Latest, false);
```
maoberlehner: Claude:

Same ~20 lines as `deduplicateUnions` below, and the same per-call full-file parse. Both also print nodes belonging to one `SourceFile` against a throwaway `SourceFile` built from `sourceText`; that only works because the text happens to match.

packages/richtext/src/static/generate/richtext-element-types.ts
```diff
@@ -53,7 +53,16 @@ function parseNodes(source: string): NodeEntry[] {
...
       continue;
     }
-    if (!declaration.name.startsWith("RichTextFieldValue")) {
+    // Hey API has changed how it names these declarations across releases:
+    // prefixed with the union name (`RichTextFieldValueParagraphNode`) in
+    // older releases, bare (`ParagraphNode`) in newer ones. Match either so
+    // regenerating with a new generator version doesn't silently drop every
+    // node/mark down to just the `RichTextFieldValue*` root declaration.
+    if (
+      !declaration.name.startsWith("RichTextFieldValue") &&
```
maoberlehner: Matching any exported declaration ending in `Node` or `Mark` does pick up the new bare names, and the regenerated output is a clean 32-for-32 rename. But it'll also pull in any future overlay type with that suffix and a `type: '<literal>'` member. The sibling test "should ignore declarations outside the RichTextFieldValue namespace" uses `SomethingElse`, so it no longer tests what its name claims. Anchoring on the `RichTextNode`/`RichTextMark` union members would keep the intent.

packages/richtext/src/extensions/richtext-attrs.ts
```diff
@@ -12,6 +12,12 @@ type ExtensionAttrMap = {
 };
 
 export type ExtensionAttrs<K extends ExtensionKey> = ExtensionAttrMap[K];
+export type NodeAttrTypeMap = {
```
maoberlehner: Claude:

Neither `NodeAttrTypeMap` nor `MarkAttrTypeMap` is referenced anywhere, and `richtext-attrs.ts` isn't re-exported from `src/index.ts`, so they're not public API either. Related: folding `RichTextDoc` into `Extension` widens `ExtensionKey` with `"doc"`, which lets `ExtensionsOptions` take a `doc` entry whose `parseHTML` resolves to `() => never | undefined`. Harmless, but meaningless.

tools/openapi-codegen/spec.lock
```diff
@@ -1,5 +1,5 @@
 {
   "repo": "storyblok/openapi-wdx",
-  "sha": "e138ab4bce8f88931bb32d81907222e7b425f600",
-  "hash": "sha256:937cd08d44e0b01db6e66e13ff6d7fe73419bedb48044d1cf2f3721dcf80d689"
+  "sha": "c47d7d1060275a4295d996673dc3eed8fb4bbe44",
```
maoberlehner: This spec sync rides along with the `0.92.4 → 0.99.0` bump, so the generated diff has three independent causes mixed into it: union dedup, Hey API renames, and changed backend schema. No reviewer can attribute a given line to one of them, and the `published` narrowing I flagged in `capi/types.gen.ts` is exactly the kind of change that disappears in that. Splitting the sync into its own commit would make both halves reviewable.

tools/openapi-codegen/src/known-types.ts
```diff
@@ -35,11 +35,29 @@ export const UPSTREAM_SCHEMAS = {
     "Story",
     "StoryAlternate",
     "Tag",
+    "TaxonomyTerm",
```
maoberlehner: Running the documented `pnpm --filter @storyblok/openapi-codegen pull` rewrites this file with single quotes and different parenthesization: a 402-line diff, identical content, because the committed version is oxfmt-formatted after generation. Looks pre-existing rather than introduced here, but it makes "regenerate and confirm no diff" noisy enough to cost real time. Running oxfmt as part of `pull` would settle it.

maoberlehner (changes requested):

packages/capi-client/src/client.ts
```diff
@@ -46,7 +46,14 @@ export type HttpRequestOptions = Omit<RequestOptions, "method" | "security" | "u
 const getFailure = (result: ApiResponse): StrategyFailure | undefined =>
   result.error === undefined
     ? undefined
-    : { transient: isTransientStatus(result.response?.status ?? 0), error: result.error };
+    : {
+        // The generated client resolves transport failures with no response. There
+        // was no origin answer to classify, so network-first must treat this as
+        // transient and use its cached value.
+        transient:
```
maoberlehner, line 53: Claude:

> With `throwOnError: true`, `network-first` and a warm cache, a dropped connection now rejects with `ClientError` (status 0, cause `TypeError: fetch failed`); `main` returns the cached story. The interceptor wraps transport failures in a `ClientError` now, and `getThrownFailure` (line 64) runs every `ClientError` through `isTransientStatus`, which returns false for 0, so the failure counts as definitive. This path got the `status === 0` case but `getThrownFailure` didn't. Could you treat status 0 as transient there too, and add a test for `throwOnError: true` + `HttpResponse.error()`? `throwOnError: false` is fine. As it stands, this contradicts "preserving the CAPI network-first cache fallback" in the description.

packages/capi-client/src/client.ts
```diff
@@ -317,11 +324,12 @@ export const createApiClientBase = <
   );
 
   client.interceptors.error.use(
```
maoberlehner, line 326: Claude:

> This changes behavior for both clients (same at `packages/mapi-client/src/client.ts:202`). With the default `throwOnError: false`, an unreachable host, a pre-aborted or mid-flight abort, or a timeout used to reject with `TypeError` / `AbortError` / `TimeoutError`. Now they resolve with `{ data: undefined, error: ClientError(status 0) }`, so an offline `try/catch` never reaches its `catch`. With `throwOnError: true` they reject with `ClientError`, so `err.name === 'AbortError'` checks stop matching and the abort is only reachable via `err.cause`. If this is the intended hey-api contract, fine, but it needs a changeset entry marking it as a behavior change; it's more than an internal adaptation under a `fix` title.

packages/capi-client/src/client.ts
```diff
@@ -367,30 +375,41 @@ export const createApiClientBase = <
...
   const asApiResponse = <TData, ThrowOnError extends boolean = false>(
     p: Promise<unknown>,
   ): Promise<ApiResponse<TData, ThrowOnError>> =>
-    p as unknown as Promise<ApiResponse<TData, ThrowOnError>>;
+    p.then((result) => {
+      const typedResult = result as ApiResponse<TData, ThrowOnError>;
+      return {
+        ...typedResult,
+        response: typedResult.response ?? Response.error(),
+        request: typedResult.request ?? new Request(baseUrl || getRegionBaseUrl(region)),
```
maoberlehner, line 394: Claude:

> With `baseUrl: 'not a url'`, `new Request(baseUrl || …)` throws while normalizing, so callers get `Failed to parse URL from not a url` instead of the original error with the full request URL (`…/v2/cdn/stories/x?token=t` on `main`). MAPI has the same at `client.ts:219`. Minor, but a guarded fallback would keep the useful message.

packages/capi-client/src/error.ts
```diff
@@ -15,10 +15,21 @@ export interface ApiErrorBody {
  */
 export class ClientError extends Error {
   readonly response: { status: number; statusText: string; data: ApiErrorBody | undefined };
+  /**
+   * The error the interceptor caught before wrapping it — an `AbortError`, a network
+   * failure, etc. `Error`'s own `cause` constructor option isn't available at this
+   * package's `lib` target, so this is declared and assigned as a plain own field
+   * instead.
+   */
+  readonly cause?: unknown;
```
maoberlehner, line 24: Claude:

> The JSDoc says `cause` is the error the interceptor caught (an AbortError, a network failure), but on a 404 it's the parsed JSON body. On transport failures `response.data` holds the raw `TypeError`, although it's typed `ApiErrorBody`. Setting `cause` only for non-HTTP errors, and keeping the Error out of `data`, would match the docs and the types. The class doc still says it's thrown when `throwOnError: true`, but it now also shows up in `result.error`.

packages/schema/src/validators/validate-story.test.ts
```diff
@@ -1098,6 +1098,8 @@ describe("validateStory — an empty object in an object-shaped field", () => {
     }
```
maoberlehner, line 1098: Claude:

> An empty asset `{}` now reports `id` and `alt` as well, with "expected number/string, received undefined". Both accept `null`, so the message is a bit misleading, but no valid content becomes invalid. Fine to keep if it's intended.

maoberlehner (approved):
Claude found a few inconsistencies and issues, mostly around error handling and reporting again. I leave it to you to judge whether we should act on them or not.

dipankarmaikap:
@maoberlehner I’ve made some changes based on Claude’s comments. Have a look when you get a chance.

maoberlehner (approved):
LGTM

PR #829, opened 2026-09-23
chore(repo): document commit types and drop the CLAUDE.md symlinks

maoberlehner:
- **Commit types:** `AGENTS.md` now says that `fix` and `feat` are for changes a package's consumers can observe, since both trigger a release. Tests, playgrounds, tooling, CI, and repo docs are `chore`.
- **`CLAUDE.md` symlinks:** removed at the root and in `packages/cli`. Claude Code reads `AGENTS.md` natively since [2.1.277](https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md) when a directory has no `CLAUDE.md`, including nested ones. Older versions and Bedrock, Vertex, or Foundry setups don't have this yet; a local `CLAUDE.md` containing `@AGENTS.md` covers them.

`.claude/skills` stays a symlink to `.agents/skills`, because Claude Code does not read `.agents/`.

Note: if a `CLAUDE.md` exists in a parent directory of the checkout, Claude Code reads that instead of `AGENTS.md` by default. Set "Project instructions" in `/config` to read both.


dipankarmaikap (approved):
LGTM

